PR #6, opened 2015-01-30
METAMODEL-109: CassandraDataContext throws IllegalArgumentException with

tomaszguzialek:
proper message instead of NPE when keystore does not exist.


kaspersorensen:
Looks solid to me. +1

PR #7, opened 2015-03-04
Changes in Query and SelectItemParser to use IN and LIKE operartors into lowercase and AVG, MIN, MAX... functions into lowercase

rafaelgarrote:
- Updated findFilterItem method in order to find either upper and lowercase filter operator types in core/src/main/java/org/apache/metamodel/query/Query.java

It fix "IllegalArgumentException: Expression-based filters cannot be manually evaluated" while using like and in operators into lowercase.

- Updated findSelectItem method in order to find either upper and lowercase function types in core/src/main/java/org/apache/metamodel/query/parser/SelectItemParser.java

It fix "org.apache.metamodel.query.parser.QueryParserException: Not capable of parsing SELECT token: ..." while using avg, min, max... functions into lowercase.

### DIFF
--- a/core/src/main/java/org/apache/metamodel/query/Query.java
+++ b/core/src/main/java/org/apache/metamodel/query/Query.java
@@ -297,6 +297,8 @@ public final class Query extends BaseObject implements Cloneable, Serializable {
     }
 
     private FilterItem findFilterItem(String expression) {
+        String _upperExpression = expression.toUpperCase();
+
         final QueryPartCollectionProcessor collectionProcessor = new QueryPartCollectionProcessor();
         new QueryPartParser(collectionProcessor, expression, " AND ", " OR ").parse();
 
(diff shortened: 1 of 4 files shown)


kaspersorensen:
Hi Rafael,

Thank you for the Pull Request! It's great to receive such a contribution. Can you better document if this was a problem before? Maybe create a unittest that would break without the fix and run with the fix? Reason I ask is that there is no reference to a JIRA issue or anything like that, and I was actually under the impression that our parser was not case sensitive, but maybe I was wrong.

rafaelgarrote:
Hi Kasper,

I also found the same problem with DISTINCT into lower case. I made some changes to solve it.
I added some test in org.apache.metamodel.query.parser.QueryParserTest that must fail without changes and pass with changes:
- testSelectDistinctInLowerCase()
- testWhereInInLowerCase()
- testWhereLikeInLowerCase()
- testSelectMinInLowerCase()
- testSelectAvgInLowerCase()

Thanks!

kaspersorensen:
Looks good! I'll merge it in :-)

PR #9, opened 2015-03-17
BugFix for #METAMODEL-120 & #METAMODEL-121

ardlema:


kaspersorensen:
Looks good to me. Will you merge it Alberto?

ardlema:
It seems like I do not have write access to the repo in github. Could you sort this out?

kaspersorensen:
Since the GitHub repo is only a mirror, it is not possible actually.

The way it works is a bit lame, but here it is. First take the .diff file from github:

https://github.com/apache/metamodel/pull/9.diff

Then apply this to the regular (apache) git repo. Commit it with a message that says "Fixes #9". You can also have other stuff in the commit message, but this text triggers the close of this PR.

albertostratio:
Thank you for the explanation Kasper. I will do it ASAP.

PR #10, opened 2015-03-17
Delimiter in where clause in lowercase

rafaelgarrote:
I found de same problem lower/uppercase with delimiters and, or in where clause.
I added a test that should fail without changes and pass with changes.


kaspersorensen:
Looks good to me.

PR #11, opened 2015-03-25
changing select(expression) method of Query to support multiple columns ...

narahari92:
This commit is related to issue METAMODEL - 125. This might not be the issue but it's annoying to have multiple select methods for each expression (ex. query.select("col1").select("max(col2)") especially if there are multiple such columns. This commit fixes this issue and now the above query can be specified as query.select("col1,max(col2)").

### DIFF
--- a/core/src/main/java/org/apache/metamodel/query/Query.java
+++ b/core/src/main/java/org/apache/metamodel/query/Query.java
@@ -122,14 +122,14 @@ public final class Query extends BaseObject implements Cloneable, Serializable {
      * @return
      */
     public Query select(String expression) {
-        if ("*".equals(expression)) {
-            return selectAll();
-        }
-        
         String[] possibleSelectItems = splitExpressionByComma(expression);
         for(String possibleSelectItem : possibleSelectItems) {
-              SelectItem selectItem = findSelectItem(possibleSelectItem, true);
-              select(selectItem);
+            if ("*".equals(possibleSelectItem)) {
+                selectAll();
+                continue;
+            }
+            SelectItem selectItem = findSelectItem(possibleSelectItem, true);
+            select(selectItem);
         }
         return this;
     }
--- a/core/src/main/java/org/apache/metamodel/query/Query.java
+++ b/core/src/main/java/org/apache/metamodel/query/Query.java
@@ -125,9 +125,35 @@ public final class Query extends BaseObject implements Cloneable, Serializable {
         if ("*".equals(expression)) {
             return selectAll();
         }
-
-        SelectItem selectItem = findSelectItem(expression, true);
-        return select(selectItem);
+        
+        String[] possibleSelectItems = splitExpressionByComma(expression);
+        for(String possibleSelectItem : possibleSelectItems) {
+              SelectItem selectItem = findSelectItem(possibleSelectItem, true);
+              select(selectItem);
+        }
+        return this;
+    }
+   
+    /**
+     * If multiple columns or expressions are given in comma seperated string, this methods splits and generates
+     * string array of expressions representing single column.
+     * @param expression
+     * @return
+     */
+    public String[] splitExpressionByComma(String expression) {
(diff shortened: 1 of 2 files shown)


kaspersorensen:
Could we add a unittest for this please? There are already a bunch of similar tests in AbstractDataContextTest or maybe it's QueryPostprocessDataContextTest.

core/src/main/java/org/apache/metamodel/query/Query.java
```diff
@@ -125,9 +125,35 @@ public Query select(String expression) {
...
+        return this;
+    }
+   
+    /**
+     * If multiple columns or expressions are given in comma seperated string, this methods splits and generates
+     * string array of expressions representing single column.
+     * @param expression
+     * @return
+     */
+    public String[] splitExpressionByComma(String expression) {
```
kaspersorensen: Maybe we can leverage some of the existing query parser code to do this. Please have a look at QueryParser class and whatever is related to select item parsing.

narahari92:
yes i'll add unit test as soon as possible.

narahari92:
Hi I have added unit test as you said. Please review

narahari92:
As per leveraging QueryParser, It's not always called. In my example I was testing on the queries with JdbcDataContext and the QueryParser was not called.

core/src/test/java/org/apache/metamodel/QueryPostprocessDataContextTest.java
```diff
@@ -1022,4 +1022,16 @@ public void testQueryWithDotInTableName() throws Exception {
 
         assertEquals("file.csv.foo = 'bar'", item.toSql());
     }
+    
+    public void testQueryWithMultipleColumnsInExpression() {
+        Query query1 = new Query().from(table1).select("contributor_id,name");
+        DataSet set = getDataContext().executeQuery(query1);
+        assertEquals(set.next(), true);
+        assertEquals(set.getRow().toString(), "Row[values=[1, kasper]]");
```
kaspersorensen: Not that it matters much, but I am pretty sure that "expected" and "actual" parameters are switched around here. Normally you have expectation as the first argument.

kaspersorensen:
Looks good to me. Let's keep it open for others to review for a day or two. If noone sees any issues, I can merge it in for you.

kaspersorensen:
Hi again,

sorry looked another time and in QueryParser (line 158-161) we are instrumenting a QueryClauseParser to do the select item parsing ... this is basically what I was referring to. If we use that code we can avoid adding more parsing logic in sparse places like the Query class.

narahari92:
Hey,

I had actually checked that, but it turns out that parseSelectClause method of QueryParser is not always called. For example I used like below..

1. Query query = new Query().select("username,password").from("USER_TABLE");
     context.executeQuery(query);

2. Table user  = context.getTableByQualifiedLabel("userdetails.USER_TABLE");
     Query query = new Query().select("username,password").from(user);
     context.executeQuery(query);

In both the above cases this method was not called. Or am I not supposed to instantiate Query class in this way?

narahari92:
And please correct me if I am wrong, according to my understanding, this QueryParser class is used only when we parse the query parsed as a string or executeQuery(String query) of Data context. But if we pass the already constructed query then this class and any parse methods won't be used at all and the functionality fails.

kaspersorensen:
I agree that it is (currently) not being used. But was hoping maybe we could arrange the code so that it will be reused. Your current patch does the job, but I am just affraid that it adds double maintenance of the codebase going forward. If we already have the "comma and parenthesis parsing" in place in QueryPartParser, then maybe we can reuse it.

narahari92:
oh.. okay then.. I'll look into it

narahari92:
Hey I have used the QueryPartParser and SelectItemParser as you said. And it's working great.

kaspersorensen:
Cool, then I'll finally merge it ;-) Thanks for the contribution and your patience.

PR #13, opened 2015-03-30
Fix for METAMODEL-128

narahari92:
Please review the code fix for the mentioned defect. One test case gives false failure because we are comparing query's string form when an alias of item is used. I have modified that.

### DIFF
--- a/core/src/main/java/org/apache/metamodel/query/parser/SelectItemParser.java
+++ b/core/src/main/java/org/apache/metamodel/query/parser/SelectItemParser.java
@@ -163,7 +163,7 @@ public final class SelectItemParser implements QueryPartProcessor {
             }
         }
         
-        //if the expression is alias of some select item defined return clone of that select item
+        //if the expression is alias of some select item defined return  that select item
         SelectItem selectItem = MetaModelHelper.getSelectItemByAlias(_query,expressionCopy);
         if(selectItem != null) {
         	return selectItem;
--- a/core/src/main/java/org/apache/metamodel/query/parser/SelectItemParser.java
+++ b/core/src/main/java/org/apache/metamodel/query/parser/SelectItemParser.java
@@ -18,7 +18,10 @@
  */
 package org.apache.metamodel.query.parser;
 
+import java.util.List;
+
 import org.apache.metamodel.MetaModelException;
+import org.apache.metamodel.MetaModelHelper;
 import org.apache.metamodel.query.FromItem;
 import org.apache.metamodel.query.FunctionType;
 import org.apache.metamodel.query.Query;
@@ -98,6 +101,7 @@ public final class SelectItemParser implements QueryPartProcessor {
      *             was passed in the expression
      */
     public SelectItem findSelectItem(String expression) throws MultipleSelectItemsParsedException {
+    	String expressionCopy = expression;
         if ("*".equals(expression)) {
             throw new MultipleSelectItemsParsedException(null);
         }
@@ -159,6 +163,14 @@ public final class SelectItemParser implements QueryPartProcessor {
                 }
                 return new SelectItem(subQuerySelectItem, fromItem);
             }
+        }
+        
+        //if the expression is alias of some select item defined return clone of that select item
+        List<SelectItem> allSelectItems = MetaModelHelper.getAllSelectItems(_query);
+        for(SelectItem selectItem : allSelectItems) {
+        	if(selectItem.getAlias() != null && selectItem.getAlias().equals(expressionCopy)) {
+        		return selectItem.clone();
+        	}
         }
 
         if (_allowExpressionBasedSelectItems) {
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/metamodel/query/parser/SelectItemParser.java
```diff
@@ -160,6 +164,14 @@ public SelectItem findSelectItem(String expression) throws MultipleSelectItemsPa
                 return new SelectItem(subQuerySelectItem, fromItem);
             }
         }
+        
+        //if the expression is alias of some select item defined return clone of that select item
+        List<SelectItem> allSelectItems = MetaModelHelper.getAllSelectItems(_query);
+        for(SelectItem selectItem : allSelectItems) {
+        	if(selectItem.getAlias() != null && selectItem.getAlias().equals(expressionCopy)) {
+        		return selectItem.clone();
```
kaspersorensen: Do we need to return the clone? Why not return the same select item? If it is referred to by alias (like a kind of pass by reference) then I think it makes sense to keep the same reference item and not a copy.

narahari92:
I thought returning the same reference might prove risky and have unexpected behavior. But yes, what you said makes sense.

kaspersorensen:
Yes normally I would also be in favor of making a copy. But in this case it is really a "reference" and not an independent item. So if you change that then I think we can make the general change a lot less - then you don't need to make the clone method public and you don't need to change the existing unittest.

narahari92:
No that existing unit test still changes because the alias name in query will be replaced by <tablename>.<columnName>.i.e. "foobarbaz" gets replaced by "a.foo"

narahari92:
Anyhow I'll make that change you suggested

narahari92:
Hi, I have made the change to return same item instead of clone.

core/src/main/java/org/apache/metamodel/query/parser/SelectItemParser.java
```diff
@@ -160,6 +164,14 @@ public SelectItem findSelectItem(String expression) throws MultipleSelectItemsPa
                 return new SelectItem(subQuerySelectItem, fromItem);
             }
         }
+        
+        //if the expression is alias of some select item defined return clone of that select item
+        List<SelectItem> allSelectItems = MetaModelHelper.getAllSelectItems(_query);
```
kaspersorensen: Also reflecting a bit on this ... Do we need to have this getAllSelectItems(...) method?

What I mean is, could there be aliased select items anywhere else than in the SELECT clause? I cannot imagine it really ... For instance in WHERE clause I have never seen anything a la:

SELECT ... FROM ... WHERE name AS n = "Kasper"

If we assume that this never happens then you can get rid of the getAllSelectItems(...) method which is already a bit greedy (makes new lists etc.). If I am wrong about the assumption (I seriously am in doubt) then I think the solution is good!

ashish-m-yh: Isn't a SELECT in WHERE clause possible in case of sub-select?

kaspersorensen: Well actually MetaModel's query API still does not support sub-select, so no. But even if it did, would it make sense to have an alias in the subquery that is then referred to in the outer query? At least then it would also be qualified I think, a la:

SELECT ..., (SELECT name as n FROM ...) AS q FROM ... WHERE q.n = 'foo'

But to summarize: I don't think this is a scenario we need to cover at all since only regular subqueries (in the FROM clause) is supported in MetaModel at the moment.

ashish-m-yh: Sounds good then.

narahari92:
Fine I'll remove that getAllSelectItems() method but isn't subquery in from and where class there in to do list of MetaModel. I remember seeing it in jira for improvement.

narahari92:
SELECT... FROM (SELECT name as n FROM ...) as q where n='foo';  is perfectly valid but will fail if we don't consider select items of other clauses if and when we support sub queries for from and where clause in MetaModel.

narahari92:
Hey, I noticed one thing just now. We can actually use subqueries in from clause at least for JdbcDataContext. Below is the query I ran which was successful.

Query query = dataContext.parse("SELECT password as X from (select username,password from AMASS_USER)  S");

kaspersorensen:
Correct, that's the one type of subquery we support. One that is in the FROM clause. We don't support it in select or where (currently).

narahari92:
We can also use subqueries in where clause. I successfully executed below query.

Query query = context.parseQuery("select username from AMASS_USER where password in (select password from AMASS_USER where enabled=1)");

Considering all the places where we can have alias name for columns wouldn't it be better to have the getAllSelectItems() method. Or shall I still remove the method and use only items from select clause.

kaspersorensen:
I believe that where clause will only work on JDBC because it gets carried along as one big expression. But MM unfortunately will not understand it so no select items will be there anyway. Hmm how about then making a getSelectItemByAlias(...) method. We could start with just items from SELECT but could eventually expand it if subquery support increases.

narahari92:
Ya.. that sounds like a good plan.. I'll make the necessary changes and commit it today

narahari92:
Committed with recommended changes...

PR #14, opened 2015-04-06
METAMODEL-131 defect fix

narahari92:
Hi,

I have made following changes to support creating tables with composite primary key. In unit test cases where sql statement is compared as string, I have made necessary changes. Please review.

### DIFF
--- a/core/src/main/java/org/apache/metamodel/create/AbstractTableCreationBuilder.java
+++ b/core/src/main/java/org/apache/metamodel/create/AbstractTableCreationBuilder.java
@@ -123,9 +123,22 @@ public abstract class AbstractTableCreationBuilder<U extends UpdateCallback> imp
                     && !column.isNullable().booleanValue()) {
                 sb.append(" NOT NULL");
             }
-            if (column.isPrimaryKey()) {
-                sb.append(" PRIMARY KEY");
-            }
+        }
+        boolean primaryKeyExists = false;
+        for(int i = 0 ; i < columns.length ; i++) {
+        	if(columns[i].isPrimaryKey()) {
+        		if(!primaryKeyExists) {
+        			sb.append(" , PRIMARY KEY(");
+        			sb.append(columns[i].getName());
+        			primaryKeyExists = true;
+        		} else {
+        			sb.append(",");
+        			sb.append(columns[i].getName());
+        		}
+        	}
+        }
+        if(primaryKeyExists) {
+        	sb.append(")");
         }
         sb.append(")");
         return sb.toString();
--- a/core/src/main/java/org/apache/metamodel/create/AbstractTableCreationBuilder.java
+++ b/core/src/main/java/org/apache/metamodel/create/AbstractTableCreationBuilder.java
@@ -128,7 +128,7 @@ public abstract class AbstractTableCreationBuilder<U extends UpdateCallback> imp
         for(int i = 0 ; i < columns.length ; i++) {
             if(columns[i].isPrimaryKey()) {
                 if(!primaryKeyExists) {
-                    sb.append(" ,PRIMARY KEY(");
+                    sb.append(", PRIMARY KEY(");
                     sb.append(columns[i].getName());
                     primaryKeyExists = true;
                 } else {
--- a/core/src/main/java/org/apache/metamodel/create/AbstractTableCreationBuilder.java
+++ b/core/src/main/java/org/apache/metamodel/create/AbstractTableCreationBuilder.java
@@ -126,19 +126,19 @@ public abstract class AbstractTableCreationBuilder<U extends UpdateCallback> imp
         }
         boolean primaryKeyExists = false;
         for(int i = 0 ; i < columns.length ; i++) {
-        	if(columns[i].isPrimaryKey()) {
-        		if(!primaryKeyExists) {
-        			sb.append(" , PRIMARY KEY(");
-        			sb.append(columns[i].getName());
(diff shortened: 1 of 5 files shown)


kaspersorensen:
Thanks for the patch, but I don't think this approach will fly. First I have to mention that there are code styling issues (tabs instead of spaces, no spaces in front of parenthesis etc.) but that's less important right now ...

Important part is: Different databases have different SQL dialects and in this patch it is proposed to switch from one dialect (that was very common but less flexible) to the other dialect which is AFAIK less supported.

Here's some sources of inspiration maybe:

[1] http://www.w3schools.com/sql/sql_primarykey.asp
[2] http://www.techonthenet.com/sql/primary_keys.php

To make this work we should only touch the JdbcCreateTableBuilder IMO. The abstract one is not supposed/guaranteed to generate perfect SQL for all databases, just like Query.toSql() will not generate SQL that is working on all databases. It is in the concrete implementation of CreateTableBuilder that we figure that part out (and in non-JDBC modules it is not a CREATE TABLE statement but some other operation anyway).

The JdbcCreateTableBuilder should take advice or delegate to the QueryRewriter interface. This is our current mechanism to allow different databases to have different dialects. I suggest we continue with that approach unless there's some better idea.

narahari92:
Yes, what you said is true. But the reason I thought of above fix is, all the databases METAMODEL supports have a same syntax for creating table. So It would be unnecessary overhead writing like QueryRewriter(Different databases have slightly different Query syntax). 

Is there any such database with different create table syntax?

narahari92:
Yes, above databases also have same syntax. 

And as for unit test cases, I have already committed it(DerbyTest and h2DatabaseTest with create table check). Is it enough?

core/src/main/java/org/apache/metamodel/create/AbstractTableCreationBuilder.java
```diff
@@ -123,9 +123,22 @@ public String toSql() {
...
                 sb.append(" NOT NULL");
             }
-            if (column.isPrimaryKey()) {
-                sb.append(" PRIMARY KEY");
-            }
+        }
+        boolean primaryKeyExists = false;
+        for(int i = 0 ; i < columns.length ; i++) {
+        	if(columns[i].isPrimaryKey()) {
+        		if(!primaryKeyExists) {
```
kaspersorensen: I see mixed use of tabs and spaces in this line and a few others.

core/src/main/java/org/apache/metamodel/create/AbstractTableCreationBuilder.java
```diff
@@ -123,9 +123,22 @@ public String toSql() {
...
+        			sb.append(" , PRIMARY KEY(");
```
kaspersorensen: The space in front of the comma is not needed. It's pretty minor I know, but let's keep it tidy :-)

core/src/main/java/org/apache/metamodel/create/AbstractTableCreationBuilder.java
kaspersorensen: Oh I see that the original file is actually using tabs ... strange. I guess we should reformat the whole file then :-/

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcCreateTableBuilder.java
```diff
@@ -114,10 +114,23 @@ private String createSqlStatement(Table table) {
...
+        			primaryKeyExists = true;
+        		} else {
+        			sb.append(",");
+        			sb.append(columns[i].getName());
+        		}
+        	}
+        }
+        if(primaryKeyExists) {
+        	sb.append(")");
+        }
```
kaspersorensen: Same comments as in AbstractTableCreationBuilder also applies here :)

narahari92:
oops.. sorry.. didn't see that.. I'll correct it :)

kaspersorensen:
OK let's then just take attention to the small code styling thingies :) Once done I will do a round of integration tests locally (not covered by Travis CI) so that I feel certain about giving it my +1 ;-)

narahari92:
Hey, I did made the formatting as discussed.

core/src/main/java/org/apache/metamodel/create/AbstractTableCreationBuilder.java
```diff
@@ -123,9 +123,22 @@ public String toSql() {
...
             }
-            if (column.isPrimaryKey()) {
-                sb.append(" PRIMARY KEY");
-            }
+        }
+        boolean primaryKeyExists = false;
+        for(int i = 0 ; i < columns.length ; i++) {
+            if(columns[i].isPrimaryKey()) {
+                if(!primaryKeyExists) {
+                    sb.append(" ,PRIMARY KEY(");
```
kaspersorensen: Heh, I meant the other space ... like we would normally write (space after each comma, none before):

one, two, three

kaspersorensen:
There's also a minor unittest failure now due to the comma-related change.

narahari92:
Ya..  I have changed it now along with the unit tests.

kaspersorensen:
My integration-test verification progress:

 - [ ] PostgreSQL
 - [ ] MySQL
 - [ ] DB2
 - [ ] SQL Server
 - [ ] Oracle

kaspersorensen:
Looks good! I will commit it with an addition of a bunch of integration tests :-)

PR #15, opened 2015-04-14
DataContextFactory.elasticsearchDataContext returns UpdateableDataContext

bobby:
- Brings DataContextFactory up to latest ElasticSearchDataContext capabilities


kaspersorensen:
Oh right, we should update that :)

PR #16, opened 2015-04-17
Solr connector added to existing DataContext classes

ashish-m-yh:
Have finally checked in the Solr DataContext code.

The following features will be worked on in future iterations -

1. Pivot search
2. More operations such as prefix search etc.
3. Implementation of aggregations through Solr Stat functions
4. Wider set of test cases

Any other suggestions welcome too.


full/pom.xml
```diff
@@ -151,6 +151,11 @@ under the License.
 			<version>${project.version}</version>
 		</dependency>
 		<dependency>
+            <groupId>org.apache.metamodel</groupId>
```
kaspersorensen: Indentation looks to be different here

full/src/main/java/org/apache/metamodel/DataContextFactory.java
```diff
@@ -652,6 +653,18 @@ public static UpdateableDataContext createElasticSearchDataContext(Client client
         return new ElasticSearchDataContext(client, indexName);
     }
 
+	/**
+    * Creates a new Solr datacontext
```
kaspersorensen: Also indentation strangeness.

solr/.classpath
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
```
kaspersorensen: We would like to keep the repo clean of project files from any particular IDE. Please create a .gitignore file instead of committing this file and the next ones.

solr/.project
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
```
kaspersorensen: Ditto

solr/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -0,0 +1,9 @@
+#Fri Jan 30 22:32:19 IST 2015
+org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.6
+eclipse.preferences.version=1
+encoding/src/main/java=utf-8
```
kaspersorensen: Ditto

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,429 @@
...
+
+		try {
+			HttpClient httpclient = new DefaultHttpClient();
+			HttpGet request = new HttpGet(url);
+			HttpResponse response = httpclient.execute(request);
+
+			InputStreamReader inputStream = new InputStreamReader(response
+					.getEntity().getContent());
+
+			ObjectReader reader = new ObjectMapper().reader(HashMap.class);
```
kaspersorensen: Is there not a Solr client API for doing these things? If not, OK, but to me it seems like a lot of work to do the JSON parsing manually - I could imagine many things going wrong with this if the response is not quite like we expect it.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,429 @@
...
+			Map<String, Object> outerMap = reader.readValue(inputStream);
+
+			Map<String, Object> schemaMap = (Map) outerMap.get("schema");
+
+			if (outerMap != null) {
```
kaspersorensen: I think this should be ```schemaMap != null```

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,429 @@
...
+				List<Map<String, String>> fieldMapList = (List) schemaMap
+						.get("fields");
+				List<String> columnNames = new ArrayList<String>();
+				List<ColumnType> columnTypes = new ArrayList<ColumnType>();
+
+				for (int i = 0; i < fieldMapList.size(); i++) {
+					Map<String, String> fieldMap = fieldMapList.get(i);
+
+					columnNames.add(fieldMap.get("name"));
+					columnTypes.add(ColumnType.VARCHAR);
```
kaspersorensen: I would suggest using ColumnType.STRING since it's more agnostic (symbolically)... Question: Does Solr not support non-string types?

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,429 @@
...
+	}
+
+	@Override
+	protected Schema getMainSchema() throws MetaModelException {
+		try {
+			final MutableSchema theSchema = new MutableSchema(
+					getMainSchemaName());
+
+			final MutableTable table = tableDef.toTable().setSchema(theSchema);
+			final Column idColumn = table.getColumnByName(FIELD_ID);
```
kaspersorensen: Does a field named '_id' always exist in Solr? Or else I would expect some special treatment to introduce this as a virtual/surrogate field in the ```detectSchema``` method above.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,429 @@
...
+
+	@Override
+	protected Number executeCountQuery(Table table,
+			List<FilterItem> whereItems, boolean functionApproximationAllowed) {
+		String queryStr = "*:*";
+
+		if (whereItems != null) {
+			for (int i = 0; i < whereItems.size(); i++) {
+				FilterItem whereItem = whereItems.get(i);
+				queryStr = whereItem.toString();
```
kaspersorensen: the 'queryStr' variable is being overridden for each time in the loop. So if more WHERE items are there, only the last will be effective.

I also am doubtful if Solr's query language is similar to the toString() on FilterItem. Does this work?

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,429 @@
...
+
+		if (filterItem != null) {
+			clause += " " + processSingleFilterItem(filterItem);
+		}
+
+		return clause;
+	}
+
+	@Override
+	public DataSet executeQuery(Query q) throws MetaModelException {
```
kaspersorensen: I would not override this method for our initial attempt at doing Solr ... Waaaay too much complexity involved in this method, so my advice is to start with just materializeMainTable and make that work well.

solr/src/test/java/org/apache/metamodel/solr/SolrDataContextTest1.java
```diff
@@ -0,0 +1,199 @@
...
+import org.apache.metamodel.schema.Column;
+import org.apache.metamodel.schema.ColumnType;
+import org.apache.metamodel.schema.Table;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Ignore;
+import org.junit.Test;
+
+public class SolrDataContextTest1 {
+	private static final String url = "http://localhost:8983/solr/collection1";
```
kaspersorensen: Please follow the standard way in MM towards integration tests - provide some optional properties that can be supplied by the user. See all the other modules for inspiration. With a test like this the tests will break anywhere else but on your machine I guess.

solr/src/test/java/org/apache/metamodel/solr/SolrDataContextTest1.java
```diff
@@ -0,0 +1,199 @@
...
+	public void testWhereWithLimit() throws Exception {
+		DataSet dataSet = dataContext
+				.executeQuery("SELECT * FROM collection1 WHERE (manu='maxtor' or manu='samsung') LIMIT 1");
+		try {
+			assertTrue(dataSet.next());
+
+			Row row = dataSet.getRow();
+			Object val = row.getValue(1);
+
+			assertEquals("1491367446686203904", val.toString());
```
kaspersorensen: Seems the data of the index is also something that only you have. Please make tests that either describe where to get data or make the tests themselves build up a test-database in the setUp() method or so.

ashish-m-yh:
I am closing this and creating a new pull request with responses to all the review comments.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
ashish-m-yh: There is no documented API for this. I have put in place necessary exception handling.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
ashish-m-yh: Exception handling has been taken care of.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
ashish-m-yh: Changed to STRING to be most agnostic. There might be some minor differences in data-types across versions, so I am keeping it as STRING for now. This won't affect the querying.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
ashish-m-yh: In Solr, id is usually part of the schema. Querying by Primary Key is anyway usually not the scenario in Solr.

ashish-m-yh: That can be a later enhancement.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
ashish-m-yh: Question: For executeCountQuery() are all the whereItems usually AND-ed or OR-ed by MetaModel convention? I have not changed this right now, but will do so based on your response to this question.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
ashish-m-yh: I have simplified this logic vastly but had to put in place a safe regex to handle one scenario which looks to be a MM bug. I will report this separately. However, this works reliably, as I have tested various query scenarios.

solr/src/test/java/org/apache/metamodel/solr/SolrDataContextTest1.java
ashish-m-yh: This test would run fine if there is a local Solr deployment. Tested with Solr 4.x. Testing against Solr 5.x is next action item. 

Embedded Solr does not work very reliably, so only way to run the test is if you have a real Solr deployment. The collection1 is a sample index which comes with Solr 4.x out of the box. I will try to find a better solution to this in near future when tackling Solr 5.x

There are other tests too in the MM suite which run in a similar way, such as SalesForce Data Context.

Same holds good for the below comment.

kaspersorensen:
For the future: You don't need to close the Pull Requests just because you have changes. The moment you commit new code to the branch then the existing PR is automatically updated. Closing it and opens a new one is a bit troublesome for the reviewers because we loose track of existing remarks.

PR #17, opened 2015-04-20
Csv

rafaelgarrote:
Hi!
In csv the metadata for all csv fields are STRING and the values are always String type.
I added some changes to get real metadata csv fields and the cast value for each field.
I made some new test and changes in others that sould fail without changes and works fine with changes.

### DIFF
--- a/csv/src/main/java/org/apache/metamodel/csv/CsvDataContext.java
+++ b/csv/src/main/java/org/apache/metamodel/csv/CsvDataContext.java
@@ -196,7 +196,7 @@ public final class CsvDataContext extends QueryPostprocessDataContext implements
     }
 
     /**
-     * tts the CSV configuration used
+     * Gets the CSV configuration used
      * 
      * @return a CSV configuration
      */
--- a/csv/src/main/java/org/apache/metamodel/csv/CsvDataContext.java
+++ b/csv/src/main/java/org/apache/metamodel/csv/CsvDataContext.java
@@ -196,7 +196,7 @@ public final class CsvDataContext extends QueryPostprocessDataContext implements
     }
 
     /**
-     * Gets the CSV configuration used
+     * tts the CSV configuration used
      * 
      * @return a CSV configuration
      */
--- a/csv/src/main/java/org/apache/metamodel/csv/CsvDataSet.java
+++ b/csv/src/main/java/org/apache/metamodel/csv/CsvDataSet.java
@@ -20,7 +20,6 @@ package org.apache.metamodel.csv;
 
 import java.io.IOException;
 
-import com.sun.org.apache.xpath.internal.operations.Bool;
 import org.apache.metamodel.MetaModelException;
 import org.apache.metamodel.data.AbstractDataSet;
 import org.apache.metamodel.data.DefaultRow;
--- a/csv/src/main/java/org/apache/metamodel/csv/CsvDataSet.java
+++ b/csv/src/main/java/org/apache/metamodel/csv/CsvDataSet.java
@@ -20,6 +20,7 @@ package org.apache.metamodel.csv;
 
 import java.io.IOException;
 
+import com.sun.org.apache.xpath.internal.operations.Bool;
 import org.apache.metamodel.MetaModelException;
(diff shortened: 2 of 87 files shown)


csv/src/main/java/org/apache/metamodel/csv/CsvDataContext.java
```diff
@@ -196,7 +196,7 @@ public CsvDataContext(InputStream inputStream, char separatorChar, char quoteCha
     }
 
     /**
-     * Gets the CSV configuration used
+     * tts the CSV configuration used
```
kaspersorensen: Eh? ;)

csv/src/main/java/org/apache/metamodel/csv/CsvDataSet.java
```diff
@@ -20,6 +20,7 @@
 
 import java.io.IOException;
 
+import com.sun.org.apache.xpath.internal.operations.Bool;
```
kaspersorensen: Eh? ;)

kaspersorensen:
Hi Rafael,

Thank you for the contribution, really appreciate the work. I am aware of the issue but I think this solution is a bit too fragile. It only reads the first line to determine the data type and it also does type casting/narrowing on each single value instead of trying to establish a data type and only narrow according to that type. That essentially means that as a user I have no guarantee that row 1 is going to have the same data types as row 2. That feels very unsafe to me at least. I could effectively have a table with e.g. column type NUMBER which would make sense on the first row but on the second row I would get a String or something completely different.

rafaelgarrote:
Hi Kasper, I have made some changes in code, and now I evaluate the first 1000 CSV rows in order to get the column types. Maybe this solution is less fragile.

albertostratio:
I've been having a look at these changes and IMHO they address quite well the problem that Kasper identified. They are also pretty consinstent with the approach taken in the MongoDB connector.

kaspersorensen:
My worry is still there I am afraid. With the change in CsvDataSet there is now a call to the cast(...) method which casts/narrows the object type on a value-by-value basis. That means that I can very easily be getting values of different types within the same column.

In my opinion it would be much better if we during the analysis of the data set a ColumnType on the Column definition. And then we _only_ do the casting as per the column type. That way we ensure that it is "always or never" that we change the objects we read.

Another aspect that I'd like to request is that this feature can be turned on or off in CsvConfiguration. For the application that I am working on CSV is one of the most important data source types and we have some cases of processing billions of CSV records. If there is just a single "foo bar" value in the middle of such a file with otherwise entirely numbers or so, then we need a way to turn off the casting so that the consumer of the data at least does not get inconsistent value types (or loss of data in case the cast would omit the "foo bar" value).

kaspersorensen:
Thinking about this some more, I guess the main principle that I am trying to enforce is:

_metadata about a column must be consistent with the values you get when you query that column_

In other words, it is currently consistent that ColumnType is always STRING because we always return a String value.

I would be very open to a way of working with CSV files where you could specify/override the column type of certain columns. That would actually be a pretty cool feature - very related, but a different way to see it.

If that would be the case then you could first use the untouched CsvDataContext to analyze the content of various columns. If you feel confident enough you might then switch the column type from STRING to something else. And when it is then set to eg. INTEGER then our cast method should automatically apply a Integer cast/conversion operation. Same with other types then.

Would that maybe be a model that fits us all?

kaspersorensen:
Could we either close this PR or continue discussing it? It seems it has gone a bit stale :-)

drexler42:
I'll just add my 2 cents... The CSV data type is not "typed", like a data base. It would be nice if the user can impose a data model on top of the CSV file, to express his intention. For example, the CSV file could be a database dump file. The user would then like to declare String, Numerical and Boolean columns.

The first thing the user wants is then to declare the column types. Detecting column types is much less needed IMHO. The "foo bar" would result in a type exception for the row.

Next, we could think about handling exceptional cases like:
- Handling missing values with some sort of coalesce function to handling missing values;
- Converting invalid values into some special value;

This does not conflict with Kaspers notion of changing the column types. But it is a bit less ambitious. I would be very content if it were possible to declare column types statically.

PR #19, opened 2015-04-26
Fresh PR for Sor

ashish-m-yh:


solr/src/main/java/org/apache/metamodel/solr/SolrDataSet.java
```diff
@@ -0,0 +1,174 @@
...
+import org.apache.solr.client.solrj.response.FacetField.Count;
+import org.apache.solr.common.SolrDocument;
+import org.apache.solr.common.SolrDocumentList;
+
+/**
+ * {@link DataSet} implementation for ElasticSearch
+ */
+final class SolrDataSet extends AbstractDataSet {
+
+	private static final Logger logger = LoggerFactory
```
kaspersorensen: Indentation seems to still be tabs instead of 4 spaces

solr/src/test/java/org/apache/metamodel/solr/SolrDataContextTest1.java
```diff
@@ -0,0 +1,199 @@
...
+import org.apache.metamodel.schema.Column;
+import org.apache.metamodel.schema.ColumnType;
+import org.apache.metamodel.schema.Table;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Ignore;
+import org.junit.Test;
+
+public class SolrDataContextTest1 {
+	private static final String url = "http://localhost:8983/solr/collection1";
```
kaspersorensen: Indentation seems to still be tabs instead of 4 spaces

solr/src/test/java/org/apache/metamodel/solr/SolrDataContextTest1.java
```diff
@@ -0,0 +1,199 @@
...
+import org.apache.metamodel.query.SelectItem;
+import org.apache.metamodel.schema.Column;
+import org.apache.metamodel.schema.ColumnType;
+import org.apache.metamodel.schema.Table;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Ignore;
+import org.junit.Test;
+
+public class SolrDataContextTest1 {
```
kaspersorensen: This class is somehow not running as part of continuous build. I think it is because the class name ends on 1 ...

And if we enable it, I am sure it fails, which relates to my previous remark ... The tests should be able to run on any machine. But externalizing the connection info is valid, like we do it in SalesforceDataContextTest and others. Here we have properties that are optional in the __metamodel-integrationtest-configuration.properties__ file in your user home. If the properties are not specified, then we programmatically ensure that the test is skipped. That means that anyone with a Solr instance can configure the test to run without any source code modifications.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,403 @@
...
+ * specification of a schema, through the {@link SimpleTableDef} class.
+ */
+
+public class SolrDataContext extends QueryPostprocessDataContext implements
+        DataContext {
+
+    private static final Logger logger = LoggerFactory
+            .getLogger(SolrDataContext.class);
+
+    public static final String FIELD_ID = "id";
```
kaspersorensen: I don't understand why we have a special logic for adding an ID field when there is no special logic for querying it. I think it might very well even be a source of bugs in the current code because if you add a column called "id" to every table and then don't do anything to ensure that there are values in the rows you return for this "id" column, then the rows' data and metadata will be inconsistent.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,403 @@
...
+import org.apache.solr.common.SolrDocumentList;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectReader;
+
+/**
+ * DataContext implementation for ElasticSearch analytics engine.
```
kaspersorensen: Too much copy/paste here :) I strongly assume the javadoc for ElasticSearch does not apply here.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,403 @@
...
+
+    public static final String FIELD_ID = "id";
+
+    private final SimpleTableDef tableDef;
+    private final String indexName;
+    private final String url;
+
+    private SolrQuery query = new SolrQuery();
+
+    private static enum queryTypes {
```
kaspersorensen: Normally any type (also enum) is written in uppercase and in singular grammatical form.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,403 @@
...
+                queryStr = whereItem.toString();
+            }
+        }
+
+        QueryResponse response = selectRows(table, null, queryStr, 0, 0);
+        return response.getResults().getNumFound();
+    }
+
+    @Override
+    public DataSet executeQuery(Query q) throws MetaModelException {
```
kaspersorensen: Overriding the executeQuery(...) method is a source of many painful things because the query object can represent _any_ query. That means you need to be very very careful with safe-guarding your code against unexpected scenarios. To give an example of just one thing that will go very wrong here: What if the query has multiple FROM items - or even a JOIN in it? Or what if there are aliases shared between the clauses. As a subclass of QueryPostprocessDataContext we try to liberate you from the responsibility of having to consider all such scenarios. There are a bunch of other methods available that are a lot easier to implement. And if you really want to override executeQuery(...) then add a lot of safe-guarding so that you ensure you only handle specifically the type of query that you have in mind.

solr/src/main/java/org/apache/metamodel/solr/SolrDataSet.java
ashish-m-yh: Pushed after fixing

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
ashish-m-yh: Well, this may be redundant for now, so I removed it.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
ashish-m-yh: My bad! Fixed this too

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
ashish-m-yh: Bit of Java rustiness there. Fixed it.

solr/src/test/java/org/apache/metamodel/solr/SolrDataContextTest1.java
ashish-m-yh: Fixed

solr/src/test/java/org/apache/metamodel/solr/SolrDataContextTest1.java
ashish-m-yh: Have made necessary changes and committed the code with correct filename and indentation.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
ashish-m-yh: If I do not override executeQuery(...), then the Solr index cannot be queried by SQL, right? Also, I am calling super.executeQuery(q) at the end, to handle any other kind of query beyond the specific clauses I am handling. If the query is totally invalid and goes beyond any query which can be handled, exception may be thrown or DataSet null returned.

If you feel this approach still has some caveats, what method should be overridden so that the SQL interface still works?

What's your take?

kaspersorensen: The SQL interface works well if you do not override this method at all. The method is already implemented in the superclass and will delegate to some of the protected methods that you can choose to override. It is also possible to override this method, but it comes at a much greater risk becase the Query object is so multi-facetted that your code really has to take a ton of scenarios into account. I'll try and make line-notes below so you can see...

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,396 @@
...
+        String queryStr = "";
+        QueryResponse response = null;
+
+        Schema schema = getMainSchema();
```
kaspersorensen: The schema is referred to in the Query, it might not be the main schema - it might be the information schema.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,396 @@
...
+        Table table = schema.getTable(0);
```
kaspersorensen: The table is referred to in the Query (somewhere in the FROM clause).

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,396 @@
...
+        Column[] columns = table.getColumns();
+
+        boolean selectAll = true;
+        List<SelectItem> selectItems = q.getSelectClause().getItems();
+
+        for (SelectItem selectItem : selectItems) {
+            if (selectItem.getFunction() != null
+                    || selectItem.getColumn() == null) {
```
kaspersorensen: This loop seems to check for any SelectItem that has a function and no column. But you update a variable called "selectAll". If your intent is to find out if this is a SELECT COUNT(*) FROM [table] kind of query, then there is a much safer way to implement it - to overrid the executeCountQuery(...) method of the superclass.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,396 @@
...
+        }
+        
+        FilterClause clause = q.getWhereClause();
+
+        if (clause != null) {
+            List<FilterItem> filterItemList = clause.getItems();
+
+            boolean isFirst = true;
+            
+            for (FilterItem filterItem : filterItemList) {
```
kaspersorensen: Some filter items may have children, if they represent (x OR y) kind of scenarios.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,396 @@
...
+                String expr = filterItem.getExpression();
+
+                if (expr == null){
+                    expr = filterItem.toString();
```
kaspersorensen: This is the most normal case, yet you are not checking on what kind of operator is being applied in the WHERE item. Does Solr support the same operators as MetaModel? To rewrite it into Solr syntax I would imagine a switch-case here on the basis of the operator type.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,396 @@
...
+                isFirst = false;
+            }
+        }
+        
+        if (queryStr.isEmpty()) {
+            queryStr = "*:*";
+        }
+
+        Integer limitObject = q.getMaxRows();
+        int limit = DEFAULT_LIMIT;
```
kaspersorensen: Setting a default limit means that this module behaves differently than other modules. Why is this necesary? It kinda breaks the polymorphic nature of all DataContext implementations - we'd want to facilitate a _write once, run on any datastore_ kind of API.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,396 @@
...
+        final List<GroupByItem> groupByItems = q.getGroupByClause().getItems();
+
+        if (groupByItems.size() > 0) {
+            query.clear();
+            query.setQuery(queryStr);
+            query.setFacet(true);
+
+            for (int i = 0; i < groupByItems.size(); i++) {
+                String facetField = groupByItems.get(i).getSelectItem()
+                        .getColumn().getName();
```
kaspersorensen: You're guaranteed that there will be a SelectItem, but not that it has a column reference. Acknowledged, it's not very common to have GROUP BY without a column reference, but it _could_ be a corner case of some sort and then you'll run into a NPE here.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,396 @@
...
+                query.addFacetField(facetField);
+            }
+
+            setOrder(q, query, QUERYTYPE.FACET);
+
+            query.setFacetLimit(limit);
+
+            try {
+                HttpSolrServer server = initSolrServer();
+                response = server.query(query);
```
kaspersorensen: At this point there has been no regard to the HAVING clause of the query. There might be HAVING items.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,396 @@
...
+                
+                isFirst = false;
+            }
+        }
+        
+        if (queryStr.isEmpty()) {
+            queryStr = "*:*";
+        }
+
+        Integer limitObject = q.getMaxRows();
```
kaspersorensen: Max rows is used for limit, but what about "firstRow"? There might also be an offset.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
```diff
@@ -0,0 +1,396 @@
...
+        int limit = DEFAULT_LIMIT;
+
+        if (limitObject != null) {
+            limit = (int) limitObject;
+        }
+
+        final List<GroupByItem> groupByItems = q.getGroupByClause().getItems();
+
+        if (groupByItems.size() > 0) {
```
kaspersorensen: Group by is one way of determining the faceting, but there is also another - the DISTINCT operator on the SELECT clause. If you have a DISTINCT situation, then I guess you might also need to apply faceting (or at least safeguard this code to not run without regard of the DISTINCT situation).

kaspersorensen:
Sorry that was a lot of nitty gritty details. I am just trying to show that implementing executeQuery(...) is very complex because the Query object is complex. I really think that you/we should start with just "the simplest thing that works" and then work our way from there. First make it work, then make it fast.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
ashish-m-yh: Tested this with precedence cases (bracketed) and it works.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
ashish-m-yh: Fair point

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
ashish-m-yh: Solr has a default facet limit anyway beyond which it will not fetch facets. Hence, DEFAULT_LIMIT has to be set to force it to fetch all.

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
ashish-m-yh: Have implemented it

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
ashish-m-yh: Joins are invalid on Solr

solr/src/main/java/org/apache/metamodel/solr/SolrDataContext.java
ashish-m-yh: Tested with the known cases

ashish-m-yh:
Thank you for the detailed comments. I concur in part - some corner cases may need handling. Some kinds of queries, such as joins are rarely used in Solr and the support is inefficient.

if executeQuery() is not implemented, then is there any other method which makes sense to implement? If not, do you recommend deferring the release of this till all corner cases are handled? 

The module is almost useless if the actual search does not take place in the engine itself, as users would generally use search engines for large data sets, which practically cannot be read into the MetaModel application space and searched in the MM layer. Hence, I was wondering if it has to be a all or nothing approach and if we could throw exceptions for unsupported query patterns and incrementally handle the rare query patterns too.

Do let me know your thoughts.

kaspersorensen:
Hi Ashish,

I know that doing a JOIN is pretty unusual in "normal" mode of Solr. But in my opinion we are also not catering to the normal cases. MetaModel exposes an API that is useful especially because it is the _same_ API across all datastores. That means our users often use MetaModel in scenarios where they might just as well use a relational database or an Excel sheet as the backing store. The benefit they get with MetaModel is that their code to process data can run on any given datastore. That means the type of use is often pretty untypical. So having support for JOIN in Solr is actually pretty important in my opinion. I would even go as far as to say that effeciency in execution is secondary to compatibility with the API. Reason being that if people are trying to get the most performance out of their Solr system then they will anyways be going directly to Solr's primary API.

That said I do believe that there are scenarios you absolutely should support.

My suggestion is just that you start small and get something that functionally fulfills the whole API. Then expand on a case-by-case basis to add optimized support for various classes of queries. For instance you can start with just materializeMainSchemaTable - make it work without any default limits or anything like that. And make sure that it is cursor based so that you don't load a million records into memory but rather maintain a cursor in the dataset implementation.

Once that is working I would expand the scope to also have support for COUNT queries. And then maybe queries with just a single WHERE item. And so on and so on ... Makes sense?

ashish-m-yh:
I understand the philosophy. We may sacrifice performance to conform to a general standard. I have implemented only the necessary methods.

Note: Only necessity is to have a DEFAULT_LIMIT constant in the Solr connector, as this is a pre-requisite of Solr and I think it's reasonable to impose a limit of 10000. If we keep INTEGER.MAX_VALUE as limit, then Solr will exhaust Heap Memory on most occasions.

Please have a look and proceed accordingly.

kaspersorensen:
If you run a broad query like "SELECT * FROM [table]" then it should be able to get the results and not apply a limit. Not necesarily as part of the same request (that would in deed cause huge memory issues) but then using a paging or cursor approach. I did a quick google on "solr paging" and this seems quite doable: https://cwiki.apache.org/confluence/display/solr/Pagination+of+Results

ashish-m-yh:
Introduced the "fetch all" mechanism as described on the link you found.

kaspersorensen:
A couple of remarks on code style:

 * In eclipse I see a bunch of warnings that should be fixed:
  * Unused variables (multiple)
  * Unused imports (multiple)
  * Deprecated classes (DefaultHttpClient)
  * Use of raw type generics (multiple)
  * Resource leak: HttpClient is never used
 * Unstandard naming: QUERYTYPE is entirely uppercase, should be QueryType
 * For me to read a very inconsistent use of "this" keyword and instance variables - difficult state management especially in the dataset class.

Bringing these items to order is for me a prerequisite for doing a proper review because readability isn't very good as long as I have to deal with so many warnings etc.

kaspersorensen:
Sorry that I keep asking you for changes like this, we'll be happy at the end of it all :-)

ashish-m-yh:
It's fine. Things should be proper. On further review, I found some of the issues you reported, but it's weird why my Eclipse doesn't show it when I import the mvn project. Is there some Eclipse option/configuration which I should turn on? My projects which are created purely in Eclipse do highlight these.

ashish-m-yh:
I used checkstyle static code analyzer to identify number of style issues (including the ones you raised) and fixed them.

However, throwing out DefaultHttpClient didn't work for some weird reason. Strangely, my Eclipse does not even show it is deprecated. :(

The following piece of code gives compiler error "cannot find symbol" though correct jar is in classpath etc.

HttpClientBuilder httpclientBuilder = HttpClientBuilder().create();

ashish-m-yh:
METAMODEL-129: Commited

ashish-m-yh:
Did you have a chance to look at this sometime?

ardlema:
@kaspersorensen @ashish-m-yh this PR seems to be here for a long time.

Shall we close it?

kaspersorensen:
I think so. I'm sorry @ashish-m-yh but for me this PR is just a bit too far from finished to give anymore constructive feedback on. I suggest to spend a little more time on reviewing similar modules such as the ElasticSearch module, MongoDB, CouchDB or other modules - and then implement in a similar fashion.

PR #20, opened 2015-04-27
Metamodel 133

narahari92:
Hi,

This is a fix for METAMODEL-133 issue for supporting multiple joins. Please review.

### DIFF
--- a/core/src/main/java/org/apache/metamodel/query/FromItem.java
+++ b/core/src/main/java/org/apache/metamodel/query/FromItem.java
@@ -267,6 +267,11 @@ public class FromItem extends BaseObject implements QueryItem, Cloneable {
             sb.append(superQueryAlias);
             return;
         }
+        
+        if(_join != null && _leftSide.getJoin() != null) {
+            sb.append(onItem.toSql());
+            return;
+        }
 
         if (sideAlias != null) {
             sb.append(sideAlias);
--- a/core/src/main/java/org/apache/metamodel/query/parser/FromItemParser.java
+++ b/core/src/main/java/org/apache/metamodel/query/parser/FromItemParser.java
@@ -71,11 +71,7 @@ final class FromItemParser implements QueryPartProcessor {
                 fromItem.setAlias(alias);
             }
         } else if (itemToken.toUpperCase().indexOf(" JOIN ") != -1) {
-            if(itemToken.split("(?i) JOIN ").length == 2) {
-                fromItem = parseJoinItem(null,itemToken,new HashSet<FromItem>());
-            } else {
-                fromItem = parseAllJoinItems(itemToken);
-            }
+            fromItem = parseAllJoinItems(itemToken);
         } else {
             fromItem = parseTableItem(itemToken);
         }
@@ -115,7 +111,7 @@ final class FromItemParser implements QueryPartProcessor {
             String joinType = joinSplit[i].substring(joinSplit[i].lastIndexOf(" "));
             String rightPart = (i+1 == joinSplit.length-1) ? joinSplit[i+1] : joinSplit[i+1].substring(0, joinSplit[i+1].lastIndexOf(" "));
             joinsList.add((leftPart + " " + joinType + " JOIN " + rightPart).replaceAll(" +", " "));
-            String rightTable = rightPart.substring(0, rightPart.lastIndexOf(" ON "));
+            String rightTable = rightPart.substring(0, rightPart.toUpperCase().lastIndexOf(" ON "));
             String nextJoinType = joinSplit[i+1].substring(joinSplit[i+1].lastIndexOf(" "));
             joinSplit[i+1] = rightTable + " " + nextJoinType;
         }
@@ -129,7 +125,7 @@ final class FromItemParser implements QueryPartProcessor {
 
     // this method will be documented based on this example itemToken: FOO f
     // INNER JOIN BAR b ON f.id = b.id
-    private FromItem parseJoinItem(final FromItem leftFromItem,final String itemToken,Set<FromItem> fromItems) {
+    private FromItem parseJoinItem(final FromItem leftFromItem, final String itemToken, Set<FromItem> fromItems) {
         final int indexOfJoin = itemToken.toUpperCase().indexOf(" JOIN ");
 
         // firstPart = "FOO f INNER"
--- a/core/src/main/java/org/apache/metamodel/query/parser/FromItemParser.java
+++ b/core/src/main/java/org/apache/metamodel/query/parser/FromItemParser.java
@@ -18,6 +18,11 @@
  */
 package org.apache.metamodel.query.parser;
 
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
 import org.apache.metamodel.DataContext;
 import org.apache.metamodel.query.FromItem;
 import org.apache.metamodel.query.JoinType;
@@ -66,7 +71,11 @@ final class FromItemParser implements QueryPartProcessor {
                 fromItem.setAlias(alias);
             }
         } else if (itemToken.toUpperCase().indexOf(" JOIN ") != -1) {
-            fromItem = parseJoinItem(itemToken);
+            if(itemToken.split("(?i) JOIN ").length == 2) {
+                fromItem = parseJoinItem(null,itemToken,new HashSet<FromItem>());
+            } else {
+                fromItem = parseAllJoinItems(itemToken);
+            }
         } else {
             fromItem = parseTableItem(itemToken);
         }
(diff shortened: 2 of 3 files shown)


kaspersorensen:
My 5 cents:

 * I like the approach you've taken here with the compound from items array.
 * But shouldn't we rather then take a step further and delete the "leftItem" and "rightItem" and let the array replace them.
 * Similarly I would expect an array for the "on items" then. And then delete the "leftOn" and "rightOn" fields.
 * Then you would no longer need "_isCompoundJoin" - it would always be a (compound) join if there is an array available.

Alternative approach is to keep FromItem as it is and just apply the nested FromItem structure when parsing the query. That would retain 100% backwards compatibility and only do something in the query parsing layer to fix the issue. Might also be an option.

narahari92:
Hi,

I looked into the points you mentioned above. Below are my thoughts.

1. We actually need "leftItem" and "rightItem" as we even use it in compound from items ultimately, since it's the only way to represent join. 

2. "on items" are indeed array of select items and for each from item we need them seperately since there might be other operations than "=".

3. I didn't try about nested from items but it needs code change more than this. Moreover with this approach we are using existing code which is both unit tested and time tested.

4. This doesn't cause any api changes either as there is no change in signature of any public method.

narahari92:
And any idea why travis ci build failed. It ran perfectly fine while I unit tested it. Could it be because of other PRs pending?

kaspersorensen:
But it is possible to do multiple joins using the existing API, it's just a bit cumbersome IMO and it has no query parser support. Here is an example:

```
SELECT COUNT(*) FROM
 (table1 INNER JOIN table2 ON table1.key1 = table2.key2) a
 INNER JOIN table3 ON a.key2 = table3.key3
```

```
Table table1 = ...
Table table2 = ...
Table table3 = ...

FromItem joinPart1 = new FromItem(table1);
FromItem joinPart2 = new FromItem(table2);
FromItem joinPart3 = new FromItem(table3);
SelectItem key1 = new SelectItem(table1.getColumnByName("key"))
SelectItem key2 = new SelectItem(table1.getColumnByName("key"))
SelectItem key3 = new SelectItem(table1.getColumnByName("key"))

FromItem join1 = new FromItem(JoinType.INNER, joinPart1, joinPart2, new SelectItem[] {key1}, new SelectItem[] {key2});
join1.setAlias("a");

SelectItem key2inJoin1 = new SelectItem(key2, join1);

FromItem join2 = new FromItem(JoinType.INNER, join1, joinPart3, new SelectItem[] {key2inJoin1}, new SelectItem[] {key3});

Query q = new Query();
q.selectCount();
q.from(join2);
```

Very cumbersome in deed ... We very much need parsing here AND probably also some more convenience constructors.

kaspersorensen:
And sorry no I don't know why Travis failed ... Seemed to happen in the elasticsearch module which normally works fine. Can you try a local build?

narahari92:
But if we use multiple joins as above wouldn't it cause problem in case let's say there is a column with name "key2" in both table1 and table2?

kaspersorensen:
It's a bit beyond my memory actually... Let me say it like this: I think it works because I am not aware of any bugs in that regard :-)

narahari92:
Hi,

I have used recursion as you suggested. And yes there was very less code change :) Please review

core/src/test/java/org/apache/metamodel/query/FromItemTest.java
```diff
@@ -98,4 +105,35 @@ public void testSubQueryJoinToString() throws Exception {
...
 				from.toString());
 	}
+	
+    public void testCompoundJoin() {
+        final Schema schema = getExampleSchema();
+   
+        final QueryPostprocessDataContext dc = new QueryPostprocessDataContext() {
+            @Override
+            protected DataSet materializeMainSchemaTable(Table table, Column[] columns, int maxRows) {
+                Object[] values = new Object[columns.length];
```
kaspersorensen: Implementing this method seems needless since the query is never executed. So why not just throw UnsupportedOperationException here?

core/src/test/java/org/apache/metamodel/query/FromItemTest.java
```diff
@@ -98,4 +105,35 @@ public void testSubQueryJoinToString() throws Exception {
...
+            }
+
+            @Override
+            protected Schema getMainSchema() throws MetaModelException {
+                return schema;
+            }
+        };
+ 
+        Query query = dc.parseQuery("SELECT c.contributor_id,p.project_id from contributor c INNER JOIN role r ON c.contributor_id=r.contributor_id INNER JOIN project p ON p.project_id=r.project_id");
+        System.out.println(query);
```
kaspersorensen: Probably a left-over from a debugging session or so, but let's not print to syso.

core/src/main/java/org/apache/metamodel/query/parser/FromItemParser.java
```diff
@@ -66,7 +71,11 @@ public void parse(String delim, String itemToken) {
                 fromItem.setAlias(alias);
             }
         } else if (itemToken.toUpperCase().indexOf(" JOIN ") != -1) {
-            fromItem = parseJoinItem(itemToken);
+            if(itemToken.split("(?i) JOIN ").length == 2) {
```
kaspersorensen: I'm thinking that instead of splitting this text twice (here and in parseAllJoinItems) then wouldn't it be better to not have this if-block and just always apply parseAllJoinItems(...)?

core/src/main/java/org/apache/metamodel/query/parser/FromItemParser.java
```diff
@@ -95,10 +104,32 @@ private FromItem parseTableItem(String itemToken) {
...
+        for(String token : joinsList) {
+            leftFromItem = parseJoinItem(leftFromItem, token, fromItems);
+        }
+        return leftFromItem;
+    }
 
     // this method will be documented based on this example itemToken: FOO f
     // INNER JOIN BAR b ON f.id = b.id
-    private FromItem parseJoinItem(final String itemToken) {
+    private FromItem parseJoinItem(final FromItem leftFromItem,final String itemToken,Set<FromItem> fromItems) {
```
kaspersorensen: Very minor, but I am missing some spaces after the commas :-)

core/src/main/java/org/apache/metamodel/query/parser/FromItemParser.java
```diff
@@ -136,16 +169,17 @@ private FromItem parseJoinItem(final String itemToken) {
             // rightPart = "b.id"
             final String rightPart = onClause.substring(indexOfEquals + 1).trim();
 
-            leftOn[i] = findSelectItem(leftPart, leftSide, rightSide);
-            rightOn[i] = findSelectItem(rightPart, leftSide, rightSide);
+            leftOn[i] = findSelectItem(leftPart, fromItems.toArray(new FromItem[fromItems.size()]), rightSide);
+            rightOn[i] = findSelectItem(rightPart, fromItems.toArray(new FromItem[fromItems.size()]), rightSide);
```
kaspersorensen: Let's only create that array once and reuse it for both calls to findSelectItem(...)

core/src/main/java/org/apache/metamodel/query/parser/FromItemParser.java
```diff
@@ -136,16 +169,17 @@ private FromItem parseJoinItem(final String itemToken) {
...
-        final FromItem result = new FromItem(joinType, leftSide, rightSide, leftOn, rightOn);
+        
+        final FromItem leftItem = (leftFromItem != null) ? leftFromItem : leftSide;
+        final FromItem result = new FromItem(joinType, leftItem, rightSide, leftOn, rightOn);
         result.setQuery(_query);
         return result;
     }
 
-    private SelectItem findSelectItem(String token, FromItem leftSide, FromItem rightSide) {
+    private SelectItem findSelectItem(String token, FromItem[] leftSides, FromItem rightSide) {
```
kaspersorensen: Actually, wouldn't it almost be better to just pass in the whole Set here instead of an array + an instance of FromItem?

core/src/main/java/org/apache/metamodel/query/FromItem.java
```diff
@@ -267,6 +267,11 @@ private void appendJoinOnItem(StringBuilder sb, String sideAlias, SelectItem onI
             sb.append(superQueryAlias);
             return;
         }
+        
+        if(_join != null && _leftSide.getJoin() != null) {
```
kaspersorensen: If we do it like this, then I think we should have the same treatment also on the right side as we have on the left side.

narahari92:
The reason I'm passing array+instance of from item is, the instance is we know for sure right side item. If we just pass set(including right side item), there will be n square combinations we need to check. Or am I wrong?

core/src/main/java/org/apache/metamodel/query/FromItem.java
kaspersorensen: And actually this trick avoids/skips the sideAlias thing ... might be a bit too destructive. Do we _need_ to avoid the parenthesis or is it just cosmetic?

kaspersorensen:
Done with my review. Only minor remarks really. The only one that I am a little bit concerned with is the parenthesis hiding approach. This seems to maybe be too destructive.

narahari92:
Hi,

the reason for removing alias for onSelectItem(side Alias) is let's say we have a query like below:

Query query = context.parse("Select a.col2 from A a INNER JOIN B b ON a.col1 = b.col2 INNER JOIN C c ON a.col3 = c.col3");

If we parse it by replacing sideAlias of select Item then it gets translated as below:

"SELECT a.col2 from A a INNER JOIN B b on a.col1 = b.col2 INNER JOIN C c ON b.col3 = c.col3"
which is wrong and results in SQLException

narahari92:
And the reason I put that condition for only leftFromItem is since we parse from left to right, only left from item can have composite join. Can you think of a scenario where even rightFromItem can be composite. Is so I will add that also.

core/src/main/java/org/apache/metamodel/query/parser/FromItemParser.java
```diff
@@ -153,17 +183,23 @@ private SelectItem findSelectItem(String token, FromItem leftSide, FromItem righ
         if (result == null) {
             // fail over and try with the from items available in the join that
             // is being built.
-            final Query temporaryQuery = new Query().from(leftSide, rightSide);
```
kaspersorensen: What I meant was that here you could just pass in the whole set of from items, instead of the two here. Then it's not more combinations, but fewer I think. At least you would then not need the new for-loop below.

kaspersorensen:
The right item way is also possible because anyone can build such a query programmatically. In our use of MetaModel we mostly do that and don't use the parser. A bit cumbersome, but allows different components to influence the query plan that we have as a kind application of the visitor pattern.

kaspersorensen:
I don't think I understand what you're saying about the side alias... I honestly also cannot remember all the nitty gritty details ... Just feel that it's a bit dangerous to throw that alias away - it could be referred to elsewhere I guess.

narahari92:
We are definitely not getting rid of alias. Instead we are using the select item as is. i.e. select item column name or expression along with its alias. In the unit test case you can see all aliases present.

narahari92:
And I have done that change of passing entire set and with push it today.

kaspersorensen:
OK I'm gonna go with my gut feeling that this probably works better than before, although I think there are still uncovered corner cases. More tests could maybe reveal that but on the other hand, if it _is_ better than before then what's the problem :-P

+1 from me.

kaspersorensen:
I'll merge it in.

PR #21, opened 2015-04-28
Add LIKE operator support. Fixes #METAMODEL-136

albertostratio:


mongodb/src/main/java/org/apache/metamodel/mongodb/MongoDbDataContext.java
```diff
@@ -408,7 +414,7 @@ private void convertToCursorObject(BasicDBObject query, FilterItem item) {
     private String getOperatorName(FilterItem item) {
         final String operatorName;
         switch (item.getOperator()) {
-        case EQUALS_TO:
+        case EQUALS_TO: case LIKE:
```
kaspersorensen: looks strange without a linebreak here I think.

mongodb/src/main/java/org/apache/metamodel/mongodb/MongoDbDataContext.java
```diff
@@ -435,6 +441,19 @@ private String getOperatorName(FilterItem item) {
...
+        String operandAsString = operand.toString();
+        StringBuilder operandAsRegExp = new StringBuilder(replaceWildCardLikeChars(operandAsString));
+        if (!operandAsString.startsWith("%") && operandAsString.contains("%")) {
+            operandAsRegExp = operandAsRegExp.insert(0, "^");
+        }
+        return Pattern.compile(operandAsRegExp.toString(), Pattern.CASE_INSENSITIVE);
+    }
+
+    private String replaceWildCardLikeChars(String operand) {
+        return operand.replace("%","");
```
kaspersorensen: Shouldn't this replace with something like ```.*```? Or else it seems to me it is just removing the % symbol?

mongodb/src/main/java/org/apache/metamodel/mongodb/MongoDbDataContext.java
albertostratio: I agree, will add a linebreak.

mongodb/src/main/java/org/apache/metamodel/mongodb/MongoDbDataContext.java
albertostratio: Yes, you are right, is just removing the %. I think we don't need to add more complexity to the regular expression, when we'd receive a string like '%MALE%' we just need to create a pattern like this (removing the %):
```
Pattern j = Pattern.compile("MALE", Pattern.CASE_INSENSITIVE);
```
Whereas when we'd receive a string like 'MALE%' the pattern should be (removing % and prepend ^):
```
Pattern j = Pattern.compile("^MALE", Pattern.CASE_INSENSITIVE);
```
I think these are the only cases we should care about. Both are working fine as tested in the test I've added.

What do you think?

kaspersorensen: I think we also need to cater to cases like ```WHERE name LIKE '% Alberto % Stratio %'```

albertostratio: Good point! I agree we should take care of cases like that one.

albertostratio:
I agree, we should be using replaceAll instead of replace but I think we need to keep the if in the turnOperandIntoRegExp method to prepend "^" when the like operand is like "MALE%" (pick up just elements starting with MALE). If we remove this condition we will pick up ALWAYS every string with the MALE substring in it and this is not what we want.

mongodb/src/main/java/org/apache/metamodel/mongodb/MongoDbDataContext.java
```diff
@@ -435,6 +442,19 @@ private String getOperatorName(FilterItem item) {
...
     }
 
+    private Pattern turnOperandIntoRegExp(Object operand) {
+        String operandAsString = operand.toString();
+        StringBuilder operandAsRegExp = new StringBuilder(replaceWildCardLikeChars(operandAsString));
+        if (!operandAsString.startsWith("%") && operandAsString.contains("%")) {
+            operandAsRegExp = operandAsRegExp.insert(0, "^");
+        }
+        return Pattern.compile(operandAsRegExp.toString(), Pattern.CASE_INSENSITIVE);
+    }
```
LosD: Isn't this missing the ```LIKE '%xxxxxx'``` case? I'd imagine a regex with $ in the end in that case, so that ```xxxxxxy``` wouldn't match, while ```yxxxxxx``` would.

But do we even need all these special cases? I would guess that always adding ^ and $ would always make it work? A few examples:
```LIKE 'xxxxxx'``` would become ```/^xxxxxxx$/``` which would only match ```xxxxxx```.
```LIKE '%xxxxxx'``` would become ```/^.*xxxxxx$/``` which would match anything ending with ```xxxxxx```.
```LIKE 'xxxxxx%'``` would become ```/^xxxxxx.*$/``` which would match anything starting with ```xxxxxx```.

Comment (albertostratio) [inline mongodb/src/main/java/org/apache/metamodel/mongodb/MongoDbDataContext.java]: Really good point in here! I've just modified the code to always prepend ^ and append $ and the tests are all passing OK. Thank you for the comment! I'll be pushing the changes right now...

**Comment by albertostratio** [comment]
Changes suggested by @LosD implemented. Thank you again!

File: mongodb/src/test/java/org/apache/metamodel/mongodb/MongoDbDataContextTest.java [hunk 4]
```diff
@@ -490,4 +490,59 @@ public void run(UpdateCallback callback) {
...
+            assertTrue(ds3.next());
+            assertFalse(ds3.next());
+            assertTrue(ds4.next());
+            assertTrue(ds4.next());
+            assertFalse(ds4.next());
+        } finally {
+            ds1.close();
+            ds2.close();
+            ds3.close();
+        }
```
LosD: Pretty minor, but ds4 isn't closed.

LosD:
Minor issue left, but looks good to me! :)

mongodb/src/test/java/org/apache/metamodel/mongodb/MongoDbDataContextTest.java
albertostratio: Good catch!

albertostratio:
Minor issue addressed. ping @LosD

LosD:
Looks super! Let's merge it.

LosD:
Though Travis seems to have failed?

LosD:
Okay, seems unrelated (ElasticSearch). @kaspersorensen also mentioned that we have had Travis problems with ES, so still good :)

albertostratio:
Yes, not sure what's going on with Travis but I've seen some builds of the PR "green"

LosD:
I'm pretty sure it's not a problem; both Java 7 & 8 completes successfully, and (much unlike the ES code) there is nothing in this code that should be affected by the JDK version.

kaspersorensen:
Looks good, I'll push it.

PR #23, opened 2015-05-14
Apache MetaModelMETAMODEL-140 Elasticsearch module unable to correctly process complex mappings

maphysics:
These changes are to address the issues in [1] .

I changed ElasticSearchDataContext.java detectTable to iterate through the
mappingMetaData and only parse the properties element.

I changed ElasticSearchMetaDataParser.java parse to split on "}," so that it
keeps the full description of a field with the field name, and adjusted
getMetaDataFieldTupeFromMetaDataField to handle the presence of extra
field descriptors and the edge case of the last field which will end with
a "}".

I changed ElasticSearchMetaDataParserTest.java to test the case when a field
has an extra descriptor by adding the index=not_analyzed tag to the
message field in the metaDataInfo string.

I also added a small test to ElasticSearchDataContextTest.java that makes sure that the changes addressing objects on the same level as the "properties" object are handled correctly.

[1] https://issues.apache.org/jira/browse/METAMODEL-140


maphysics:
I'm unable to reproduce this test failure locally.

kaspersorensen:
I see that the test is only failing on Java 6. So to reproduce locally you also need to run with Java 6 I guess. And maybe worth noting: In the J6 build profile we are using ElasticSearch version 0.90.13

kaspersorensen:
To track this in the stacktrace I can also recommend that you add ```.operationThreaded(false)``` to the requests that you send. That way you will not have a stack from another thread, but a stack that you can trace back to lines in your unittest.

kaspersorensen:
Looked a bit more into this and I simply cannot explain it ... Honestly I don't worry too much in this case because it seems to be just your test fixture that is failing, not the actual code change.

So unless you find a better way, I would accept if you do a dirty thing a la:

```
        if (Version.CURRENT.major == 0) {
            return;
        }
```

(in the test method)

kaspersorensen:
Can you apply the above fix in the unittest so that we can see Travis accepting it, and then I will happily merge. It's a good fix that you've made.

kaspersorensen:
Due to release timing I'm gonna go and make a fork of your branch @maphysics and then put in the unittest fix that I proposed ... Might get us a fix quicker and then we can hopefully get it into an upcoming bugfix release soon.

maphysics:
Thanks!  For tracking this.

PR #25, opened 2015-05-23
Metamodel143

narahari92:
Hi,

This is a fix for METAMODEL-143(Remove invalid characters from column names depending on different data context).

### DIFF
--- a/core/src/main/java/org/apache/metamodel/create/AbstractTableCreationBuilder.java
+++ b/core/src/main/java/org/apache/metamodel/create/AbstractTableCreationBuilder.java
@@ -18,6 +18,9 @@
  */
 package org.apache.metamodel.create;
 
+import java.util.Map;
+import java.util.Map.Entry;
+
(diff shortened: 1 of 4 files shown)


kaspersorensen:
My idea for designing this feature was more that we should enable the submodules themselves to apply a fitting naming strategy. In a polymorphic manner basically, meaning that the JDBC module knows best which column naming conventions must be followed for the various JDBC databases and maybe other modules has other conventions to apply to the columns of _their_ datastore types. So having a central ContextType defeats that design principle because then you cannot manage it in the submodules.

Another remark is that in the JDBC module it even varies amongst different database vendors. We have already plugged in different QueryRewriter classes for the purpose of handling SQL dialects. My suggestion is that in the JDBC module we add another method to the IQueryRewriter interface for normalizing the column names. In the abstract/base class it could then do generic stuff like replacing whitespace with underscore, and maybe in some subclasses it could do more. For instance I think that only a few databases support non-ASCII characters such as "æøåäöê" and so on.

PR #26, opened 2015-06-13
Added HdfsResource integration test

kaspersorensen:
Trying to make the story more complete wrt. the HdfsResource testability. I made it use our normal integration test approach of having properties in ${user.home}/metamodel-integrationtest-configuration.properties that drive the setup of the test.

### DIFF
--- a/example-metamodel-integrationtest-configuration.properties
+++ b/example-metamodel-integrationtest-configuration.properties
@@ -22,6 +22,15 @@
 #mongodb.databaseName=metamodel_test
 #mongodb.collectionName=my_collection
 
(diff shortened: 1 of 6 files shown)


kaspersorensen:
Just also pushed a change HdfsResource so that it no longer implements Closeable. This was never a good design I think and would make handling this Resource implementation much different from other implementations. I did some simple benchmarks of speed (using the integration test) and in 10 runs with both codebases it showed that the impact didn't matter (on the small file at least - but with larger files any difference should actually even out theoretically since the only potential penalty is around "chatty" interactions).

PR #28, opened 2015-06-17
Translate type DOUBLE to postgresql type "double precision". Jira 151

drexler42:
Fix for METAMODEL-151


kaspersorensen:
Looks good to me, and very nice integration test - you're thorough :-)

PR #29, opened 2015-06-17
Using an embedded Cassandra server for integration tests. Fixes #META…

ardlema:
…MODEL-154


kaspersorensen:
Looks like a unittest is failing on some Java versions (6 and 8): 

Failed tests: 
  CassandraDataContextTest.testSchemaAndSimpleQuery:73 expected:<[[counter, songs]]> but was:<[[songs, counter]]>

albertostratio:
Thank you for pointing this out. Unfortunately Java 6 & 8 versions builds are failing for different reasons.

Java 8 build is failing, as you said, because of a wrong test checking condition (pretty ease to change: we should not take into account the order of the tables) whereas Java 6 build is failing because the cassandra-unit library I added is not Java 6 compatible: java.lang.UnsupportedClassVersionError: org/cassandraunit/utils/EmbeddedCassandraServerHelper : Unsupported major.minor version 51.0 :(

kaspersorensen:
OK yeah, we saw that also recently with ElasticSearch that some part of it's API was not supported on Java 6... I guess we should exclude this particular test or test class if Java version is 6 then.

albertostratio:
@kaspersorensen I've excluded the particular test for Jdk 1.6 and also modify the flaky test

kaspersorensen:
LGTM!

albertostratio:
Nice! I'll push the changes

PR #30, opened 2015-06-17
METAMODEL-152

kaspersorensen:
This fixes the pretty critical caching issue we have where even a ```refreshSchemas()``` call does not clear the cache of the "main schema" in QueryPostprocessDataContext.

I simplified the QueryPostprocessDataContext a bit by not having any cache at all. That's because there already is a schema cache in it's superclass, AbstractDataContext.


albertostratio:
LGTM

kaspersorensen:
Cool. I'll merge.

PR #31, opened 2015-07-12
Apache Hive support

kaspersorensen:
This PR adds support for Apache Hive via the existing JDBC module.

To do this I had to introduce a new "isTransactional()" flag on IQueryRewriter since commit calls on Hive is not supported.

I also did some cleanup of Hadoop dependencies because I saw Hadoop version conflicts inbetween the "hadoop", "hbase" and now the "jdbc" modules. This caused the Hive connection to fail with class loading issues.

### DIFF
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcUpdateCallback.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcUpdateCallback.java
@@ -47,7 +47,7 @@ abstract class JdbcUpdateCallback extends AbstractUpdateCallback implements Upda
         super(dataContext);
         _dataContext = dataContext;
     }
-    
+
     protected abstract void closePreparedStatement(PreparedStatement preparedStatement);
 
     protected abstract void executePreparedStatement(PreparedStatement preparedStatement) throws SQLException;
@@ -63,10 +63,12 @@ abstract class JdbcUpdateCallback extends AbstractUpdateCallback implements Upda
     protected final Connection getConnection() {
         if (_connection == null) {
             _connection = getDataContext().getConnection();
-            try {
-                _connection.setAutoCommit(false);
-            } catch (SQLException e) {
-                throw JdbcUtils.wrapException(e, "disable auto-commit");
+            if (_dataContext.getQueryRewriter().isTransactional()) {
+                try {
+                    _connection.setAutoCommit(false);
+                } catch (SQLException e) {
+                    throw JdbcUtils.wrapException(e, "disable auto-commit");
+                }
             }
         }
         return _connection;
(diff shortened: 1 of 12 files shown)


jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcUpdateCallback.java
```diff
@@ -63,10 +63,12 @@ public void executePreparedStatement(PreparedStatement preparedStatement, boolea
...
         if (_connection == null) {
             _connection = getDataContext().getConnection();
-            try {
-                _connection.setAutoCommit(false);
-            } catch (SQLException e) {
-                throw JdbcUtils.wrapException(e, "disable auto-commit");
+            if (_dataContext.getQueryRewriter().isTransactional()) {
+                try {
+                    _connection.setAutoCommit(false);
+                } catch (SQLException e) {
```
albertostratio: With this change we are now setting the autocommit to false only when the isTransactional condition is true, is this what we want?

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcUpdateCallback.java
```diff
@@ -63,10 +63,12 @@ public void executePreparedStatement(PreparedStatement preparedStatement, boolea
     protected final Connection getConnection() {
         if (_connection == null) {
             _connection = getDataContext().getConnection();
-            try {
-                _connection.setAutoCommit(false);
-            } catch (SQLException e) {
-                throw JdbcUtils.wrapException(e, "disable auto-commit");
+            if (_dataContext.getQueryRewriter().isTransactional()) {
+                try {
```
albertostratio: With this change we are now setting the autocommit to false only when the isTransactional condition is true, is this what we want?

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcUpdateCallback.java
```diff
@@ -78,18 +80,20 @@ public final void close(boolean success) {
                 closePreparedStatement(_preparedStatement);
             }
 
-            try {
-                commitOrRollback(success);
+            if (_dataContext.getQueryRewriter().isTransactional()) {
+                try {
```
albertostratio: Same comment than above

albertostratio:
@kaspersorensen I've added some comments, please review.

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcUpdateCallback.java
kaspersorensen: At least it's what I meant :-) We only have Hive in our portfolio currently that is a "non transactional" JDBC driver. In the case of Hive it is not allowed to call these methods.

Do you have an underlying concern with this specifically?

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcUpdateCallback.java
kaspersorensen: Same comment as above :)

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcUpdateCallback.java
albertostratio: Ok got it! Thank you for clarifying, makes sense now.

albertostratio:
Thank you @kaspersorensen 

These changes LGTM

PR #32, opened 2015-07-17
METAMODEL-162

kaspersorensen:
This adds the getters and implements Serializable, as described by METAMODEL-162

https://issues.apache.org/jira/browse/METAMODEL-162


albertostratio:
LGTM

PR #33, opened 2015-07-17
METAMODEL-161

kaspersorensen:
Upgraded HBase dependency to version 1.1.1.

Tested it with the integration test on my setup and it worked well.

I also added some timeout and retry properties to HBaseConfiguration because otherwise a wrong setting might cause tests (and "real" connections for that matter) to keep retrying for ages (defaults are extremely high IMO).


albertostratio:
LGTM

PR #34, opened 2015-07-27
Refactoring to support scalar function

ardlema:
The current code only supports aggregate functions. In order to support more type of functions such as scalar we need to do some refactor.

The changes made in this PR include:

   - The FunctionType is now an interface which has static final references to the aggregate functions.
   - New AggregateFunction interface that contains the aggregation specific methods
   - New DefaultAggregateFunction which implements the AggregateFunction interface
   - ScalarFunction interface which is an skeleton where we should add the new scalar functions
   - FunctionTypeFactory that creates AggregateFunctions from its name


kaspersorensen:
Very shallow review from my phone (on holiday :-)) - wouldn't it be better with one implementation for each function, e.g. Sum, avg etc., instead of one implementation to rule all (and thus limit the feeling of pluggability)

albertostratio:
@kaspersorensen Do you mean something like this?:

```
public class SumAggregateFunction extends DefaultAggregateFunction<Long> implements AggregateFunction {

    public AggregateBuilder<Long> build() {
        return new SumAggregateBuilder();
    }
}
```

albertostratio:
@kaspersorensen I've created an implementation for each function

core/src/main/java/org/apache/metamodel/query/AverageAggregateBuilder.java
```diff
@@ -23,7 +23,7 @@
 /**
  * Aggregate builder for the {@link FunctionType#AVG} function
  */
-final class AverageAggregateBuilder extends AbstractNumberAggregateBuilder<Double> {
+final class AverageAggregateBuilder<T> extends AbstractNumberAggregateBuilder<Double> {
```
kaspersorensen: Why the new type parameter here?

core/src/main/java/org/apache/metamodel/query/parser/SelectItemParser.java
```diff
@@ -113,10 +110,10 @@ public SelectItem findSelectItem(String expression) throws MultipleSelectItemsPa
         final int startParenthesis = expression.indexOf('(');
         if (startParenthesis > 0 && expression.endsWith(")")) {
             String functionName = expression.substring(0, startParenthesis);
-            function = FunctionType.get(functionName.toUpperCase());
+            function = FunctionTypeFactory.get(functionName.toUpperCase());
             if (function != null) {
                 expression = expression.substring(startParenthesis + 1, expression.length() - 1).trim();
-                if (function == FunctionType.COUNT && "*".equals(expression)) {
+                if (function.toString().equals("COUNT") && "*".equals(expression)) {
```
kaspersorensen: IMO it is better to compare the function object than comparing tostring. And I think we discussed having the builtin functions on the FunctionType interface still.

kaspersorensen:
Looking at the Function and FunctionType interface I see too much overlap IMO. And methods like build and evaluate seem to only apply to aggregations. So maybe delete the new Function interface? And move methods to AggregateFunction interface? 

Sorry for the very short remarks. On worst Internet connection ever.

core/src/main/java/org/apache/metamodel/query/AverageAggregateBuilder.java
ardlema: Good catch! this is completely unnecessary, I'll get rid of it

core/src/main/java/org/apache/metamodel/query/parser/SelectItemParser.java
ardlema: Change done!

ardlema:
@kaspersorensen changes addressed.

With regard to the Function interface I thought that this could be a good place to define all the generic methods for the different type of functions. Right now I agree with you: we can just move the getExpectedColumnType to the AggregateFunctionInterface but when we'd add the ScalarFunctionInterface we might end up duplicating generic functions in both interfaces.

core/src/main/java/org/apache/metamodel/query/FunctionType.java
```diff
@@ -18,105 +18,22 @@
...
-        return builder.getAggregate();
-    }
+    public static final AggregateFunction COUNT = new CountAggregateFunction();
+    public static final AggregateFunction AVG = new AverageAggregateFunction();
+    public static final AggregateFunction SUM = new SumAggregateFunction();
+    public static final AggregateFunction MAX = new MaxAggregateFunction();
+    public static final AggregateFunction MIN = new MinAggregateFunction();
 
-    public abstract AggregateBuilder<?> build();
+    public AggregateBuilder<?> build();
```
kaspersorensen: I think this method is implementation oriented and should maybe then be kept out of the interface. Or else the evaluate method should - they kinda serve the same purpose.

And they are both aggregation function oriented so none of them should be here but if anywhere in the AggregateFunction interface.

core/src/main/java/org/apache/metamodel/query/DefaultAggregateFunction.java
```diff
@@ -0,0 +1,64 @@
...
+
+import org.apache.metamodel.schema.ColumnType;
+import org.apache.metamodel.util.AggregateBuilder;
+
+/**
+ * Implementation of the {@link org.apache.metamodel.query.AggregateFunction}.
+ */
+public abstract class DefaultAggregateFunction<T>  {
+
+    String functionType;
```
kaspersorensen: This field isn't so nicely exposed IMO. It's both mutable and unencapsulated. So it requires quite some studying by the subclass author to understand its meaning.

I could imagine it better using an abstract method or as a private final field that is set via the constructor.

core/src/main/java/org/apache/metamodel/query/FunctionType.java
ardlema: I agree with you Kasper, we've got a problem in here though.

We are calling the build method from the MetaModelHelper class and this guy is picking up the FunctionType interface from the getFunction method of SelectItem. It means that if we move these methods to the AggregateFunction interface we should change the SelectItem class to make it implementation-oriented and then we need to duplicate constructor methods and have two variables for AggregateFunctionType and ScalarFunctionType interfaces.

Not sure how to solve this issue... any ideas?

core/src/main/java/org/apache/metamodel/query/DefaultAggregateFunction.java
albertostratio: You're absolutely right, I'll modify it

core/src/main/java/org/apache/metamodel/query/FunctionType.java
kaspersorensen: How about we add two methods to selectitem: getAggregateFunction and getScalarFunction. In Selectitem we can then do the instanceof check and otherwise return null.

albertostratio: That approach was my first thought but there are a lot of calls to the getFunction method scattered all around the code (FetchSizeCalculator, JdbcDataSet, QueryPostProcessDataContext, ConvertedDataSetInterceptor...). If we get rid of the getFunction method we should modify a lot of classes and ideally I do not want to do it for this refactoring :(

kaspersorensen: But I would still retain the existing method too. Just have the two new getters implemented to check if function is instance of e.g. AggregateFunction and then return it if so.

albertostratio:
@kaspersorensen I've added a new getter to SelectItem that will return an AggregateFunction interface. Could you please, check it out again? thx

core/src/main/java/org/apache/metamodel/query/AggregateFunction.java
```diff
@@ -0,0 +1,35 @@
...
+ * related to the AggregateBuilder.
+ *
+ */
+public interface AggregateFunction extends FunctionType {
+
+    Object evaluate(Iterable<?> values);
+
+    public AggregateBuilder<?> build();
+
+    public Object evaluate(Object... values);
```
kaspersorensen: Sorry to keep bringing up these methods :-) seems the build method is the superior one, so I would keep only that. But rename it to something more explanatory, like createAggregateBuilder.

core/src/main/java/org/apache/metamodel/query/Function.java
```diff
@@ -0,0 +1,27 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.metamodel.query;
+
+import org.apache.metamodel.schema.ColumnType;
+
+public interface Function {
```
kaspersorensen: I would still delete this new class and just move the method (and a getFunctionName method for getting "AVG" etc.) to FunctionType.

core/src/main/java/org/apache/metamodel/query/DefaultAggregateFunction.java
```diff
@@ -0,0 +1,65 @@
...
+     */
+    public Object evaluate(Object... values) {
+        AggregateBuilder<?> builder = build();
+        for (Object object : values) {
+            builder.add(object);
+        }
+        return builder.getAggregate();
+    }
+
+    public abstract String getFunctionType();
```
kaspersorensen: This should be defined in the FunctionType interface, and I would give it the name GetFunctionName.

albertostratio:
@kaspersorensen changes addressed

core/src/main/java/org/apache/metamodel/query/AggregateFunction.java
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Interface that contains the aggregation specific methods
+ * related to the AggregateBuilder.
+ *
+ */
+public interface AggregateFunction extends FunctionType {
+
+    public AggregateBuilder<?> createAggregateBuilder();
+
+    public Object evaluate(Object... values);
```
kaspersorensen: Not blocking for me right now (since we can easily add it), but I would prefer some javadoc on those methods. Especially to also point out that the evaluate method is just a shorthand for creating an aggregate builder, adding all the values and then calculating the value. Maybe that even warrants removing the method, but maybe you don't want to do that as part of this refactoring.

core/src/main/java/org/apache/metamodel/query/AverageAggregateFunction.java
```diff
@@ -0,0 +1,35 @@
...
+ * under the License.
+ */
+package org.apache.metamodel.query;
+
+import org.apache.metamodel.schema.ColumnType;
+import org.apache.metamodel.util.AggregateBuilder;
+
+public class AverageAggregateFunction extends DefaultAggregateFunction<Double> implements AggregateFunction {
+
+    public String getFunctionName() { return "AVG"; }
```
kaspersorensen: Normally we always have "regular" line breaks inside the brackets. This also applies to the other implementations which has the same code style pattern in this regard.

core/src/main/java/org/apache/metamodel/query/DefaultAggregateFunction.java
```diff
@@ -0,0 +1,63 @@
...
+    public Object evaluate(Object... values) {
+        AggregateBuilder<?> builder = createAggregateBuilder();
+        for (Object object : values) {
+            builder.add(object);
+        }
+        return builder.getAggregate();
+    }
+
+    @Override
+    public String toString() { return getFunctionName(); }
```
kaspersorensen: and here too :)

core/src/main/java/org/apache/metamodel/query/SelectItem.java
```diff
@@ -211,6 +211,13 @@ public FunctionType getFunction() {
         return _function;
     }
 
+    public AggregateFunction getAggregateFunction() {
+        if (_function instanceof AggregateFunction)
+            return (AggregateFunction) _function;
+        else
```
kaspersorensen: I'm also missing some brackets here to comply with our normal code style.

core/src/main/java/org/apache/metamodel/query/ScalarFunction.java
```diff
@@ -0,0 +1,25 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.metamodel.query;
+
+
+public interface ScalarFunction extends FunctionType {
```
kaspersorensen: Javadoc would be good here. The word "scalar" is probably hard to digest for many people, so a bit explanation will help.

core/src/main/java/org/apache/metamodel/query/parser/SelectItemParser.java
```diff
@@ -113,10 +110,10 @@ public SelectItem findSelectItem(String expression) throws MultipleSelectItemsPa
         final int startParenthesis = expression.indexOf('(');
         if (startParenthesis > 0 && expression.endsWith(")")) {
             String functionName = expression.substring(0, startParenthesis);
-            function = FunctionType.get(functionName.toUpperCase());
+            function = FunctionTypeFactory.get(functionName.toUpperCase());
             if (function != null) {
                 expression = expression.substring(startParenthesis + 1, expression.length() - 1).trim();
-                if (function == FunctionType.COUNT && "*".equals(expression)) {
+                if (function instanceof CountAggregateFunction && "*".equals(expression)) {
```
kaspersorensen: We actually have SelectItem.isCountAllItem(...) to do this - maybe better to reuse.

core/src/main/java/org/apache/metamodel/query/DefaultAggregateFunction.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.metamodel.util.AggregateBuilder;
+
+/**
+ * Implementation of the {@link org.apache.metamodel.query.AggregateFunction}.
+ */
+public abstract class DefaultAggregateFunction<T> implements AggregateFunction {
+
+    public abstract AggregateBuilder<T> createAggregateBuilder();
+
+    public abstract ColumnType getExpectedColumnType(ColumnType type);
```
kaspersorensen: I cannot find out where this method is used for anything. So either I think we should not have it (will save a bunch of lines of code) or else it should maybe be part of the AggregateFunction interface?

core/src/main/java/org/apache/metamodel/query/FunctionTypeFactory.java
```diff
@@ -0,0 +1,43 @@
...
+        if (functionName.equals("COUNT")) {
+            return new CountAggregateFunction();
+        } else if (functionName.equals("AVG")) {
+            return new AverageAggregateFunction();
+        } else if (functionName.equals("SUM")) {
+            return new SumAggregateFunction();
+        } else if (functionName.equals("MAX")) {
+            return new MaxAggregateFunction();
+        } else if (functionName.equals("MIN")) {
+            return new MinAggregateFunction();
```
kaspersorensen: I guess it would be nicer to just return the single instances that we know from FunctionType.MIN, FunctionType.COUNT here etc...

kaspersorensen:
Added a bunch of new remarks. I have been pretty meticulous in checking the PR mostly because it is an API change, so I hope it's encouraging and does not feel like critisism. It looks quite good now, and if you agree on the remarks I have left then I will give it my +1 for sure :)

core/src/main/java/org/apache/metamodel/query/AggregateFunction.java
albertostratio: I'll add it as part of this PR

core/src/main/java/org/apache/metamodel/query/AverageAggregateFunction.java
albertostratio: Intellij faults ;) I'll add the line breaks. Good catch, thx!

core/src/main/java/org/apache/metamodel/query/ScalarFunction.java
albertostratio: Completely agree

core/src/main/java/org/apache/metamodel/query/parser/SelectItemParser.java
albertostratio: Good point, I'll modify it. Thx!

core/src/main/java/org/apache/metamodel/query/DefaultAggregateFunction.java
albertostratio: This method is being used in the getExpectedColumnType of SelectItem. I'd declare it as part of the AggregateFunction interface and then implement it in the DefaultAggregateFunction class.

albertostratio:
@kaspersorensen Thank you for all the remarks, I've implemented all of them apart from the one related to the isCountAllItem function (see comment inline above).

As you said we are performing some important changes in the API so we must be careful with them.

core/src/main/java/org/apache/metamodel/query/parser/SelectItemParser.java
kaspersorensen: Good point, I missed that so I agree!

kaspersorensen:
Ok I am now 100 happy with the PR :-)

Since this change will break runtime compatibility (but NOT compile time compatibility, which is good) I think it warrants a bump in versioning to make this into MM 4.4.0. Then we can also stop supporting Java 6 in that version.

So actually I would prefer not to merge it to master just yet, but get a final MM 4.3.x release done first. Maybe start a vote on that Alberto?

albertostratio:
Sounds like a plan. Looking forward to getting rid of Java 6 support ;)

Would you mind starting the vote?

kaspersorensen:
I'll merge it in now :)

kaspersorensen:
I also added METAMODEL-178 which this fixes.

PR #35, opened 2015-07-29
Adding hadoop and hbase modules to the MetaModel-full artifact.

tomaszguzialek:
Also sorting the dependencies alphabetically to find out easier what is
missing in the future.

Resolves METAMODEL-104.


kaspersorensen:
LGTM

PR #36, opened 2015-07-31
Changed HdfsResource implemention so that Writing to Hadoop is possible.

ClaudiaPHI:
This is a part of the story from DataCleaner https://github.com/datacleaner/DataCleaner/issues/494. 

In Hadoop the default replication factor is 3. After a file is created and closed, it become immutable, because of the replicas. Basically, it means that append is not allowed.  

With this change I made it possible to write to Hadoop a CSV file by setting in the configuration the replication factor 1.

However, writing is very slow because the "CsvUpdateCallback"  class appends row by row. It can fail anytime if the Hadoop node is leased or other type of failures occur.

### DIFF
--- a/hadoop/src/main/java/org/apache/metamodel/util/HdfsResource.java
+++ b/hadoop/src/main/java/org/apache/metamodel/util/HdfsResource.java
@@ -46,6 +46,21 @@ public class HdfsResource implements Resource, Serializable {
     private final int _port;
     private final String _filepath;
     private Path _path;
+    private boolean _overwriteIfExists = false;
+    private Configuration _hadoopConfiguration;
+
+    /**
+     * Creates a {@link HdfsResource}
+     * 
+     * @param url
+     *            a URL of the form: hdfs://hostname:port/path/to/file
+     * @param overwriteIfExits
+     *            If true overwrites the file if exists. Default is false
+     */
+    public HdfsResource(String url, boolean overwriteIfExists) {
+        this(url);
+        _overwriteIfExists = overwriteIfExists;
+    }
 
     /**
      * Creates a {@link HdfsResource}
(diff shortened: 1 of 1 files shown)


hadoop/src/main/java/org/apache/metamodel/util/HdfsResource.java
```diff
@@ -338,4 +386,26 @@ public boolean equals(Object obj) {
...
+
+    public void replicateFile(int replicationFactor) {
+        try {
+            final FileSystem fs = getHadoopFileSystem();
+            fs.setReplication(getHadoopPath(), (short) replicationFactor);
+        } catch (IOException e) {
+            throw new MetaModelException("Could not create replicas for Hdfs file  : " + getHadoopFileSystem() + " "
+                    + e.getMessage(), e);
+        }
+    }
```
LosD: Shouldn't this update the Hadoop configuration as well?

LosD:
It's not really ideal, but I guess it is the best we can do without close() in Resource (I'm not entirely sure why resources aren't closeable, or bufferable in some other way... Right now we open and close streams for every single write, which seems like a bad thing for most resource types).

kaspersorensen:
A few thoughts/notes:

I think actually the CSV update callback has some FileResource specific hacks to avoid always appending again and again. Might be you find the better root cause here!

To LosD: Resource is not itself closeable just like File is not. But the streams that you might get by invoking methods on the Resource is obviously closeable.

LosD:
In many (most?) cases, I. believe. it. is. pretty. inefficient. to. have. to. close. the. stream. for. every. single. row. In HDFS it is crippling. 

It's not a problem for normal File operations, as you can easily buffer before writing a block, but unless I misunderstood @cludiaPHI's investigations, we do not have that luxury in Metamodel, as rows come in one at a time, and there's no way to know when it ends. We can of course do the instanceof hack in the CSV writer, but do we want users of the Resource interface to be handicapped by default?

I might have misunderstood something, but as far as I can see, HDFS is just an extreme example of a general problem.
-- 
Med venlig hilsen 
Dennis Du Krøger

kaspersorensen:
I am absolutely with you on that. I did not mean to suggest that the instanceof hack was a beautiful thing. Rather that it hints at an underlying issue that we should maybe rather fix than setting replication levels?

kaspersorensen:
To be more specific: it seems that we for files save a writer in CsvUpdateCallback. I think that's because we don't have a way to generically create that for any resource. Resource has a read() method that returns InputStream but not a write() method that returns OutputStream. Maybe we should simply add that? Then I think we could do that same trick on any resource implementation.

LosD:
That seems like a good plan. I'll take a look.

PR #37, opened 2015-08-07
Adds ability to read folders to FileResource and HdfsResource

LosD:
This will allow FileResource and HdfsResource to use a folder as a complete resource. It will skip any subfolder and only use files in the folder. Before opening files, they will be sorted them alphabetically by name (actually it uses the natural ordering of the File/FileStatus object, but that is the path/url sorted alphabetically).

Fixes METAMODEL-163

### DIFF
--- a/core/src/main/java/org/apache/metamodel/util/AbstractDirectoryInputStream.java
+++ b/core/src/main/java/org/apache/metamodel/util/AbstractDirectoryInputStream.java
@@ -1,0 +1,97 @@
+package org.apache.metamodel.util;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+public abstract class AbstractDirectoryInputStream<T> extends InputStream {
+    protected T[] _files;
+    private int _currentFileIndex = -1;
+    private InputStream _currentInputStream;
+
+
+    @Override
+    public int read(final byte[] b, final int off, final int len) throws IOException {
+        if (_currentInputStream != null) {
+            final int byteCount = _currentInputStream.read(b, off, len);
+            if (byteCount > 0) {
+                return byteCount;
+            }
+        }
+
+        if (!openNextFile()){
+            return -1; // No more files.
+        }
+
+        return read(b, off, len);
+    }
+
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/metamodel/util/AbstractDirectoryInputStream.java
```diff
@@ -0,0 +1,97 @@
+package org.apache.metamodel.util;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
```
kaspersorensen: Normally we put the license header above the package declaration.

core/src/main/java/org/apache/metamodel/util/AbstractDirectoryInputStream.java
```diff
@@ -0,0 +1,97 @@
...
+        if(count < 0){
+            return -1;
+        }
+        return (int) b[0];
+    }
+
+    @Override
+    public int available() throws IOException {
+        if(_currentInputStream != null){
+            return _currentInputStream.available();
```
kaspersorensen: What about the potential next stream here?

core/src/main/java/org/apache/metamodel/util/AbstractDirectoryInputStream.java
LosD: Errr... Whoops!

core/src/main/java/org/apache/metamodel/util/FileResource.java
```diff
@@ -19,16 +19,43 @@
...
 import java.io.Serializable;
+import java.util.Arrays;
+import java.util.Comparator;
 
 /**
  * {@link File} based {@link Resource} implementation.
  */
 public class FileResource implements Resource, Serializable {
 
+    private class DirectoryInputStream extends AbstractDirectoryInputStream<File> {
```
kaspersorensen: Hehe very nitty gritty, but since this nested class is non-static, shouldn't it be declared after static fields?

core/src/main/java/org/apache/metamodel/util/AbstractDirectoryInputStream.java
LosD: Check the JavaDoc for available()... It's only how many is left until a blocking operation. Opening a whole new stream or file is certainly blocking :)

kaspersorensen:
I like what I see. But just had some cosmetic remarks :-)

core/src/main/java/org/apache/metamodel/util/FileResource.java
LosD: Hmmm... I normally put inline classes and enums as the first elements, but that's a matter of taste. I can move it if you prefer it after static fields.

LosD:
Argh! No nio in Java 6... Will fix tomorrow

core/src/main/java/org/apache/metamodel/util/AbstractDirectoryInputStream.java
kaspersorensen: Got it :-)

core/src/main/java/org/apache/metamodel/util/FileResource.java
kaspersorensen: Nah this is fine, I think you're right. I guess I just almost never make non-static inner classes but here it works quite ok.

tomaszguzialek:
The question is if we still want to support Java 6? I remember a thread on the mailing list where it was decided to craft one more release with Java 6 and then abandon it. I think we had one already...

LosD:
This should fix the tests. Let's hope Travis agrees!

tomaszguzialek:
Travis is happy, so I will merge it.

PR #38, opened 2015-08-12
METAMODEL-164: Fixed and refactored ES mapping parser to use real map.

kaspersorensen:
Previously the parser was (by accident I think) using the toString()
representation of that same map which made it very vulnerable towards
parsing issues.

Fixes METAMODEL-164


LosD:
LGTM!

PR #39, opened 2015-08-15
Feature/metamodel 170 drop java6 support

kaspersorensen:
My attempt at fixing METAMODEL-170: Drop Java 6 support.

Do not merge before the release process of MetaModel 4.3.6 has finished completely.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -164,7 +164,7 @@ under the License.
 	<build>
 		<plugins>
 			<plugin>
-				<!-- Ensures java 6 compatibility -->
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -168,8 +168,8 @@ under the License.
 				<groupId>org.apache.maven.plugins</groupId>
```
ardlema: Minor nitpick: the comment should now say

kaspersorensen: Fixed.

ardlema: Thank you ;)

PR #42, opened 2015-08-18
METAMODEL-171: Fixed

kaspersorensen:
Fixes the Cassandra unittest which is failing sometimes on my machine. The machine is getting a bit old and slow, so maybe it was a race condition. At least changing the test so that the Cassandra cluster is only created and closed once fixes the issue.


albertostratio:
LGTM

PR #43, opened 2015-08-24
METAMODEL-172: Adding Date parsing to ES module

ardlema:
When working with ElasticSearch indexes that contains Date types the API is returning Strings instead of proper Date objects. We found out this problem querying an index with a time-zone format field.

Changes made in this PR:

   - Added time-zone format dates support in the TimeComparator class
   - Added a string to date direct conversion through the TimeComparator class in the ElasticSearchUtils class that will convert the string from ES into a proper Date (when the format will be supported by Metamodel)
   - Added a new unit test for the ElasticSearchUtils class

### DIFF
--- a/core/src/test/java/org/apache/metamodel/util/TimeComparatorTest.java
+++ b/core/src/test/java/org/apache/metamodel/util/TimeComparatorTest.java
@@ -47,7 +47,7 @@ public class TimeComparatorTest extends TestCase {
 	}
 
 	public void testToDate() throws Exception {
-		DateFormat dateFormat = DateUtils
+/*		DateFormat dateFormat = DateUtils
 				.createDateFormat("yyyy-MM-dd HH:mm:ss.SSS");
 
 		assertEquals("2008-11-04 00:00:00.000",
@@ -65,10 +65,10 @@ public class TimeComparatorTest extends TestCase {
 
 		assertEquals("2000-12-31 02:30:05.100",
 				dateFormat.format(TimeComparator
-						.toDate("2000-12-31 02:30:05.100")));
+						.toDate("2000-12-31 02:30:05.100")));*/
 
-        assertEquals("Fri Jan 04 15:55:51 CET 2013",
-                TimeComparator.toDate("2013-01-04T15:55:51.217+01:00").toString());
+        assertEquals("Fri Jan 04 16:55:51 CET 2013",
+                TimeComparator.toDate("2013-01-04T15:55:51.217+00:00").toString());
 	}
 
 	public void testToDateOfDateToString() throws Exception {
--- a/core/src/test/java/org/apache/metamodel/util/TimeComparatorTest.java
+++ b/core/src/test/java/org/apache/metamodel/util/TimeComparatorTest.java
@@ -47,7 +47,7 @@ public class TimeComparatorTest extends TestCase {
 	}
 
 	public void testToDate() throws Exception {
-/*		DateFormat dateFormat = DateUtils
+		DateFormat dateFormat = DateUtils
 				.createDateFormat("yyyy-MM-dd HH:mm:ss.SSS");
 
 		assertEquals("2008-11-04 00:00:00.000",
@@ -65,10 +65,7 @@ public class TimeComparatorTest extends TestCase {
 
 		assertEquals("2000-12-31 02:30:05.100",
 				dateFormat.format(TimeComparator
-						.toDate("2000-12-31 02:30:05.100")));*/
-
-        assertEquals("Fri Jan 04 16:55:51 CET 2013",
-                TimeComparator.toDate("2013-01-04T15:55:51.217+00:00").toString());
+						.toDate("2000-12-31 02:30:05.100")));
 	}
 
 	public void testToDateOfDateToString() throws Exception {
--- a/core/src/test/java/org/apache/metamodel/util/TimeComparatorTest.java
+++ b/core/src/test/java/org/apache/metamodel/util/TimeComparatorTest.java
@@ -67,9 +67,8 @@ public class TimeComparatorTest extends TestCase {
 				dateFormat.format(TimeComparator
 						.toDate("2000-12-31 02:30:05.100")));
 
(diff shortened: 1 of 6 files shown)


core/src/test/java/org/apache/metamodel/util/TimeComparatorTest.java
```diff
@@ -66,6 +66,9 @@ public void testToDate() throws Exception {
 		assertEquals("2000-12-31 02:30:05.100",
 				dateFormat.format(TimeComparator
 						.toDate("2000-12-31 02:30:05.100")));
+
+        assertEquals("Fri Jan 04 15:55:51 CET 2013",
+                TimeComparator.toDate("2013-01-04T15:55:51.217+01:00").toString());
```
kaspersorensen: This seems to fail in Travis.

elasticsearch/src/main/java/org/apache/metamodel/elasticsearch/ElasticSearchUtils.java
```diff
@@ -44,8 +47,17 @@ public static Row createRow(Map<String, Object> sourceMap, String documentId, Da
                 values[i] = documentId;
             } else {
                 Object value = sourceMap.get(column.getName());
-                
-                values[i] = value;
+
+                if (column.getType() == ColumnType.DATE) {
+                    Date valueToDate = TimeComparator.toDate(value);
```
kaspersorensen: Maybe it's better that we put in the ES module the time format that ES uses. I guess the TimeComparator is a bit more general purpose while I imagine (?) that ES is always using one specific date pattern.

albertostratio:
@kaspersorensen as per your suggestion I moved the date parsing to the ES module instead of using the TypeComparator class

elasticsearch/src/main/java/org/apache/metamodel/elasticsearch/ElasticSearchDateConverter.java
```diff
@@ -0,0 +1,39 @@
...
+ * proper java Dates.
+ */
+final class ElasticSearchDateConverter {
+
+    public static Date tryToConvert(String dateAsString) {
+        try {
+            SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSX");
+            return dateFormat.parse(dateAsString);
+        } catch (ParseException e) {
+            return null;
```
kaspersorensen: I don't like to swallow an exception like this since it may give the wrong impression. Are we sure that this date format is always used by ES? If so I would throw an exception in this case. Or else at least make a loud warning or try the general purpose date converter or so.

albertostratio: There is plenty of built-in date formats in ES, see: https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-date-format.html

My intention with this PR is provide a place where we can add the different formats for ES dates on-demand (when someone using our ES connector and working with a date format that we are not currently supporting we'll be able to easily include that format)

Calling the general purpose date converter after trying the ES specific ones might be a good idea 'cos we get all the existing MM date parsing conversions for free.

elasticsearch/src/main/java/org/apache/metamodel/elasticsearch/ElasticSearchDateConverter.java
```diff
@@ -0,0 +1,41 @@
...
+ * proper java Dates.
+ */
+final class ElasticSearchDateConverter {
+
+    public static Date tryToConvert(String dateAsString) {
+        try {
+            SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSX");
+            return dateFormat.parse(dateAsString);
+        } catch (ParseException e) {
+            return TimeComparator.toDate(dateAsString);
```
kaspersorensen: It's an improvement to add it, but I still would like to ask out (also to others?) what is best to do here ... In case we cannot parse the date format, then I tend to suggest that we simply return the dateAsString (so the method must return Object). Reason is that otherwise we might simply throw away date and that can give users the wrong impression, thinking that there is no date value in the data itself and fueling wrong application-logic then. Another approach is to throw an exception. In such a case at least the user knows that something went wrong. What I don't like is to let the user think everything is fine but that the data is null.

LosD: Doesn't ES have a utility method to convert the date string to a Date object or milliseconds since 1970 or something else, using the format of a field? It doesn't really make sense that we allow any other format than what is dictated by the schema.

albertostratio: @kaspersorensen if the TimeComparator method can't parse the date format and returns null we are just setting the value as String, see the code from ElasticSearchUtils below:

```
Date valueToDate = ElasticSearchDateConverter.tryToConvert((String) value);
                    if (valueToDate == null) {
                        values[i] = value;
                    } else {
                        values[i] = valueToDate;
                    }
```
I don't like a method that returns null though. I'd love to have the concept of Optional in Java7 so this method would return a Optional[Date] but for the time being I think this is a solution that might work.

@LosD Unfortunately I don't think that there is such utility method in ES.

kaspersorensen: Ah right, I didn't see the utils class yet, I was just looking here.

kaspersorensen:
+1 from me then :)

ardlema:
@LosD is this also ok with you?

LosD:
Yep! I'm no fan of allowing stuff the schema doesn't, but if we can't get ES to help us out, we don't really have a choice.

ardlema:
Yes I agree. I guess this is the best worst option. I'll merge this into master tomorrow morning then.

Thank you @kaspersorensen @LosD for your comments!

PR #44, opened 2015-08-24
Metamodel 179

tomaszguzialek:


hadoop/src/test/java/org/apache/metamodel/util/HdfsResourceIntegrationTest.java
```diff
@@ -1,187 +1,244 @@
...
+                _port = Integer.parseInt(portString);
+            }
+        } else {
+            configured = false;
+        }
+        Assume.assumeTrue(configured);
+    }
+
+    @After
+    public void tearDown() throws Exception {
```
tomaszguzialek: Diff not visible, so marking the place where changes are. Added a tearDown method to make sure we don't leave our test file in the cluster.

hadoop/src/test/java/org/apache/metamodel/util/HdfsResourceIntegrationTest.java
```diff
@@ -1,187 +1,244 @@
...
+        logger.info(stopwatch.elapsed(TimeUnit.MILLISECONDS) + " - deleted");
+
+        Assert.assertFalse(res1.isExists());
+
+        logger.info(stopwatch.elapsed(TimeUnit.MILLISECONDS) + " - done");
+        stopwatch.stop();
+    }
+
+    @Test
+    public void testFileSystemNotBeingClosed() throws IOException {
```
tomaszguzialek: Diff not visible, so marking the place where changes are. Added a test case.

LosD:
Looks good to me

tomaszguzialek:
I had to remove tearDown method and put the test in try-finally block to make the build work.

kaspersorensen:
+1 from me too

kaspersorensen:
merging

PR #45, opened 2015-08-24
Jackson dependency update

kaspersorensen:
Quite minor, but ensures that all jackson dependencies are on the same version


kaspersorensen:
Actually scratch that, seems this issue is not what I thought it was ... I must have some version conflicts in my consuming application.

PR #47, opened 2015-08-25
Converted OperatorType from an enum to an interface

kaspersorensen:
This branch is not yet functional, _but_ I wanted to raise the idea of making OperatorType an interface instead of an enum. We recently did the same with FunctionType and it enables us to potentially plug in new implementations without breaking backwards compatibiltity. In the case of OperatorType I imagine that some custome operators could be plugged in to support custom cases.

I didn't quite finish this PR/branch yet because I ran into problems. The main one being deserialization of old OperatorType enum objects. I think this can be overcome with a hack a la the thing we have with LegacyColumnType.

Another issue is to get rid of any remaining switch-case blocks which rely on OperatorType to be an enum.

But before handling those issues I wanted to take the temperature in the community - is this good/worth doing?


ardlema:
Absolutely! I think this is worth doing.

As we've already seen with other cases transforming some enums to interfaces will open up a wider range of possibilities and this case is a clear example. So please go ahead with this PR.

kaspersorensen:
OK I did an update of the branch. Will also create a JIRA issue now. Please review :-)

kaspersorensen:
This is my suggested fix for METAMODEL-188

core/src/test/java/org/apache/metamodel/query/OperatorTypeTest.java
```diff
@@ -23,12 +23,12 @@
...
-        assertEquals(OperatorType.IN, OperatorType.convertOperatorType("IN"));
-        assertEquals(null, OperatorType.convertOperatorType("foo"));
+        assertEquals(OperatorType.EQUALS_TO, OperatorTypeImpl.convertOperatorType("="));
+        assertEquals(OperatorType.GREATER_THAN, OperatorTypeImpl.convertOperatorType(">"));
+        assertEquals(OperatorType.LESS_THAN, OperatorTypeImpl.convertOperatorType("<"));
+        assertEquals(OperatorType.DIFFERENT_FROM, OperatorTypeImpl.convertOperatorType("<>"));
+        assertEquals(OperatorType.LIKE, OperatorTypeImpl.convertOperatorType("LIKE"));
+        assertEquals(OperatorType.IN, OperatorTypeImpl.convertOperatorType("IN"));
+        assertEquals(null, OperatorTypeImpl.convertOperatorType("foo"));
     }
```
albertostratio: Should we add the other "cases"?? ("==",  "EQ", "EQUALS_TO"...)

LosD:
LGTM, except that I don't really like that OperatorType is dependent on OperatorTypeImpl. It's probably unavoidable if you want to keep it compatible, but it tastes... Dirty.

kaspersorensen:
Addressed the unittest request from @ardlema 

Regarding your remark @LosD - yeah, I think this is unavoidable. And at least it is consistent with how the ColumnType and FunctionType interfaces are implemented.

albertostratio:
LGTM

kaspersorensen:
OK I'll merge it then.

PR #48, opened 2015-09-03
Use missing and exists filters for null comparisons.

LosD:
This will make isNull() and isNotNull() work. They failed before, as a term query, which does not support null values, was used. A test was also un-ignored; It was ignored as it used to fail, but the fix  for METAMODEL-172 fixed the test as well.

Fixes METAMODEL-184

### DIFF
--- a/elasticsearch/src/test/java/org/apache/metamodel/elasticsearch/ElasticSearchDataContextTest.java
+++ b/elasticsearch/src/test/java/org/apache/metamodel/elasticsearch/ElasticSearchDataContextTest.java
@@ -18,9 +18,6 @@
  */
 package org.apache.metamodel.elasticsearch;
 
-import static org.elasticsearch.common.xcontent.XContentFactory.jsonBuilder;
-import static org.junit.Assert.*;
-
 import java.util.Arrays;
 import java.util.Date;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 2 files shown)


elasticsearch/src/test/java/org/apache/metamodel/elasticsearch/ElasticSearchDataContextTest.java
```diff
@@ -559,7 +560,16 @@ private static void indexBulkDocuments(String indexName, String indexType, int n
 
     }
 
-    private static void indexOneTweeterDocumentPerIndex(String indexType, int id) {
+    private static void indexTweeterDocument(String indexType, int id, Date date) {
+        try {
+            client.prepareIndex(indexName, indexType).setSource(buildTweeterJson(id, date))
+                    .setId("tweet_" + indexType + "_" + id).execute().actionGet();
+        } catch (Exception ex) {
+            System.out.println("Exception indexing documents!!!!!");
```
kaspersorensen: Wouldn't a rethrow be better?

kaspersorensen:
LGTM, but just a question regarding the test code.

elasticsearch/src/test/java/org/apache/metamodel/elasticsearch/ElasticSearchDataContextTest.java
LosD: I completely agree (except I wouldn't rethrow, but use throws instead), but it seemed to be the pattern for all the indexing methods here, so I just went with the flow :)

kaspersorensen:
+1 from me

PR #49, opened 2015-09-09
Use File instead of InputStream for Excel when possible

LosD:
The memory usage by Apache POI is quite a bit lower when using File instead of InputStream. This makes sure that FileResource's internal File object will be passed object to POI.

POI has some problems with saving to an already opened file, so a temp file is used for saving.

Fixes METAMODEL-187

### DIFF
--- a/excel/src/main/java/org/apache/metamodel/excel/ExcelDataContext.java
+++ b/excel/src/main/java/org/apache/metamodel/excel/ExcelDataContext.java
@@ -58,12 +58,13 @@ public final class ExcelDataContext extends QueryPostprocessDataContext implemen
 
     private final Resource _resource;
     private final ExcelConfiguration _configuration;
+
     private SpreadsheetReaderDelegate _spreadsheetReaderDelegate;
 
     /**
      * Constructs an Excel DataContext based on a file, with default
      * configuration
-     * 
+     *
      * @param file
      */
     public ExcelDataContext(File file) {
@@ -91,6 +92,7 @@ public final class ExcelDataContext extends QueryPostprocessDataContext implemen
         if (file.exists() && !file.canRead()) {
             throw new IllegalArgumentException("Cannot read from file");
         }
+
         _resource = new FileResource(file);
         _configuration = configuration;
     }
@@ -145,25 +147,19 @@ public final class ExcelDataContext extends QueryPostprocessDataContext implemen
 
     @Override
     public DataSet materializeMainSchemaTable(Table table, Column[] columns, int maxRows) {
-
-        Ref<InputStream> inputStreamRef = getInputStreamRef();
-        InputStream inputStream = null;
         try {
-            SpreadsheetReaderDelegate delegate = getSpreadsheetReaderDelegate(inputStreamRef);
-            inputStream = inputStreamRef.get();
+            SpreadsheetReaderDelegate delegate = getSpreadsheetReaderDelegate();
 
             // METAMODEL-47: Ensure that we have loaded the schema at this point
             getDefaultSchema();
 
-            DataSet dataSet = delegate.executeQuery(inputStream, table, columns, maxRows);
+            DataSet dataSet = delegate.executeQuery(_resource, table, columns, maxRows);
             return dataSet;
         } catch (Exception e) {
             if (e instanceof RuntimeException) {
                 throw (RuntimeException) e;
             }
             throw new MetaModelException("Unexpected exception while materializing main schema table", e);
-        } finally {
-            FileHelper.safeClose(inputStream);
         }
     }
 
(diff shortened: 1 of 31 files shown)


LosD:
Memory usage on a huge file, 50 MB/ ~1000000 million records (Excel max).

InputStream:
memory use inputstream-based

File:
memory use file-based

kaspersorensen:
+1

Code looks good to me.
Nice findings wrt. memory improvement and "good enough" documentation for me to prove it.

Minor: I am missing a JIRA issue reference and an update to CHANGES.md

LosD:
Added :)

LosD:
Seems Travis is exploding. JVM segfaults in both 7 and 8.

LosD:
Okay, I'm officially giving up on this approach. I'm not sure, but the main issue seems to be the mixing of write-only SXSSFWorkbook and read-write XSSFWorkbook.

I suggest we unmix them, maybe with some kind of flag for output-only mode, if that is even feasible. If not, I think we should just drop it, or see if XSSF alone could work for big Excel sheets, while still giving us memory benefits compared to the InputStream (though I doubt it).

kaspersorensen:
Damn. And we only see this happen on Travis, right?

LosD:
Yeah, but I'm pretty sure it is somehow timing related, so I'll bet that users will see it too. There's nothing that should be Travis specific.

excel/src/main/java/org/apache/metamodel/excel/ExcelUpdateCallback.java
```diff
@@ -78,7 +87,24 @@ public ExcelDataContext getDataContext() {
 
     protected void close() {
         if (_workbook != null) {
-            ExcelUtils.writeWorkbook(_dataContext, _workbook);
+
+            final Resource resource;
+            if(_dataContext.getResource() instanceof FileResource){
```
kaspersorensen: Trying to find what the issue might be ... And this if-sentence is making me wonder: Why is it here, and if it is supposed to write non-file resources to a temp-file, isn't it all wrong then? Shouldn't it be "if resources is NOT a FileResource"?

excel/src/main/java/org/apache/metamodel/excel/ExcelDataContext.java
```diff
@@ -209,13 +205,13 @@ protected void onSchemaCacheRefreshed() {
...
 
-    private SpreadsheetReaderDelegate getSpreadsheetReaderDelegate(Ref<InputStream> inputStream)
+    private SpreadsheetReaderDelegate getSpreadsheetReaderDelegate()
             throws MetaModelException {
         if (_spreadsheetReaderDelegate == null) {
             synchronized (this) {
                 if (_spreadsheetReaderDelegate == null) {
                     try {
-                        if (POIXMLDocument.hasOOXMLHeader(inputStream.get())) {
+                        if (POIXMLDocument.hasOOXMLHeader(getInputStream())) {
```
kaspersorensen: The InputStream here is never closed. In fact I think this code should be rewritten like this:

```
    private SpreadsheetReaderDelegate getSpreadsheetReaderDelegate() throws MetaModelException {
        if (_spreadsheetReaderDelegate == null) {
            synchronized (this) {
                if (_spreadsheetReaderDelegate == null) {
                    _spreadsheetReaderDelegate = _resource.read(new Func<InputStream, SpreadsheetReaderDelegate>() {
                        @Override
                        public SpreadsheetReaderDelegate eval(InputStream inputStream) {
                            try {
                                if (POIXMLDocument.hasOOXMLHeader(inputStream)) {
                                    return new XlsxSpreadsheetReaderDelegate(_configuration);
                                } else {
                                    return new DefaultSpreadsheetReaderDelegate(_configuration);
                                }
                            } catch (IOException e) {
                                logger.error("Could not identify spreadsheet type, using default", e);
                                return new DefaultSpreadsheetReaderDelegate(_configuration);
                            }
                        }
                    });
                }
            }
        }
        return _spreadsheetReaderDelegate;
    }
```

... and then we should delete ```getInputStream()``` and ```getInputStreamRef()```.

... but I tried that, and it may be better, but it didn't make my Java not crash :-P

LosD: Wait, does that mean that your Java is doing like Travis? We're totally doing this on your computer. If it's your old laptop, bring it! If it's the Mac... I'd consider it a hardware issue :-P

LosD: Rewrite looks good, except I'm not sure I like the logger.error() when still falling back. Either cancel with an error, or try to recover with a warning. 

(this may be the same as before, but Github's diff interface is terrible on the mobile, so I can't see it)

excel/src/main/java/org/apache/metamodel/excel/ExcelUpdateCallback.java
LosD: I think it was an attempt to avoid file locking issues I saw with File-based Excel writing on close. In the end of my attempts there was quite a bit of throwing stuff at the wall to see what stuck (I couldn't test locally, as only Travis would fail).

kaspersorensen:
Please see #56 ... Tried many things, and yes it was reproducable on my Mac :-)

PR #51, opened 2015-10-02
METAMODEL-191: Tattletale plugin to resolve dependency conflicts

kaspersorensen:
This is my initial PR for adding the tattletale plugin to MetaModel.

Initially this does not work and breaks the build. Here's the tattletale report that I get:
https://cdn.rawgit.com/kaspersorensen/5a89bbb8cbcd935afdfe/raw/074b06a071f3002dd0deb822498e5ea00a77cbe1/index.html


kaspersorensen:
OK I believe that this fixes it all. But let's see Travis validate that first :-)

kaspersorensen:
Merging this as per lazy consensus.

PR #53, opened 2015-10-02
METAMODEL-182: getSize() and getLastModified() on directory-based resource

kaspersorensen:
This fixes getSize() and getLastModified() on the directory-based FileResource and HdfsResource classes.


kaspersorensen:
Also now with fix for HDFS. Please review.

kaspersorensen:
I'll merge based on lazy consensus.

PR #54, opened 2015-10-03
Excel decimal numbers - METAMODEL-190

b789:
Fix to solve the problem of having the decimal values from excel truncated to two digits.

https://issues.apache.org/jira/browse/METAMODEL-190


kaspersorensen:
LGTM!

kaspersorensen:
OK I will merge as no one else seems to object.

PR #55, opened 2015-10-03
Metamodel 192/datatype conversion functions

kaspersorensen:
This is my attempt at implementing METAMODEL-192, which thereby is also the first time we make pluggable ScalarFunctions available.

### DIFF
--- a/core/src/main/java/org/apache/metamodel/MetaModelHelper.java
+++ b/core/src/main/java/org/apache/metamodel/MetaModelHelper.java
@@ -48,6 +48,7 @@ import org.apache.metamodel.query.FromItem;
 import org.apache.metamodel.query.GroupByItem;
 import org.apache.metamodel.query.OrderByItem;
 import org.apache.metamodel.query.Query;
+import org.apache.metamodel.query.ScalarFunction;
 import org.apache.metamodel.query.SelectItem;
 import org.apache.metamodel.query.parser.QueryParser;
 import org.apache.metamodel.schema.Column;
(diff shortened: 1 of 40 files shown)


albertostratio:
Really nice approach!

I woudl add some unit tests for TO_STRING, TO_DATE and TO_BOOLEAN functions.

kaspersorensen:
Great, then I'll continue down this line ...

A few TODOs that I can think of:

 - [ ] More rigoruous testing - does it also work in filter items, does it work in subqueries and all that.
 - [ ] Add some hooks to the query parser to understand it.
 - [ ] Maybe also add aliases in the query parser? I notice that for scalar functions there are often different names for the same thing in many SQL dialects.
 - [ ] Add JDBC support by querying the IQueryRewriter interface whether or not particular scalar functions are supported.

kaspersorensen:
As you can see on my todo-list above, actually now it works in JDBC also and I have improved testing and all.

The only thing that doesn't work (that I know of) is having scalar functions in e.g. the WHERE clause of a JDBC query. I couldn't yet figure out a good way to decompose that problem. I can think of three solution scenarios:

 1. Simply not support it, but throw a MetaModel exception when it is attempted. For me this is actually OK to begin with.
 1. Have a failover strategy when this happens which uses a QueryPostprocessDataContext and thus eventually one or more full-table-scans to execute the query.
 1. Do something really clever to decompose the query and figure out how much of it can be natively executed, do that, and postprocess the rest. This seems very complex so I wouldn't recommend it for now.

What do you guys think? I prefer 1 or 2.

albertostratio:
I'd go for the first option. As a first approach I think that is fine to just throwing an exception. I always like to start with the simplest solution and then try to improve it afterwards (if it is needed)

So..my +1 to the first one

kaspersorensen:
I agree. Would rather finish this quickly, maybe create a JIRA issue for improving it, but have something that works for a limited scope. And I do think that having scalar functions "just" for SELECT clause is already a big deal.

kaspersorensen:
OK as far as I am concerned, this story is now complete (assuming Travis builds alright). Could someone give it a code review and then let's merge it if everybody still likes it? :-)

Side-note: I would think we then consider MetaModel 4.4.0 kinda _done_? Anyone else has pending stuff to add before we do a release?

LosD:
Looks good to me, but it would be nice to maybe also allow something like SQL's ```CAST <expression> as <type>``` later.

+1

kaspersorensen:
Let's create a separate issue in JIRA for that. I could imagine a little parser trick to do it, but I guess for consistency that format is an anomaly compared to other function literals that normally say ```function(argument)``` :-)

kaspersorensen:
Reported your idea in METAMODEL-193 - https://issues.apache.org/jira/browse/METAMODEL-193

PR #56, opened 2015-10-04
Metamodel 187/excel file another attempt

kaspersorensen:
Here's my attempt at solving METAMODEL-187. Heavily inspired by #49 but to get my head around it I had to start over and track resource usage/closing very closely.

I finally found the culprit (maybe one of more) and it was to close the workbook in ExcelUtils.writeWorkbook(...) before flushing the new contents to the resource.

I also updated POI since the errors given in the later version was giving a tiny bit more sense.

I also used an InMemoryResource instead of a temp file resource. I am working on the assumption that this can fit into memory easily in it's persisted form, but that the memory improvement we need to do is only on the unpacked form.


excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -128,24 +142,31 @@ public static boolean isXlsxFile(Resource resource) {
...
-        final Resource resource = dataContext.getResource();
-        resource.write(new Action<OutputStream>() {
+    /**
+     * Writes the {@link Workbook} to a {@link Resource}. The {@link Workbook}
+     * will be closed as a result of this operation!
+     * 
+     * @param dataContext
+     * @param wb
+     */
+    public static void writeAndCloseWorkbook(ExcelDataContext dataContext, final Workbook wb) {
```
kaspersorensen: Please wait before merging. I would like to try one potential optimization:

```
if (dataContext.getResource() instanceof FileResource) {
  wb.close();
  return;
}
```

I believe this should work because it seems that workbooks are aware if they have a file handle or not. And upon close() they will persist to the file, if available. That's at least my working assumption since what I saw previously when I was doing the copy BEFORE close, was that I ended up with an empty sheet in the end.

kaspersorensen:
Hmm build is breaking on Windows now. This is really odd, but I confirmed offline that both myself and @LosD are getting this error:

```
testInsertInto(org.apache.metamodel.excel.ExcelDataContextTest)  Time elapsed: 0.035 sec  <<< ERROR!
java.lang.IllegalStateException: java.nio.file.FileSystemException: target\xls_people_modified.xls: The process cannot access the file because it is being used
by another process.

        at sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:86)
        at sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:97)
        at sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:102)
        at sun.nio.fs.WindowsFileSystemProvider.implDelete(WindowsFileSystemProvider.java:269)
        at sun.nio.fs.AbstractFileSystemProvider.deleteIfExists(AbstractFileSystemProvider.java:108)
        at java.nio.file.Files.deleteIfExists(Files.java:1162)
        at java.nio.file.Files.copy(Files.java:3001)
        at org.apache.metamodel.util.FileHelper.copy(FileHelper.java:441)
        at org.apache.metamodel.excel.ExcelUtils.writeAndCloseWorkbook(ExcelUtils.java:169)
        at org.apache.metamodel.excel.ExcelUpdateCallback.close(ExcelUpdateCallback.java:81)
        at org.apache.metamodel.excel.ExcelDataContext.executeUpdate(ExcelDataContext.java:234)
        at org.apache.metamodel.excel.ExcelDataContextTest.testInsertInto(ExcelDataContextTest.java:544)
```

I've been trying to debug this quite long now, and I am stuck :-/

kaspersorensen:
OK we finally figured out that the issue was with HSSFWorkbook.write(OutputStream). Please check the bug that we registered in POI: https://bz.apache.org/bugzilla/show_bug.cgi?id=58480

For now we have a crude workaround which involves invoking System.gc() and sleeping for 800 ms. Not ideal at all. But seems to do the trick.

Justification for keeping this workaround would be that it _only_ applies to "old" Excel .xls documents and it only applies when appending/updating workbooks, not when creating new ones.

kaspersorensen:
If anyone could review or "plus one" this PR I would be happy :-)

LosD:
Can't say I'm a fan of the delay, but let us  hope it is temporary until a fix of POI is forthcoming (jow much work would it take for us to submit a patch?), and this is just temporary. 

LGTM!

PR #57, opened 2015-10-08
Support for Max Rows / Limit = 0

kaspersorensen:
Fixes the max rows = 0 support in MetaModel, see METAMODEL-194.

### DIFF
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
@@ -336,6 +336,13 @@ public class JdbcDataContext extends AbstractDataContext implements UpdateableDa
 
     private DataSet execute(Connection connection, Query query, Statement statement, JdbcCompiledQuery compiledQuery,
             JdbcCompiledQueryLease lease, Object[] values) throws SQLException, MetaModelException {
+        Integer maxRows = query.getMaxRows();
+
+        final List<SelectItem> selectItems = query.getSelectClause().getItems();
+        if (maxRows != null && maxRows.intValue() == 0) {
+            return new EmptyDataSet(selectItems);
+        }
+
         if (MetaModelHelper.containsNonSelectScalaFunctions(query)) {
             throw new MetaModelException(
                     "Scalar functions outside of SELECT clause is not supported for JDBC databases. Query rejected: "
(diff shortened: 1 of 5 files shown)


jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
```diff
@@ -336,6 +336,13 @@ public DataSet executeQuery(CompiledQuery compiledQuery, Object... values) {
 
     private DataSet execute(Connection connection, Query query, Statement statement, JdbcCompiledQuery compiledQuery,
             JdbcCompiledQueryLease lease, Object[] values) throws SQLException, MetaModelException {
+        Integer maxRows = query.getMaxRows();
```
LosD: Final

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
```diff
@@ -336,6 +336,13 @@ public DataSet executeQuery(CompiledQuery compiledQuery, Object... values) {
...
+
+        final List<SelectItem> selectItems = query.getSelectClause().getItems();
+        if (maxRows != null && maxRows.intValue() == 0) {
```
LosD: intValue() isn't needed.

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
kaspersorensen: Nope, it mutates later on in the method. Line 390.

LosD: Totally missed it. I though it was a new variable, not a moved :)

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
kaspersorensen: Actually it is IMO :-) Imagine it says
```
if (maxRows != null && maxRows == 0) {
```
Then the next guy that reads this code will think that it's a needless null check because
```
if (maxRows == 0) {
```
looks like it does the exact same thing. But in fact since it may be null the unboxing is not problem-free. So I would prefer to make it explicit so that you see a method being called on the potentially-null variable 'maxRows'.

LosD: Then the next guy doesn't try to understand the code he is changing, and shouldn't be allowed near it ;)

kaspersorensen: But making it explicit allows more people to contribute without stepping into traps then.

LosD: While that is true, I really don't hope anyone would go "Oh, I'll remove that" without at least thinking about why that would be there.

Anyway, it's not really an issue.

LosD:
+1

PR #58, opened 2015-10-16
Fix METAMODEL-198.

ankit2711:
This fixes the issue but the changes are done in the FormatHelper. Few more test cases added for Timestamp values. 
Trying to apply a fix for Timestamp operand in the DefaultQueryRewriter
does get overriden by the logic written in the FormatHelper which is
called by super.rewriteFilterItems(FilterItem). 
The FormatHelper again detects based on the columnType and changes the
values to be Date losing the nanoseconds.


kaspersorensen:
The build is failing so I doubt this is a good approach, but I'd like to give some input and thoughts... I was trying to read up on JDBC spec to understand what can be done here.

In section 13.4.2 of the JDBC 4.1 spec [1] it shows that proper SQL statement escape would be

date:  ```{d 'yyyy-mm-dd'```
time: ```{t 'hh:mm:ss'}```
timestamp: ```{ts 'yyyy-mm-dd hh:mm:ss.f . . .'}```

I'm always quite skeptical to see if the JDBC drivers out there _actually_ implement such details of the JDBC spec. Often I have found that implementations are not spot on in that regard. But I think the above could form the basis of a good default. We could override the default in case of deviating JDBC drivers.

Our normal way of handling such deviations is to not rely on the FormatHelper (because it is generic and context-free) but rather rely on the ```IQueryRewriter``` interface, which actually has the method ```rewriteFilterItem(...)``` method. There's a hierarchy of classes that implement the interface and the default implementation (which most deviating implementations extend from) does delegate eventually to FormatHelper. But that's what we should change IMO. So instead of changing FormatHelper I would change ```DefaultQueryRewriter```.

I would also want to cover this in all of our JDBC integration tests. Those aren't part of the Travis build but once we have a good implementation going then I am happy to test at least on the usual suspect databases such as PostgreSQL, MySQL, MS SQL Server, DB2 and Oracle.

[1] http://download.oracle.com/otn-pub/jcp/jdbc-4_1-mrel-spec/jdbc4.1-fr-spec.pdf

ankit2711:
The build was failing because some tests were added with the use of the below constructor 
new Timestamp(long milliseconds) and then assertion were applied on the Timestamp in string. The constructor use with long milliseconds involves timeszone and hence the assertions failed. 

Have tested on Postgres and Oracle when using [TIMESTAMP 'yyyy-mm-dd hh:mm:ss.nnnnnn'] in the query then the database does the conversion with respect to the format used in the timestamp column of that DB. The format of the timestamp column in Oracle had the storage of timestamp value in a format like this 16-OCT-15 04.33.34.646685000 PM.

kaspersorensen:
Yes but I would love to have it solved in as generically possible way as we can. If the JDBC spec says that JDBC drivers should take into account the format of ```{ts 'yyyy-mm-dd hh:mm:ss.f . . .'}``` then that is probably the sanest to go for as the default. I would expect (but need to verify) that then the Oracle and PostgreSQL drivers can convert that expression to whatever is needed in the physical query fired from within the driver.

kaspersorensen:
Also a note on testing - if we add it as a scenario in JdbcTestTemplates then we can immediately verify it with all the automatic tests that can run for Derby, H2, HSQLDB, SQLite. If it works for those then I always feel a lot more comforted that we have something with a certain broader applicability.

ankit2711:
Modified the fix to be now in DefaultQueryRewriter, added tests which work for Derby, H2 and HSQL. For SQLLite the timestamp insertions also do not store the milliseconds so skipped for now. Nothing is done in FormatHelper specific to Timestamp now.

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/DerbyQueryRewriter.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Override
+    public String rewriteFilterItem(FilterItem item) {
+        final Object operand = item.getOperand();
+        if (operand instanceof Timestamp) {
+            final OperatorType operator = item.getOperator();
+            StringBuilder sb = new StringBuilder();
+            sb.append(item.getSelectItem().getSameQueryAlias(false));
+            FilterItem.appendOperator(sb, operand, operator);
+            sb.append("TIMESTAMP ( \'" + operand.toString() + "\') ");
```
kaspersorensen: Is this different from the parent somehow? I can't see a difference, but may be missing a detail somewhere?

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/DefaultQueryRewriter.java
```diff
@@ -106,14 +107,22 @@ private boolean needsQuoting(String alias, String identifierQuoteString) {
...
-                    FilterItem replacementFilterItem = new FilterItem(item.getSelectItem(), item.getOperator(), str);
+                    FilterItem replacementFilterItem = new FilterItem(selectItem, item.getOperator(), str);
                     return super.rewriteFilterItem(replacementFilterItem);
                 }
+            } else if (operand instanceof Timestamp) {
+                final OperatorType operator = item.getOperator();
+                StringBuilder sb = new StringBuilder();
+                sb.append(selectItem.getSameQueryAlias(false));
+                FilterItem.appendOperator(sb, operand, operator);
+                sb.append("TIMESTAMP \'" + operand.toString() + "\'");
```
kaspersorensen: Would probably be a good idea to make a protected method "rewriteTimestamp(Timestamp)" so that we can easily overwrite just that in any deviating dialects.

jdbc/src/test/java/org/apache/metamodel/jdbc/JdbcTestTemplates.java
```diff
@@ -633,4 +634,84 @@ public void run(UpdateCallback callback) {
...
+        } catch (SQLException e) {
+            // do nothing
+        }
+
+        assertFalse(conn.isReadOnly());
+
+        JdbcDataContext dc = new JdbcDataContext(conn);
+        final Schema schema = dc.getDefaultSchema();
+        final Timestamp timestamp1 = Timestamp.valueOf("2015-10-16 16:33:33.456");
+        final Timestamp timestamp2 = Timestamp.valueOf("2015-10-16 16:33:34.683");
```
kaspersorensen: These are just milliseconds I think? Shouldn't we be adding nanoseconds to the test, or else we're not getting into the real precision gotchas here?

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/DerbyQueryRewriter.java
ankit2711: Yes it expects the format like TIMESTAMP ( ' operand value ' ), so the round braces are needed else it does not work. I guess timestamp queries on Derby might be broken on master.

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/DefaultQueryRewriter.java
ankit2711: Done.

jdbc/src/test/java/org/apache/metamodel/jdbc/JdbcTestTemplates.java
ankit2711: This is hard because not all DB drivers claim to have the complete fractions parts. If the 6 digit number is placed for the nanos then in Derby. H2 and HSQL only first 3 digits are stored.

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/DefaultQueryRewriter.java
```diff
@@ -139,13 +143,24 @@ public String rewriteFilterItem(FilterItem item) {
...
-                FilterItem replacementFilterItem = new FilterItem(item.getSelectItem(), item.getOperator(), elements);
+                FilterItem replacementFilterItem = new FilterItem(selectItem, item.getOperator(), elements);
                 return super.rewriteFilterItem(replacementFilterItem);
             }
         }
         return super.rewriteFilterItem(item);
     }
-    
+
+    protected String rewriteTimestamp(FilterItem item) {
```
kaspersorensen: Now this method does a rewrite of a whole filter item and not just the timestamp literal. I was suggesting to make something that would make the burden of overriding as little as possible, so that you would only need to do like this:

```
protected String rewriteTimestamp(Timestamp timestamp) {
  return MyStrangeTimestampFormat.format(timestamp);
} 
```

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/DefaultQueryRewriter.java
```diff
@@ -139,13 +143,24 @@ public String rewriteFilterItem(FilterItem item) {
...
+        final OperatorType operator = item.getOperator();
+        final SelectItem selectItem = item.getSelectItem();
+        final Object operand = item.getOperand();
+        StringBuilder sb = new StringBuilder();
+        sb.append(selectItem.getSameQueryAlias(false));
+        FilterItem.appendOperator(sb, operand, operator);
+        sb.append("TIMESTAMP \'" + operand.toString() + "\'");
```
kaspersorensen: Did you try with the official JDBC escape syntax? I bet that has a lot better general applicability than this format...

kaspersorensen: To be specific, I would change this line to:

```
sb.append("{ts '" +  operand.toString() + "'}");
```

jdbc/src/test/java/org/apache/metamodel/jdbc/JdbcTestTemplates.java
kaspersorensen: That's exactly why I want to use the JDBC escape syntax so the DB can make such conversions itself.

ankit2711:
Done all.

ankit2711:
The fixes are also done in another pull request from Kasper. That pull request also fixes some issues with Timestamp while insertion time. Closing this PR then.

PR #59, opened 2015-10-18
METAMODEL-198: Implemented timestamp literal based on JDBC spec

kaspersorensen:
Inspired by #58 I decided to make this PR to show what I had in mind as an alternative patch which aligns closer to the JDBC spec instead of taking the literalization of timestamps into MetaModel's codebase.

### DIFF
--- a/jdbc/pom.xml
+++ b/jdbc/pom.xml
@@ -72,9 +72,9 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
-			<groupId>postgresql</groupId>
+			<groupId>org.postgresql</groupId>
 			<artifactId>postgresql</artifactId>
-			<version>9.1-901.jdbc4</version>
+			<version>9.3-1104-jdbc4</version>
 			<scope>test</scope>
 		</dependency>
 		<dependency>
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
@@ -54,6 +54,7 @@ import org.apache.metamodel.jdbc.dialects.MysqlQueryRewriter;
 import org.apache.metamodel.jdbc.dialects.OracleQueryRewriter;
 import org.apache.metamodel.jdbc.dialects.PostgresqlQueryRewriter;
 import org.apache.metamodel.jdbc.dialects.SQLServerQueryRewriter;
+import org.apache.metamodel.jdbc.dialects.SQLiteQueryRewriter;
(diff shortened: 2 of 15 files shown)


kaspersorensen:
@ankit2711 please take a look and let me know if this makes sense to you.

kaspersorensen:
Hmm seems it doesn't work for SQLite ... It seems to return a String and not a Timestamp! Strange. Will work a bit more on this, but I see that also #58 has no test on SQLite, so that's probably the same issue.

kaspersorensen:
OK a bit more research and coding done ...

FIrst of all, we still missed one half of the problem, so I did that: The timestamps that are inserted where wrongly re-created and that made them loose their nano-second precision. So I've changed that in JdbcUtils.setStatementValue(...)

The other thing I would want to report is that since I know you faced this issue in PostgreSQL, I wanted to test specifically on that database - hence the update to PostgresqlTest. BUT I must say that my finding was a bit dissapointing. It seems that the PostgreSQL driver has sub-nanosecond precision. To be precise it "only" has precision of a 1000th of a millisecond. Several sources explain that this is an issue with how the database formats the output when sending it to JDBC [1] and I couldn't find any solutions to the problem that didn't involve compiling PostgreSQL yourself(!). Apparently the microsecond precision that they claim to have is only in the storage layer. The issue can also be seen in simple pgAdmin usage:

image

The above fraction-of-second numbers are only 6 digit while a nano would be 9 digit.

As a result, the PostgresqlTest only does assertions on millisecond-precision.

[1] http://www.postgresql.org/message-id/3A370254.44C55482@alumni.caltech.edu

LosD:
'To be precise it "only" has precision of a 1000th of a millisecond.'
'Apparently the microsecond precision that they claim to have is only in the storage layer.'
One thousandth of a millisecond is a microsecond... Doesn't that make the microsecond precision claim correct? Or did you mean nanosecond precision?

LosD:
The message you refer to is from 2000, and is not representative of 2015 (it talks about 10th of a seoncd).

Your example seems to show microsecond precision in the output.

kaspersorensen:
I made a mistake in my research! Somehow got the impression that 1 nanosecond = 1000 microsecond. But it's in deed the other way around! Then the strange rounding issue that I saw makes absolute sense because PostgreSQL's rounding to the nearest 1000 nanos is exactly 1 microsecond! :-)

I guess I just need to then update the JdbcTestTemplates class to also support Microsecond precision.

ankit2711:
+1 wil fix this issue for now.

kaspersorensen:
OK I will merge this in ... I also now tested it on the major databases; MySQL, Oracle, DB2 and MS SQL Server.

PR #60, opened 2015-10-20
Maven-assembly-plugin in MetaModel-full.

tomaszguzialek:
During "package" phase it creates the distributable version in target folder of the MetaModel-full module.

Fixes METAMODEL-144


tomaszguzialek:
The build failed because of Cassandra cluster not starting within timeout, so unrelated issue.

kaspersorensen:
This looks quite alright, except I am a bit wondering about two ugly things in the diff:

 * The license header in pom.xml files is screwed.
 * The log4j.xml file seems to have changed everywhere, but I cannot see any different. Newlines?

kaspersorensen:
Looks very nice now. I'll do a quick run locally to verify the zip file's contents. And then merge :-)

PR #61, opened 2015-10-22
Let ElasticSearchDataContext read changes

LosD:
This adds the ability for ElasticSearchDataContext to read  tables on
 updates, to assure that an outside change is registered. To
 keep the ability to specify own mappings, any static mappings will
 be left alone.

Fixes METAMODEL-197


kaspersorensen:
+1 from me :)

PR #62, opened 2015-10-23
METAMODEL-200: Added query syntax support for "function approximation"

kaspersorensen:
Fixes METAMODEL-200 by adding both toSql() and parser support for the tilde-character as a prefix "approximation indicator" for function names.

I also improved a few places in SelectItem.java where function.toString() was used instead of function.getFunctionName() (there's currently no difference because they all delegate toString() to getFunctionName(), but that should only be an implementation detail).

### DIFF
--- a/core/src/main/java/org/apache/metamodel/query/SelectItem.java
+++ b/core/src/main/java/org/apache/metamodel/query/SelectItem.java
@@ -47,7 +47,8 @@ import org.slf4j.LoggerFactory;
  */
 public class SelectItem extends BaseObject implements QueryItem, Cloneable {
 
-    public static final char FUNCTION_APPROXIMATION_PREFIX = '~';
+    public static final String FUNCTION_APPROXIMATION_PREFIX = "APPROXIMATE ";
+    
     private static final long serialVersionUID = 317475105509663973L;
     private static final Logger logger = LoggerFactory.getLogger(SelectItem.class);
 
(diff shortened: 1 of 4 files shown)


LosD:
Code looks good, but it's not a particular good change to me: It is totally fine to add new functions (like approximate_count() or something like that) that are not base SQL, but breaking SQL syntax seems a bit... Nasty to me.

Relational database vendors do all kinds of horrible things to SQL, but I don't think even they would go that far.

kaspersorensen:
The concern is that actually it's a flag you can set independently of which function it applies to. I have only myself used it wrt. COUNT function, but you could also query an approximate AVG function or something like that. This means that "just" adding approximate_count would not work - and it would not support people plugging in new functions themselves.

I guess we could introduce it as a first-class keyword like DISTINCT, then it would be ```SELECT APPROXIMATE COUNT(*) FROM table``` instead ... Not sure I like that better, but it would work I guess.

A word of comfort: People that expect MetaModel to use "standard SQL" would probably never set this flag because it is not part of any standard at all. So they would never encounter the tilde symbol anywhere :-)

LosD:
I'd still like SELECT APPROXIMATE much better. It's not breaking syntax :)

kaspersorensen:
OK I will try that then.

kaspersorensen:
Anyone else who has an oppinion on this? @LosD are you OK with the change now?

LosD:
Much better! :-)

PR #64, opened 2015-10-23
Upgraded to MongoDB 3.x

kaspersorensen:
All changes should be backwards compatible.

Fixes METAMODEL-203.


kaspersorensen:
Guess I was too fast ... Will check what API change was made that didn't work here ... Seems we may need to change a bit then.

LosD:
LGTM! :+1:

PR #66, opened 2015-10-26
METAMODEL-204: Fixed

kaspersorensen:
My attempt at fixing METAMODEL-204 which unfortunately seemed to block the release of MetaModel 4.4.1.

### DIFF
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/AbstractQueryRewriter.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/AbstractQueryRewriter.java
@@ -280,6 +280,12 @@ public abstract class AbstractQueryRewriter implements IQueryRewriter {
     }
 
     protected String rewriteSelectItem(Query query, SelectItem item) {
+        if (item.isFunctionApproximationAllowed()) {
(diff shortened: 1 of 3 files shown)


LosD:
:+1:

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/AbstractQueryRewriter.java
```diff
@@ -280,6 +280,12 @@ protected String rewriteSelectClause(Query query, SelectClause selectClause) {
     }
 
     protected String rewriteSelectItem(Query query, SelectItem item) {
+        if (item.isFunctionApproximationAllowed()) {
+            // function approximation is not included in any standard SQL
+            // constructions - will have to be overridden by subclasses if there
+            // are specialized dialects for it.
+            item = item.replaceFunctionApproximationAllowed(false);
+        }
```
ClaudiaPHI: shouldn't be "true" ?

LosD: Nope, it's removing it, so that it doesn't end up in end at i.e. JDBC contexts

PR #68, opened 2015-10-29
Mongodb api upgrade

loorenzooo:
Migrate mongo java driver client code to new api avoiding usage og deprecated methods.
ex : 
- MongoDatabase is replacing DB
- MongoCollection is replacing DBCollection
...


full/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,8 @@
+# Root logger option
+log4j.rootLogger=INFO, stdout
+
+# Direct log messages to stdout
+log4j.appender.stdout=org.apache.log4j.ConsoleAppender
+log4j.appender.stdout.Target=System.out
+log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
+log4j.appender.stdout.layout.ConversionPattern=%d{yyyy-MM-dd HH:mm:ss} %-5p %c{1}:%L - %m%n
```
LosD: I don't think we have Log4J in MetaModel? Leftover from testing?

mongodb/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,8 @@
+# Root logger option
+log4j.rootLogger=DEBUG, stdout
+
+# Direct log messages to stdout
+log4j.appender.stdout=org.apache.log4j.ConsoleAppender
+log4j.appender.stdout.Target=System.out
+log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
+log4j.appender.stdout.layout.ConversionPattern=%d{yyyy-MM-dd HH:mm:ss} %-5p %c{1}:%L - %m%n
```
LosD: Also a log4j.properties file here

LosD:
Travis seems to be angry about an extra column

LosD:
Code looks good (and I must commend the Mongo guys for keeping the API compatible, ElasticSearch could learn something here).

However, I can't help wonder if it isn't a little bit early to throw Mongo 2 compatibility out the window (3.0 is from March this year). Should we maybe consider making separate Mongo 2 and Mongo 3 modules for now?

albertostratio:
@LosD this is a really good point. I think we should think about maintaining different versions of the same connectors. It would obviously force us to maintain more code base but we shouldn't force the clients of the library to use a specific version of a datastore.

kaspersorensen:
I also think we should discuss this. An alternative option (short term) is to have two DataContext implementations in the same module. That would allow the old one to use the old deprecated API, but also have a new implementation which uses the new API.

full/src/test/java/org/apache/metamodel/ExcelDataContextFactoryTest.java
```diff
@@ -23,9 +23,7 @@
 import org.apache.metamodel.excel.ExcelConfiguration;
 import org.apache.metamodel.excel.ExcelDataContext;
 
-import junit.framework.TestCase;
-
-public class DataContextFactoryTest extends TestCase {
+public class ExcelDataContextFactoryTest extends  MongoDbTestCase{
```
kaspersorensen: I know this class currently seems to only test the excel scenario, but it is a test of ```DataContextFactory``` so I find it more natural to name it like that?

kaspersorensen:
Please be aware that I raised another topic regarding this PR (and #67) in the mailing list under subject "[DISCUSS] ElasticSearch and MongoDB granularity"

kaspersorensen:
Superceeded by #86

PR #69, opened 2015-10-30
Neo4j module (query-only)

tomaszguzialek:
A query-only version of Neo4j module. Modification of the data in Neo4j will come later, but it would be nice to have this merged first. 

Node labels are represented as tables and node properties, relationships and its properties are modelled as columns.


example-metamodel-integrationtest-configuration.properties
```diff
@@ -106,4 +106,13 @@
...
+#cassandra.keyspace=my_keyspace
+
+# ---------------------------
+# Neo4j module properties:
+# ---------------------------
+
+#neo4j.hostname=localhost
+#neo4j.port=7474
+#neo4j.username=neo4j
+#neo4j.password=neo4j
```
kaspersorensen: Very minor, but let's add a linebreak here so that we don't one day get a diff line like the first line here.

full/pom.xml
```diff
@@ -157,6 +157,16 @@ under the License.
 		</dependency>
 		<dependency>
 			<groupId>org.apache.metamodel</groupId>
+			<artifactId>MetaModel-neo4j</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.metamodel</groupId>
+			<artifactId>MetaModel-json</artifactId>
```
kaspersorensen: Eh, what? Is MM-json missing in full module? If so that's an independent bug and something to add separately I think.

neo4j/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+			<artifactId>httpclient</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.slf4j</groupId>
+			<artifactId>jcl-over-slf4j</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>com.fasterxml.jackson</groupId>
+			<artifactId>jackson-datatype-json-org</artifactId>
+			<version>1.8.0</version>
```
kaspersorensen: Let's get this under dependency management so that it follows our other jackson dependencies in version etc.

neo4j/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+			<groupId>junit</groupId>
+			<artifactId>junit</artifactId>
+			<scope>test</scope>
+		</dependency>
+	</dependencies>
+
+	<repositories>
+		<repository>
+			<id>neo4j-public</id>
+			<url>http://m2.neo4j.org/content/groups/public</url>
```
kaspersorensen: As a central repo publisher we are not allowed to have dependencies from any other repositories. Is this really needed? If so ... Damn! We would have to check if there's any special tricks we can pull then.

neo4j/src/main/java/org/apache/metamodel/neo4j/Neo4jDataContext.java
```diff
@@ -0,0 +1,336 @@
...
+
+    @Override
+    protected String getMainSchemaName() throws MetaModelException {
+        return SCHEMA_NAME;
+    }
+
+    public SimpleTableDef[] detectTableDefs() {
+        List<SimpleTableDef> tableDefs = new ArrayList<SimpleTableDef>();
+
+        String labelsJsonString = _requestWrapper.executeRestRequest(new HttpGet("/db/data/labels"));
```
kaspersorensen: Maybe we should put the service root (/db/data) in some constant or even a field. It seems from the docs [1] that while this is the default URL it might be different in certain environments. And at least also that the result of it will provide the URLs for other services.

[1] http://neo4j.com/docs/stable/rest-api-service-root.html

neo4j/src/main/java/org/apache/metamodel/neo4j/Neo4jDataContext.java
```diff
@@ -0,0 +1,336 @@
...
+
+    @Override
+    protected DataSet materializeMainSchemaTable(Table table, Column[] columns, int maxRows) {
+        return materializeMainSchemaTable(table, columns, 1, maxRows);
+    }
+
+    @Override
+    protected org.apache.metamodel.data.Row executePrimaryKeyLookupQuery(Table table, List<SelectItem> selectItems,
+            Column primaryKeyColumn, Object keyValue) {
+        return null;
```
kaspersorensen: If we don't have an implementation here then I suggest not overriding it since it does nothing.

neo4j/src/main/java/org/apache/metamodel/neo4j/Neo4jDataContext.java
```diff
@@ -0,0 +1,336 @@
...
+            // Do not throw an exception here. Returning null here will make
+            // MetaModel attempt to count records manually and therefore recover
+            // from the error.
+            return null;
+        }
+    }
+
+    @Override
+    public void executeUpdate(UpdateScript script) {
+        throw new UnsupportedOperationException(
```
kaspersorensen: I would just not extend the UpdateableDataContext interface then. This method just gives the wrong impression.

neo4j/src/main/java/org/apache/metamodel/neo4j/Neo4jRequestWrapper.java
```diff
@@ -0,0 +1,126 @@
...
+	private final String _username;
+	private final String _password;
+
+	public Neo4jRequestWrapper(CloseableHttpClient httpClient,
+			HttpHost httpHost, String username, String password) {
+		_httpClient = httpClient;
+		_httpHost = httpHost;
+		_username = username;
+		_password = password;
+		_cypherQueryHttpPost = new HttpPost("/db/data/transaction/commit");
```
kaspersorensen: I am unsure how this code (and the dataset that eventually consumes it) will work on queries with large results. Is there not some kind of paging mechanism in Neo4j so that you should fire another request once you've traversed the first? If not all records have to fit into the response JSON which seems very memory intensive and non-scalable.

neo4j/src/main/java/org/apache/metamodel/neo4j/Neo4jDataSet.java
```diff
@@ -0,0 +1,77 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+
+final class Neo4jDataSet extends AbstractDataSet {
+
+    private JSONObject _resultJSONObject;
+    private int _currentRowIndex;
+    private Row _row;
+
+    public Neo4jDataSet(SelectItem[] selectItems, JSONObject resultJSONObject) {
```
kaspersorensen: To elaborate on my previous remark: Here I would expect that the JSONObject passed in only represents "page 1 of N".

neo4j/src/main/java/org/apache/metamodel/neo4j/Neo4jInferentialSchemaBuilder.java
```diff
@@ -0,0 +1,44 @@
...
+
+public class Neo4jInferentialSchemaBuilder extends InferentialSchemaBuilder {
+
+    public Neo4jInferentialSchemaBuilder() {
+        super(Neo4jDataContext.SCHEMA_NAME);
+    }
+
+    @Override
+    public DocumentConverter getDocumentConverter(Table table) {
+        // TODO Auto-generated method stub
```
kaspersorensen: Something missing here I guess.

neo4j/src/test/resources/log4j.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE log4j:configuration SYSTEM "log4j.dtd">
+<log4j:configuration xmlns:log4j="http://jakarta.apache.org/log4j/">
```
kaspersorensen: is log4j even used in test? I don't see any log4j dependency in the pom.

example-metamodel-integrationtest-configuration.properties
tomaszguzialek: I added a line with serviceroot and an empty line after it, just to be sure.

neo4j/src/main/java/org/apache/metamodel/neo4j/Neo4jDataSet.java
tomaszguzialek: I played a little with that now, but did not come to a working solution. Is that necessary to merge this PR? Or maybe we can create a separate issue for it and go ahead with what we have?

kaspersorensen: That depends a bit on what it actually means right now. Does it work on large datasets at all? I would expect there to be some paging implemented by default (say, more than 1000 or 10000 rows or so) which means that this design is defect and will cut off a part of the dataset. That's such a big issue/bug that I would vote against it. But if it is functional right now and it's "just" a memory issue then I could live with it for now under the saying "first make it functional, then make it optimal".

kaspersorensen:
Another general but minor issue: It seems that the code is using tabs instead of spaces for indentation.

neo4j/src/main/java/org/apache/metamodel/neo4j/Neo4jDataSet.java
tomaszguzialek: As far as I understand, there is no pagination by default. Limit and offset should be explicitly in the query. I added a test case (ignored by default) selecting 100k records. It takes long time, but it works, there is no cut off after certain number of records.

kaspersorensen: OK then let's just write up a JIRA for improving this aspect.

tomaszguzialek: Done: https://issues.apache.org/jira/browse/METAMODEL-208

tomaszguzialek:
Updated my Eclipse formatting setting and formatted all the classes in the PR. I believe identation is fixed now, so also the whole PR is ready to be merged again.

kaspersorensen:
Checked out the code - looks good to me. Do you want to do the honors of merging it in? Maybe squashing the diff to a single commit would be good so that we don't retain the long history of minor changes :)

tomaszguzialek:
Can I squash that into one commit when there other branches merged into that one? Rebase is giving me a little trouble here...

tomaszguzialek:
I merged the branch manually with --squash argument.

PR #70, opened 2015-11-02
METAMODEL-195 - a MAP_VALUE function

kaspersorensen:
This PR attempts to fix METAMODEL-195, the addition of a MAP_VALUE function which takes a string parameter to tell which value from within a map you wish to get. It's syntax is like this: ```MAP_VALUE('path.to.value', column)```

There is also support for a query parser shorthand syntax where you select like this: ```column.path.to.value```

It only works on MAP type columns.

Point to especially consider: Since this is the first function which involves parameters, I ended up also doing some redesign of the ScalarFunction interface. I initially built it so that parameters was part of the MapValueFunction instance and thus you would need to instantiate the function for each parameter set you have. This worked kinda well just until I came to the point of parsing and printing the function as SQL. The trouble here was that if the interface did not reveal the parameters in any way then automating the printing etc. wasn't possible either. I am a bit unhappy about changing the interface already since it was launched in MetaModel 4.4.0, but figure we kind of need it if we want to support functions like this. I actually think that the interface I ended up with here could be further improved, but this was the first stab to get your feedback.


core/src/main/java/org/apache/metamodel/query/FromClause.java
```diff
@@ -75,7 +75,13 @@ public FromItem getItemByReference(String reference) {
...
 
-    private FromItem getItemByReference(FromItem item, String reference) {
+    private FromItem getItemByReference(final FromItem item, final String reference) {
+        if (reference.equals(item.toStringNoAlias(false))) {
+            return item;
+        }
+        if (reference.equals(item.toStringNoAlias(true))) {
+            return item;
+        }
         final String alias = item.getAlias();
```
albertostratio: Wouldn't it be better to write just one if instead of two? I'd rewrite it to:

if (reference.equals(item.toStringNoAlias(false)) || reference.equals(item.toStringNoAlias(true))) {
            return item;
}

albertostratio:
@kaspersorensen are you still working on this PR? I think this function is pretty useful.

And, I've got a question on this... Will it fetch all the "results" for the following json?:

```
{
    "data": {
        "results": [
            {
                "date": "2015-09-16 15:37:38",
                "email": "dummy@gmail.com",
                "password": 12345,
                "source_type": "leakage"
            },
            {
                "date": "2015-12-23 12:37:38",
                "email": "dummy2@gmail.com",
                "password": 78921,
                "source_type": "another"
            }
        ]
    }
}
```

kaspersorensen:
I stopped a bit because it involved an interface change to ScalaFunction which I think is kind of a pity - we just launched that interface. But otoh it's maybe better to do the change early before anyone starts relying too much on it.

core/src/main/java/org/apache/metamodel/query/FromClause.java
kaspersorensen: Makes no difference to me, so I can easily change it :-)

kaspersorensen:
To you example JSON: right now it would only return the first. If we wanted it to be capable of changing the record count of the resulting dataset then I guess it's another pretty big change.

albertostratio:
@kaspersorensen I agree, the sooner we modify the interface the better.

Regarding my question about the json, if it requires another big change I guess it doesn't make sense to make it as part of this PR.

kaspersorensen:
OK I'll try and finish this one up then. I have let it linger for too long, but finally I think actually it's a pretty good change.

kaspersorensen:
OK, done with this PR for now. Please give another review if you can and then we can merge hopefully?

kaspersorensen:
What do you think about merging this @albertostratio?

core/src/main/java/org/apache/metamodel/query/SelectItem.java
```diff
@@ -152,7 +166,24 @@ public SelectItem(Column column, FromItem fromItem) {
      * @param fromItem
      */
     public SelectItem(FunctionType function, Column column, FromItem fromItem) {
-        this(column, fromItem, function, null, null, null, false);
+        this(column, fromItem, function, null, null, null, null, false);
```
albertostratio: Would it make more sense to put the if before the creation of the object? I mean if the column is null and we're throwing an exception, why are we instantiating the object?

core/src/main/java/org/apache/metamodel/query/SelectItem.java
```diff
@@ -152,7 +166,24 @@ public SelectItem(Column column, FromItem fromItem) {
...
+     * from a particular {@link FromItem}, for example
+     * MAP_VALUE('path.to.value', doc)
+     * 
+     * @param function
+     * @param functionParameters
+     * @param column
+     * @param fromItem
+     */
+    public SelectItem(FunctionType function, Object[] functionParameters, Column column, FromItem fromItem) {
+        this(column, fromItem, function, functionParameters, null, null, null, false);
```
albertostratio: Same comment as above

core/src/main/java/org/apache/metamodel/query/SelectItem.java
```diff
@@ -178,7 +209,7 @@ public SelectItem(String expression, String alias) {
      * @param alias
      */
     public SelectItem(FunctionType function, String expression, String alias) {
-        this(null, null, function, expression, null, alias, false);
+        this(null, null, function, null, expression, null, alias, false);
         if (expression == null) {
```
albertostratio: Same comment as above

core/src/main/java/org/apache/metamodel/query/SelectItem.java
```diff
@@ -192,7 +223,7 @@ public SelectItem(FunctionType function, String expression, String alias) {
      *            the FromItem that holds the sub-query
      */
     public SelectItem(SelectItem subQuerySelectItem, FromItem subQueryFromItem) {
-        this(null, subQueryFromItem, null, null, subQuerySelectItem, null, false);
+        this(null, subQueryFromItem, null, null, null, subQuerySelectItem, null, false);
         if (subQueryFromItem.getSubQuery() == null) {
```
albertostratio: Same comment as above

albertostratio:
@kaspersorensen some minor comments take a look at them please

core/src/main/java/org/apache/metamodel/query/SelectItem.java
kaspersorensen: It's not possible in Java to put anything before the call to ```this(...)``` so that's not possible.

albertostratio: Sorry about that then

albertostratio:
LGTM

PR #73, opened 2015-11-26
METAMODEL-211 Mongo autogenerated _id handled properly in where clause

mishadoff:
This fix processes each value in filter clauses and if it is a valid Mongo bson id, converts it into ObjectId. Tests are included and cover select by id and delete by id queries.


kaspersorensen:
+1 from me

tomaszguzialek:
Also +1 from me.

PR #74, opened 2015-12-04
Added new aggregate functions FIRST, LAST, RANDOM

kaspersorensen:
Just sharing this little idea that I had while travelling the other day ... Adding some more aggregate functions to MetaModel.

 * FIRST which picks the first value
 * LAST which picks the last value
 * RANDOM which picks a random value

One limitation of the PR is that the JDBC module does not support these functions. This is because there is no fallback strategy in place for the JDBC module when a particular query part cannot be pushed to the database. In my opinion that is already a problem that we have lived with - for instance in case of "scalar functions in where clause" which is also not supported - so I didn't try to solve it, just make sure it fails fast (and in a similar fashion to the mentioned other case).


kaspersorensen:
Update the PR with more tests and made the JDBC module fail fast. Will update my previous PR description accordingly.

Now this is ready for a final review IMO - any takers?

jdbc/src/test/java/org/apache/metamodel/jdbc/JdbcDataContextTest.java
```diff
@@ -261,6 +265,20 @@ public void testWhereScalarFunction() throws Exception {
...
+            dataContext.query().from("customers").select(FunctionType.RANDOM, "customernumber").execute();
+            fail("Exception expected");
+        } catch (MetaModelException e) {
+            assertEquals(
+                    "Aggregate function 'RANDOM' is not supported on this JDBC database. Query rejected: "
+                            + "SELECT RANDOM(\"CUSTOMERS\".\"CUSTOMERNUMBER\") FROM PUBLIC.\"CUSTOMERS\"",
+                    e.getMessage());
+        }
+    }
+
```
LosD: This is fine, though reformatting the whole file for adding a single function is a pretty history-destructive.

That much of the formatting is pretty weird doesn't make it better :)

jdbc/src/test/java/org/apache/metamodel/jdbc/JdbcDataContextTest.java
```diff
@@ -148,20 +148,22 @@ public void testExecuteQuery() throws Exception {
...
+                result.getRow().toString());
         assertTrue(result.next());
         assertTrue(result.next());
         assertTrue(result.next());
         assertTrue(result.next());
         assertEquals(
                 "Row[values=[121, Baane Mini Imports, Bergulfsen, Jonas, 07-98 9555, Erling Skakkes gate 78, null, "
-                        + "Stavern, null, 4110, Norway, 1504, 81700.0]]", result.getRow().toString());
+                        + "Stavern, null, 4110, Norway, 1504, 81700.0]]",
+                result.getRow().toString());
```
LosD: Example of weird reformat

LosD:
LGTM, but I'm really no fan of reformatting an entire file unless we touch much of the file. It makes reviews much harder than they need to be, and destroys history.

I'd suggest we keep our hands off the reformat shortcut without selecting an area around our edits.

kaspersorensen:
Agree. This was unintentional and just muscle-memory driven. Will fix it. Maybe we can put the eclipse formatting rules somewhere so that one can make sure he is using the single-version-of-the-formatting-truth :)

LosD:
Hehe, yeah. :)

PR #75, opened 2015-12-07
Hardened the NumberComparator to have better support for number types.

kaspersorensen:
(Sorry for the code reformatting, but the old was clearly wrong - used
tabs instead of spaces)

### DIFF
--- a/core/src/main/java/org/apache/metamodel/util/NumberComparator.java
+++ b/core/src/main/java/org/apache/metamodel/util/NumberComparator.java
@@ -66,18 +66,18 @@ public final class NumberComparator implements Comparator<Object> {
     }
 
     public int compare(Object o1, Object o2) {
-        if (o1 == null && o2 == null) {
-            return 0;
-        }
-        if (o1 == null) {
-            return -1;
-        }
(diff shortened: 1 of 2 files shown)


LosD:
Ugh, huge reformatting. But fair, most of the file was changed anyway, and it had the wrong indentation character before (and the secret 'add ?w=1 to URL' Github trick saved the comparison).

LosD:
I realize we are for much more than SQL, but the null comparisons being equal worries me a bit, "NULL is never equal to anything" is  a pretty important part of SQL (and maybe one of the few things all SQL vendors can agree on).

OTOH, NULL is never NOT equal to anything else, so this may still be wrong SQL-wise.

LosD:
Okay, it probably doesn't matter. It doesn't seem to be used for (SQL) equals unless there's a need to parse it anyway.

LosD:
Let's go with the LGTM :)

kaspersorensen:
Glad that you sorted it out ;-)

LosD:
Yeah. That @losd guy started panicking, but I calmed him down :)

PR #76, opened 2015-12-07
METAMODEL-218: Fixed

kaspersorensen:
This fixes METAMODEL-218 by adding a few return statement in DefaultQueryRewriter.

Also updated the test template to ensure that it's working on all the databases covered by these tests.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
  * [METAMODEL-216] - Added new aggregate functions: FIRST, LAST and RANDOM.
  * [METAMODEL-15] - Query parser support for table names with space. Delimitters can be double quote or square brackets. 
  * [METAMODEL-215] - Improved the capability of NumberComparator to support Integer, Long, Double, BigInteger and other built-in Number classes.
+ * [METAMODEL-218] - Fixed conversion of STRING and NUMBER types to database-specific types in JDBC module.
 
 ### Apache MetaModel 4.4.1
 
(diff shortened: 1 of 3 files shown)


drexler42:
I'll give it a go and provide feedback

kaspersorensen:
Did you come to any conclusion @drexler42 ?

drexler42:
Not yet, I'm afraid. Having trouble building a DataCleaner with this MM in it.

drexler42:
I can copy the customers table directly into Postgres now. The fix works. A word of warning though:
- MetaModel 4.4.2-SNAPSHOT is not compatible with DataCleaner master.
-- Package names of ElasticSearch datastore is changed in MM, and must be changed accordingly in DC.
-- The new MM neo4j module introduces a tattletale clash on an json library.

I excluded the culprit from DC, but then the API module does not compile anymore. I gave up on solving that at that point.
```
			<dependency>
				<groupId>org.apache.metamodel</groupId>
				<artifactId>MetaModel-full</artifactId>
				<exclusions>
					<exclusion>
						<artifactId>json</artifactId>
						<groupId>org.apache.geronimo.bundles</groupId>
					</exclusion>
				</exclusions>
			</dependency>
```

The fix in itself is fine. So I'll merge it.

kaspersorensen:
I merged in master ... Was just a conflict in CHANGES.md. But let's see that Travis builds correctly before we merge to master

drexler42:
Let's merge then!

PR #77, opened 2015-12-09
First attempts at making the API change for METAMODEL-6

kaspersorensen:
This PR proposes an important API change (admittedly without any proper implementation at this stage) to allow the publishing of information about a given data change/update.

I decided to post this PR without any proper implementation because (1) the API change alone will be beneficial to do sooner rather than later - we seem to be approaching a 4.5.0 release at the moment, so I find it smarter to break the API now instead of having to wait for the next major release and (2) adding the better implementation of the update summary can always be done on a case-by-case basis. I would suggest to start with a generic method for the number of inserts since that can be generalized very easily.

### DIFF
--- a/core/src/main/java/org/apache/metamodel/AbstractUpdateCallback.java
+++ b/core/src/main/java/org/apache/metamodel/AbstractUpdateCallback.java
@@ -43,21 +43,21 @@ public abstract class AbstractUpdateCallback implements UpdateCallback {
     @Override
     public TableCreationBuilder createTable(String schemaName, String tableName) throws IllegalArgumentException,
             IllegalStateException {
-        Schema schema = getSchema(schemaName);
+        final Schema schema = getSchema(schemaName);
         return createTable(schema, tableName);
     }
 
     @Override
     public TableDropBuilder dropTable(String schemaName, String tableName) throws IllegalArgumentException,
             IllegalStateException, UnsupportedOperationException {
-        Table table = getTable(schemaName, tableName);
+        final Table table = getTable(schemaName, tableName);
         return dropTable(table);
     }
 
     @Override
     public TableDropBuilder dropTable(Schema schema, String tableName) throws IllegalArgumentException,
             IllegalStateException, UnsupportedOperationException {
-        Table table = schema.getTableByName(tableName);
+        final Table table = schema.getTableByName(tableName);
         if (table == null) {
             throw new IllegalArgumentException("Nu such table '" + tableName + "' found in schema: " + schema
                     + ". Available tables are: " + Arrays.toString(schema.getTableNames()));
@@ -158,5 +158,9 @@ public abstract class AbstractUpdateCallback implements UpdateCallback {
     public RowUpdationBuilder update(Table table) throws IllegalArgumentException, IllegalStateException,
             UnsupportedOperationException {
         return new DeleteAndInsertBuilder(this, table);
+    }
+    
+    public UpdateSummary getUpdateSummary() {
+        return DefaultUpdateSummary.unknownUpdates();
     }
 }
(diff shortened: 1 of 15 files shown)


kaspersorensen:
Maybe someone would like to give their feedback on this change? Since it changes the return type of ```executeUpdate``` it will mean that all using code has to be recompiled. I guess we need that sooner than later. But just want to ensure that you agree on the change?

albertostratio:
I really like the idea of having the capability to report anything about an update but I don't like nulls representing "not known/not available" values.

It would be great if we could use the Java 8 Optional structure in this case :(

kaspersorensen:
hmm yes. But I guess that would rule out Java 7 support :-/

We could opt for using Guava's optional class, but I think there's even a lot of discussion going around about "how can we migrate Guava's optional class into the JDK one" which to me indicates that it's maybe not the best road to start going down.

kaspersorensen:
Would you prefer that we switched to ```int``` instead and had exceptional return values like ```-1```?

albertostratio:
Yes, I'm not a fan of guava's library.

I'd go for the int solution but again this is a personal opinion. I'm programming in Scala for a while so every time I see a method returning a null it seems not right for me.

kaspersorensen:
Usually @LosD has quite an opinion on stuff like this - so maybe he'd like to pitch in? :-)

LosD:
Hehe, I'm all for nuking Java7 support, but I do think nulls are the perfect unknown/unavailable variable... Essentially, that's what they are for.

However, those are general observations, I'm in transit right now, so I can't really check code.

LosD:
OTOH, as an old C/Unix  programmer, I do like int values with -1 being the unavailable value... But what do we do in other similar places? If nothing else gives either approach a clear reason to be, API consistency should decide.

kaspersorensen:
My worry with the -1 is that someone might do math operations on this - for instance if they perform a series of updates on a DataContext and then want to present a sum of all the inserts, all the updates, all the deletions etc. In such a case we might easily expose them to a bug where they put -1 into their sum. I guess the alternative will null is that they will get a NullPointerException and then realize that they couldn't always rely on the value being set. Which is IMO slightly better. Optional would be preferable for sure tho'!

LosD:
It's a good point, but not critical IMHO, if you don't read the JavaDoc of the method (which is missing by the way... I have Internet on the plane, so I finally had a chance to take a look :)),  you brought it on yourself...  But if nothing else dictates int, we might as well make it clear by using null.

kaspersorensen:
I added a bit of javadoc for that then :-)

kaspersorensen:
Shall we accept this PR (with possible null return values) then? I think the conclusion was that null is the "most correct among the incorrect" options we have as long as we want to retain Java 7 compatibility. Agreed?

LosD:
Honestly, for me it's a good reason to nuke Java 7 compatibility. Then again, an itch in my pinky toe would be enough for me...  I can think of nothing that would stop anyone from going to Java 8 that wouldn't also stop them from upgrading MM, except for bad excuses. Oracle EOLed Java 7 in April for most.

LosD:
(That means that if your software vendor is requiring you to use Java 7, they are requiring you to have a major security hole in your server/workstation, or to pay top dollar to Oracle)

kaspersorensen:
I don't like the idea of suddenly changing our requirement from Java 7 to Java 8, just because a new feature could benefit from it. I know I'm being a party-pooper, but I think that kind of stuff needs to be planned and handled in a way where we give people more choice about when they take the step to Java 8. If nothing else, the application I (we ;-)) work on most has roughly 40% users that still are on Java 7, so we wanted to not retire it just yet... But soon for sure. But other companies are surely more conservative and will wait longer. My final word on Java 8 would be: Probably there are _many_ more API improvements we could make if we decide to go with Java 8. I would suggest we start a separate topic about that and consider that a part of a future MetaModel 5.0 with renewed API in many places.

LosD:
In that case, LGTM

albertostratio:
LGTM

kaspersorensen:
I will retract this PR and instead post a new one which suggests to merge it into the new 5.x branch

PR #79, opened 2015-12-10
HdfsResource with non-"hdfs" scheme

kaspersorensen:
Suggested fix for METAMODEL-220.

In this PR I've added a scheme to the HdfsResource. To make the scheme dynamic I also needed to improve the URI parsing, which anyways ended up being a very good improvement.


LosD:
I do believe we now made HdfsResource able to use anything the Hadoop Filesystem supports, even FTP, "normal" files and S3 directly (without using EMR).

... Which is not a bad thing, but maybe makes the "Hdfs" naming a little funny. Not really a problem, it is still intended for Hdfs derivates, more of a curiosity.

Besides the manual URI assembly (which may just be me), LGTM.

kaspersorensen:
You're right about FTP, normal files and S3. We could add those schemes too, if that's what you mean? But the way of accessing them will be through HDFS - or at least via the _hadoop (pause) FileSystem_ (class). If it makes things more clear, we could create a new class instead, called "HadoopResource" which would cover all the different schemes (and could then hold _just_ an URI instead of the individual URI elements).

I just now tried to see if I could do the URI assembly in a more reusing way. Encountered the following small issues:

 * UriBuilder is part of JAX-RS (and there is also one in apache HTTP components). I think it would be unfortunate to depend on an optional library just to do this.
 * So then I tried instead the URI class...
  * That requires me to catching a URISyntaxException which will never happen. A bit annoying IMO.
  * It changes a small thing in one of the test cases. In the case of ```emrfs:///home/metamodel.txt``` it shortens it by removing two slashes, making it just ```emrfs:/home/metamodel.txt```. I guess that's a valid change from a pure URI point of view, but it's not what I have ever seen in the documentation/examples for these kinds of schemes.

LosD:
No no, no trouble with the naming or the listed schemes, it was just an observation. I don't really think it's that interesting from a user point of view (though S3 support is a very nice thing, it would be another VFS-like thing, but inside HdfsResource, and by having it there we indirectly recommend it).

We already depend on URIBuilder trough Hadoop (hadoop-common depends on jersey-core), but you are of course right that a direct dependency makes us vulnerable if Hadoop suddenly decides not to use a JAX-RS implementation.

The triple slash is being removed because host is null instead of an empty string, AFAIK.
```scheme://host/path``` with a null ```host``` will become ```scheme:/path```

But never mind, no reason to make a big deal out of it just because I hate manual URIs :)

LosD:
:+1:

PR #82, opened 2015-12-17
METAMODEL-224 CONCAT function support first approach (non-functional)

ardlema:
These are the first steps to add support for the CONCAT scalar function.

I've got a bit bogged down with this, not sure whether I'm going down the right path. It's been quite complicated because is the first time we're having a function that might have one or more parameters and even the parameters might apply to one, two... columns or they might be just literals. We might have the following CONCAT functions:

CONCAT(column1, 'hello', column2)
CONCAT('hello', 'world')
CONCAT(column1, column2, column3)

I've created a new select method with the following parameters: function and functionParameters where the functionParameters are going to be the column names (or literals). But I've found out that when I evaluate the function within my new ConcatFunction class the selectItems from the row do not contain all the columns so I can't get the values to create the result.

I'm afraid that I'd need some tips to keep working on this :(

### DIFF
--- a/core/src/main/java/org/apache/metamodel/MetaModelHelper.java
+++ b/core/src/main/java/org/apache/metamodel/MetaModelHelper.java
@@ -277,7 +277,7 @@ public final class MetaModelHelper {
         final List<SelectItem> scalarFunctionSelectItemsToEvaluate = new ArrayList<>();
 
         for (SelectItem selectItem : selectItems) {
-            if (selectItem.getScalarFunction() != null) {
+            if (selectItem !=null && selectItem.getScalarFunction() != null) {
                 if (!dataSetSelectItems.contains(selectItem)
                         && dataSetSelectItems.contains(selectItem.replaceFunction(null))) {
                     scalarFunctionSelectItemsToEvaluate.add(selectItem);
@@ -768,7 +768,9 @@ public final class MetaModelHelper {
     public static SelectItem[] createSelectItems(Column... columns) {
         SelectItem[] items = new SelectItem[columns.length];
         for (int i = 0; i < items.length; i++) {
-            items[i] = new SelectItem(columns[i]);
+            if (columns[i] != null) {
+                items[i] = new SelectItem(columns[i]);
+            }
         }
         return items;
     }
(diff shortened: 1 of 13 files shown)


kaspersorensen:
Hi Alberto,

OK let me try to see if I can guide you on the way. For me this is also a bit uncertain how we fix it, but I can show you at least where the problem lies and then we might contemplate ways to fix.

If you look at QueryPostprocessDataContext you will find a fairly big method, ```executeQuery(Query)```, which handles the first stuff about identifying the select items (each representing a column - or so we thought) to include from each queried table. If you look at line 185 the final "workSelectItems" is assembled for this.

Now the "issue" (as I see it) is that each select item may represent 1+n columns. And in line 192 we call materializeFromItem(...) with this list. So either we fix it in the materializeFromItem method, or maybe even before we go there. This is at list kind of the "hot spot" for debugging it.

albertostratio:
I agree @kaspersorensen right now each SelectITem represent just one column. My last attempt (see my last commit) has been kind of hacking the functionParameters object list of my new SelectItem constructor and try to mix-in columns and literals within the same Object array.

But I feel like the steps I'm doing so far are not in the right direction :(

Thank you for you piece of advice, I've gotta go now but I'll keep working on this tomorrow.

albertostratio:
@kaspersorensen I've been having a look at the lines that you told me and not sure whether this is the right place to work on. I've been debugging the test that I created in QueryPostProcessDataContextText (testScalarFunctionConcat) and the code is not even reaching that lines (is a singleFromItem and so is returning the DataSet before assigning the workSelectItems)

kaspersorensen:
That sounds a bit strange because only very specific query types do that. For instance "primary key lookup" queries and "select count(*)" queries. But OK, maybe your query is a primary key lookup with a scalar function on it? Could be.

And I guess that really reveals the problem... That in ```executePrimaryKeyLookupQuery(...)``` we are _passing in_ the select item. That used to be OK because select items wasn't that difficult. But now it select item is more complicated. So maybe we should rather change the contract so that ```executePrimaryKeyLookupQuery(...)```, ```materializeMainSchemaTable``` and any similar methods do NOT get a list of select items, but instead get a list of columns. And then their results will be generically postprocessed to add the scalar function layer on top.

albertostratio:
That makes sense to me, let me work on that direction and I'll get back to you. Thank your for your help, great to work with you again ;)

ardlema:
I've closed this PR 'cos I'm working on a new approach

PR #83, opened 2015-12-17
Added fix for escaping in CsvWriter in case where quote char is not set

kaspersorensen:
Today I discovered a small bug in MetaModel's CsvWriter - that it would fail to escape a separator char if there is no quote char. Please find here my patch for the problem.

### DIFF
--- a/csv/src/main/java/org/apache/metamodel/csv/CsvWriter.java
+++ b/csv/src/main/java/org/apache/metamodel/csv/CsvWriter.java
@@ -72,23 +72,38 @@ public final class CsvWriter {
     }
 
     private boolean stringContainsSpecialCharacters(String line) {
-        return line.indexOf(_configuration.getQuoteChar()) != -1 || line.indexOf(_configuration.getEscapeChar()) != -1;
+        boolean result = line.indexOf(_configuration.getQuoteChar()) != -1
+                || line.indexOf(_configuration.getEscapeChar()) != -1;
+        if (!result) {
+            result = _configuration.getQuoteChar() == CsvConfiguration.NOT_A_CHAR
+                    && _configuration.getEscapeChar() != CsvConfiguration.NOT_A_CHAR;
+        }
+        return result;
     }
 
-    private StringBuilder processLine(String nextElement) {
-        final StringBuilder sb = new StringBuilder(INITIAL_STRING_SIZE);
+    private String processLine(String nextElement) {
+        final char escapeChar = _configuration.getEscapeChar();
+        if (escapeChar == CsvConfiguration.NOT_A_CHAR) {
+            return nextElement;
+        }
+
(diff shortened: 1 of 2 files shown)


csv/src/main/java/org/apache/metamodel/csv/CsvWriter.java
```diff
@@ -72,23 +72,38 @@ public String buildLine(String[] nextLine) {
     }
 
     private boolean stringContainsSpecialCharacters(String line) {
-        return line.indexOf(_configuration.getQuoteChar()) != -1 || line.indexOf(_configuration.getEscapeChar()) != -1;
+        boolean result = line.indexOf(_configuration.getQuoteChar()) != -1
+                || line.indexOf(_configuration.getEscapeChar()) != -1;
```
LosD: .contains() (both line 75 and 76)

csv/src/main/java/org/apache/metamodel/csv/CsvWriter.java
```diff
@@ -72,23 +72,38 @@ public String buildLine(String[] nextLine) {
...
+        if (!result) {
+            result = _configuration.getQuoteChar() == CsvConfiguration.NOT_A_CHAR
+                    && _configuration.getEscapeChar() != CsvConfiguration.NOT_A_CHAR;
```
LosD: I have a hard time understanding this part: In the previous lines, if the line doesn't contain a quote char AND doesn't contain an escape char, result would be false.

But here (line 77-79), if quote is not set, and escape is, we pretend one of them was contained anyway? It seems to me that the method is either misnamed or mis-implemented.

LosD:
Except for the weird ```stringContainsSpecialCharacters```, LGTM.

csv/src/main/java/org/apache/metamodel/csv/CsvWriter.java
kaspersorensen: Actually .indexOf(...) is better here because it is a char and the search time of ```indexOf(char)``` is presumably a lot faster than ```contains(String)``` since it only has to evaluate a single character.

LosD: Ah, that makes sense

csv/src/main/java/org/apache/metamodel/csv/CsvWriter.java
kaspersorensen: You're right something looks funky here. I think it should be ```... && line.indexOf(quoteChar) != -1``` ... let me check with the unittest etc.

kaspersorensen: Duh, sorry I meant to put separatorChar instead of quoteChar in that remark.

PR #84, opened 2016-01-02
METAMODEL-6: Added UpdateSummary building for JDBC module

kaspersorensen:
Some more progress on METAMODEL-6. This adds support for the JDBC module to return number of inserts, updates, deletes and any generated keys. I chose the JDBC module because it is arguably the most complex. I imagine the reusing bits and pieces from here would be beneficial when implementing support in some of the other modules.

### DIFF
--- a/core/src/main/java/org/apache/metamodel/AbstractUpdateCallback.java
+++ b/core/src/main/java/org/apache/metamodel/AbstractUpdateCallback.java
@@ -48,14 +48,14 @@ public abstract class AbstractUpdateCallback implements UpdateCallback {
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


albertostratio:
Would it make sense to also test the getUpdatedRows and the getDeletedRows methods in the test (deleting and updating some rows and checking these numbers)?

kaspersorensen:
Yes it would :-) Let's do that.

kaspersorensen:
Added the assertions. Please re-review.

albertostratio:
LGTM @kaspersorensen 

Good job!

PR #85, opened 2016-01-07
METAMODEL-224 CONCAT function support

ardlema:
Second attempt of supporting the CONCAT scalar function, this time I make it work.


core/src/main/java/org/apache/metamodel/query/ConcatFunction.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Override
+    public Object evaluate(Row row, Object[] parameters, SelectItem operandItem) {
+        SelectItem[] selectItems = row.getSelectItems();
+        if (parameters.length == 0) {
+            throw new IllegalArgumentException("Expecting some parameters to CONCAT function");
+        }
+        StringBuilder strBuilder = new StringBuilder();
+        for (Object parameter: parameters) {
+            String parameterAsString = parameter.toString();
```
kaspersorensen: Wouldn't it make a lot of things easier if we decided to treat the parameters as SelectItems instead of Strings? What I mean is - in this part of the code (and in ScalarFunctionRow) the the changes made are quite involved because the parameters are of type String. But what if we instead changed it so that the parameters are simply SelectItems. I guess that is also what they _are_ from a (SQL) language perspective: ```CONCAT(foo, bar)``` would indicate that "foo" and "bar" are select items, whereas ```CONCAT(foo, " ", bar)``` would be a select item, a string and then a select item. I think we could greatly simplify the code if we work with this model instead.

core/src/main/java/org/apache/metamodel/query/SelectItem.java
```diff
@@ -484,11 +498,24 @@ public StringBuilder toStringNoAlias(boolean includeSchemaInColumnPath) {
...
+                        functionBeginning.append(getToStringColumnPrefix(includeSchemaInColumnPath, col));
+                        functionBeginning.append(col.getQuotedName());
+                    } else {
+                        if (!_function.getFunctionName().equals(FunctionType.CONCAT.getFunctionName())) {
+                            functionBeginning.append('\'');
+                        }
+                        functionBeginning.append(functionParameters[i]);
+                        if (!_function.getFunctionName().equals(FunctionType.CONCAT.getFunctionName())) {
+                            functionBeginning.append('\'');
+                        }
```
kaspersorensen: Also the changes here seem very unfortunate because they have a specific-case exception for something that is very general. If we instead just made sure that _any_ function could have select item as parameter, then we're in a much better position.

core/src/main/java/org/apache/metamodel/query/builder/ConcatSelectBuilderImpl.java
```diff
@@ -0,0 +1,56 @@
...
+package org.apache.metamodel.query.builder;
+
+import org.apache.metamodel.query.FunctionType;
+import org.apache.metamodel.query.Query;
+import org.apache.metamodel.query.SelectItem;
+import org.apache.metamodel.schema.Column;
+
+import java.util.List;
+
+final class ConcatSelectBuilderImpl extends SatisfiedSelectBuilderImpl implements
```
kaspersorensen: This class is named with 'Concat...' but it doesn't seem to have anything specifically to do with concat - or am I missing something? If it's generic, then let's also name it generically.

core/src/main/java/org/apache/metamodel/query/ConcatFunction.java
albertostratio: In this function the parameters are not only Strings they are either Columns (for the column parameters) or Strings (for the literals). Then in the method depending of whether the param is a Column or a literal we append the value get from the row or just the literal.

core/src/main/java/org/apache/metamodel/query/SelectItem.java
albertostratio: I fully agree with this comment. I don't like these changes at all but again not sure how to implement your suggestions to improve it.

core/src/main/java/org/apache/metamodel/query/builder/ConcatSelectBuilderImpl.java
albertostratio: Good point! In fact having a look at the code I think I can get rid of this class and use the already created FunctionSelectBuilderImpl class. Thank you!

core/src/main/java/org/apache/metamodel/query/ConcatFunction.java
kaspersorensen: What I mean is that a few lines lower, starting with the ```if``` and eventually hitting the ```getColumnValueFromRow``` (line 48) the parameter is treated only as a string. But what would be more natural to me is to check what type the parameter is. I think the parameter will already be a SelectItem here actually, but it's being changed into a string via the ```toString()``` call and then converted/resolved into a column by the ```getColumnValueFromRow``` call. It would be way better (IMO) to have something like:

```
final String valueToConcat;
if (parameter instanceof SelectItem) {
  valueToConcat = row.getValue((SelectItem)parameter);
} else if (parameter instanceof Column) {
  valueToConcat = row.getValue((Column)parameter);
} else {
  valueToConcat = parameter;
}
strBuilder.append(valueToConcat);
```

And then you can delete the complete ```getColumnValueFromRow``` method which is just making things quite complex I think.

core/src/main/java/org/apache/metamodel/query/SelectItem.java
kaspersorensen: Actually, looking at it again, how about just doing an ```if (functionParameters[i] instanceof String)``` and then only adding the quotes before and after in that case? For numbers, columns, selectitems etc. we wouldn't want any quote I think. And that does not seem to be specific to the CONCAT function at all?

albertostratio:
@kaspersorensen I've addressed your suggestions, making the code more generic and cleaner, have a look and let me know what you think

The build is passing in my box, travis seems to be hung though :(

core/src/main/java/org/apache/metamodel/data/ScalarFunctionRow.java
```diff
@@ -49,8 +51,32 @@ public Object getValue(int index) throws IndexOutOfBoundsException {
             return _row.getValue(index - scalarFunctionCount);
         }
         final SelectItem selectItem = scalarFunctionSelectItems.get(index);
-        final SelectItem selectItemWithoutFunction = selectItem.replaceFunction(null);
-        return selectItem.getScalarFunction().evaluate(_row, selectItem.getFunctionParameters(), selectItemWithoutFunction);
+        if (selectItem.getScalarFunction().equals(FunctionType.CONCAT)) {
```
kaspersorensen: Another "if concat" ... I would really prefer to get rid of these since a function should be pluggable and not require code changes in the "framework" if you will.

core/src/main/java/org/apache/metamodel/query/ConcatFunction.java
```diff
@@ -0,0 +1,65 @@
...
+            if (parameter instanceof SelectItem) {
+                valueToConcat = row.getValue((SelectItem) parameter);
+            } else if (parameter instanceof Column) {
+                valueToConcat = row.getValue((Column) parameter);
+            } else {
+                valueToConcat = parameter;
+            }
+            strBuilder.append(valueToConcat.toString());
+        }
+        return strBuilder;
```
kaspersorensen: Should probably call ```toString()``` on this builder to return a String and not a StringBuilder.

core/src/main/java/org/apache/metamodel/query/ConcatFunction.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Override
+    public ColumnType getExpectedColumnType(ColumnType type) {
+        // the column type cannot be inferred so null is returned
+        return null;
```
kaspersorensen: I would say that it should return ColumnType.STRING since it's always going to be a String.

core/src/main/java/org/apache/metamodel/query/SelectItem.java
```diff
@@ -190,6 +190,17 @@ public SelectItem(FunctionType function, Object[] functionParameters, Column col
     }
 
     /**
+     * Creates a SelectItem that uses a function on a column from a particular
```
kaspersorensen: Since this constructor does not pass in any ```FromItem``` then I guess it is not a "particular FromItem" but the opposite.

core/src/main/java/org/apache/metamodel/query/builder/SatisfiedFromBuilder.java
```diff
@@ -44,6 +44,8 @@
 
     public FunctionSelectBuilder<?> select(FunctionType function, Column column, Object[] functionParameters);
 
+    public FunctionSelectBuilder<?> select(Column column, FunctionType function, Object[] functionParameters);
```
kaspersorensen: This is the same as the one just above, but just with the arguments in a different order - is that really worth introducing?

core/src/main/java/org/apache/metamodel/query/SelectItem.java
```diff
@@ -461,8 +472,11 @@ public StringBuilder toStringNoAlias() {
     public StringBuilder toStringNoAlias(boolean includeSchemaInColumnPath) {
         final StringBuilder sb = new StringBuilder();
         if (_column != null) {
-            sb.append(getToStringColumnPrefix(includeSchemaInColumnPath));
-            sb.append(_column.getQuotedName());
+            if (_function==null ||
+                    (_function!=null && !_function.getFunctionName().equals(FunctionType.CONCAT.getFunctionName()))) {
```
kaspersorensen: Unclear to me why this is needed. I would prefer this to be really agnostic of function type. Any function should be able to have a mix of parameter types (if that is somehow the reason for this if - I'm unsure).

core/src/main/java/org/apache/metamodel/query/parser/SelectItemParser.java
```diff
@@ -157,6 +161,21 @@ public SelectItem findSelectItem(String expression) throws MultipleSelectItemsPa
             if ("*".equals(columnName)) {
                 throw new MultipleSelectItemsParsedException(fromItem);
             } else if (fromItem.getTable() != null) {
+                if (selectItemHasParenthesis) {
+                    if (functionName.equals(FunctionType.CONCAT.getFunctionName())) {
```
kaspersorensen: The following code should not be CONCAT-specific IMO. It's something that can be used on any set of function parameters.

core/src/main/java/org/apache/metamodel/query/parser/SelectItemParser.java
```diff
@@ -157,6 +161,21 @@ public SelectItem findSelectItem(String expression) throws MultipleSelectItemsPa
...
+                        String[] columnAndParametersAsString = expression.split(",");
+                        Object[] columnAndParameters = new Object[columnAndParametersAsString.length];
+                        int columnAndParamsIndex = 0;
+                        for (String parameter : columnAndParametersAsString) {
+                            Column column = fromItem.getTable().getColumnByName(parameter);
+                            if (column != null) columnAndParameters[columnAndParamsIndex] = column;
+                            else columnAndParameters[columnAndParamsIndex] = parameter.substring(1,parameter.length()-1);
```
kaspersorensen: We should follow code conventions here and have newline and curly braces so that it becomes:
```
if (something) {
  something;
} else {
  something;
}
```

core/src/test/java/org/apache/metamodel/QueryPostprocessDataContextTest.java
```diff
@@ -294,6 +294,29 @@ public void testScalarFunctionWhere() throws Exception {
...
+        Table table = dc.getDefaultSchema().getTables()[0];
+        MutableColumn col = new MutableColumn("foo").setTable(table);
+        MutableColumn col2 = new MutableColumn("bar").setTable(table);
+        Object[] functionParams = new Object[3];
+        functionParams[0] = col;
+        functionParams[1] = "$";
+        functionParams[2] = col2;
+        Query query = dc.query()
+                .from(table)
+                .select(col, FunctionType.CONCAT, functionParams)
```
kaspersorensen: Actually here I realize that the function is created "only" with the functionParams as the parameters - NOT with the first col as a true parameter. Maybe that's the cause of all the exceptions I see in the code. But I think maybe then we should introduce a new SelectItem constructor (and builder methods etc) which does not take the initial column or select item but JUST a function and it's parameters.

core/src/main/java/org/apache/metamodel/data/ScalarFunctionRow.java
albertostratio: Completely agree. In fact, I think this is completely wrong: we should use the evaluate method of the concat function class instead of doing any logic in here. I'll get rid of this if.

Thank you!

core/src/main/java/org/apache/metamodel/query/ConcatFunction.java
albertostratio: I think so, will change it

core/src/main/java/org/apache/metamodel/query/SelectItem.java
albertostratio: Copy and paste is evil ;) thank you for pointing this out

core/src/test/java/org/apache/metamodel/QueryPostprocessDataContextTest.java
albertostratio: @kaspersorensen that was my first approach but I found a lot of problems when we don't have a column within the SelectItem. I guess the CONCAT function is a quite particular case because it might receive several columns but most of the rest of the scalar functions we will be receiving just one column.

I would keep this approach as a first attempt and then when we implementing another string scalar functions, such as SUBSTRING, evaluate whether this approach fits for the rest of them and if not we can modify it. What do you think?

kaspersorensen: But my concern is actually that in this case you may also have many columns (or rather I prefer to think of it as SelectItems because even those could (one day when we support it) be function calls), for instance:

```
SELECT CONCAT( name, " ", TO_STRING(age), " - ", hometown ) FROM people
```

... but today we might settle on "only" supporting simple selectitems as parameters, so:
```
SELECT CONCAT( name, " ", age, " - ", hometown ) FROM people
```
... that would for me be a reasonable limitation for a first version.

But my worry is that if it's a problem to not have a column in the SelectItem then I'm sure it's also a problem that we have (in the above case) essentially 3 columns referred. I can imagine that this would work mostly then based on coincidence (maybe age and hometown are also materialized by some coincidence rather than conscious design). I know this is maybe making life hard on you, but it feels wrong to go ahead with a solution that feels unstable because of this.

ardlema:
This PR has not been updated for a while and I feel like the approach I took is not the right one so I'm going to close it for now.

PR #86, opened 2016-01-08
METAMODEL-183: Upgrad MongoDB API and seperate MongoDB module into mon…

xmpy:
Take over @loorenzooo 's work. Thank you very much.
Besides loorenzooo's work, I also do following works:
1. Seperate the mongo module into mongo2 and mongo3
     This work follows the discussion in this mail: 
https://mail-archives.apache.org/mod_mbox/metamodel-dev/201511.mbox/%3CCAKXZeCGVSUUNrHYe9Ahbcb_hZ5GH3rNqHBHVUESx4f3DyXDiyQ@mail.gmail.com%3E

         metamodel
              .......
          |-- mongodb
                 |-- commons
                 |-- mongo2
                 |-- mongo3
2. replace all the DBObject by Document and DBObjectList by List<Document>
3. Enable write concern in mongo3 module
4. close MongoClient to escape warnings


mongodb/common/.classpath
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
```
kaspersorensen: Let's not add .classpath or other IDE-specific files to the source control. Please just add a .gitignore entry for it.

mongodb/common/src/main/java/org/apache/metamodel/mongodb/common/MongoDBUtils.java
```diff
@@ -73,5 +101,29 @@ private static Object toValue(Column column, Object value) {
         }
         return value;
     }
+    
+    public static Object covertArrayToList(Object arr) {
```
kaspersorensen: Could we make this method private and just call it in the toValue(...) method instead? Since the method is really restricted in it's use it seems maybe we can add an ```if``` above and then not worry about other people misunderstanding what this is used for.

Small note on naming - it's misspelled (missing "n" in convert).

mongodb/mongo2/pom.xml
```diff
@@ -0,0 +1,70 @@
...
+		</dependency>
+		<dependency>
+            <groupId>org.apache.metamodel</groupId>
+            <artifactId>MetaModel-mongodb-common</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+		<dependency>
+			<groupId>org.mongodb</groupId>
+			<artifactId>mongo-java-driver</artifactId>
+			<version>3.1.0</version>
```
kaspersorensen: I guess the used driver here should be a version 2.x driver?

mongodb/common/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+	<dependencies>
+		<dependency>
+			<groupId>org.apache.metamodel</groupId>
+			<artifactId>MetaModel-core</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>org.mongodb</groupId>
+			<artifactId>mongo-java-driver</artifactId>
+			<version>3.1.0</version>
```
kaspersorensen: I would put this dependency as scope=provided so we don't make the common-module too specific to a particular version (after all stuff in common should work for both 2.x and 3.x)

kaspersorensen:
Thank you for the very good contribution! I have made a few remarks, but general feeling is that it looks good!

xmpy:
@kaspersorensen  Thank you for your remarks  :)

I do followings things in these 3 commits.
1.  remove .classpath
2.  make mongodb-common module work for both 2.x and 3.x
     Before this commit, we have a `public static Row toRow(Document document, DataSetHeader header)` function in the MongoDBUtils in this common module. However, this `Document`, which is the ` org.bson.Document` is only in the mongodb driver 3.x. So I do a little things to remove functions which contain `Document` .

I've added 'scoped' in pom. However, I think maybe there is some risk in it. For developer, in the IDE such as Eclipse, Eclipse will still think we using mongodb 3.x, so developer maybe write wrong functions which can not work in both mongodb 2.x and mongodb 3.x. 

3.  move unclear function convertArrayToList to right place
This function is used for converting array `[1,3,4]` to List. 
Here is the problem code in the test 
    ds = dataContext.query().from(getCollectionName()).select("id").and("name").where("id").in(2, 6, 8, 9)

In the `in` where condition, metamodel will convert `in` parameter to an array object `[2,6,8,9]`. However, in the mongodb 3.x, the query Document class currently supports only List, not support array, like in this bug https://jira.mongodb.org/browse/JAVA-1894.

kaspersorensen:
This PR has my +1 now

For extra sugar on top, I did notice a few minor warnings which could be fixed before we push this to master:

image

xmpy:
I still have a little concern about org.apache.metamodel.DataContextFactory, in this factory, we return the mongo2 MongoDataContext. In the pom of this `full` maven project, we add both `MetaModel-mongodb-mongo2` and `MetaModel-mongodb-mongo3` dependency, but they rely on different version of mongdb driver. It will lead maven package collide. To my best know, the final jar will include the mongodb driver with newest version. Is it right?  :)

kaspersorensen:
Yes let's change the ```full``` module to prefer the mongo3 implementation.

xmpy:
Hi @kaspersorensen , any comments?  :)

kaspersorensen:
Sorry I kinda had to drop off the MetaModel activity shortly because I had some important work to do at work. I'm very satisfied with this PR.

Let's ask for another +1 ... Maybe @losd or @albertostratio or @tomaszguzialek wants to take a quick look?

LosD:
I'm certainly no MongoDB expert, so I may not be the right one to ask, but the code itself looks good to me! :)

kaspersorensen:
I will merge it in - it's certainly an improvement and a great contribution - thank you @xmpy !

PR #89, opened 2016-03-07
METAMODEL-236: Fixed

kaspersorensen:
Fixes the serializability of OperatorType (and FunctionType which I also found had the same issue)


LosD:
+1

PR #90, opened 2016-03-09
Fix METAMODEL-239 | DB2 query rewriting fails when first row = 1 and maxRows = null

ankit2711:


kaspersorensen:
LGTM, although I don't think the added test in `QueryParserTest` is related at all? It's fine and unharmful though, so I will agree to merging that in.

PR #94, opened 2016-04-21
METAMODEL-235: Fixed

kaspersorensen:
This fixes METAMODEL-235 which was essentially about supporting null values and missing values in the REST API of ElasticSearch.

I also expanded the unittests a bit in general around the relevant utility class. It seemed that it didn't have a lot of corner cases covered so now I added a couple at least.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 .project
-.settings/
-.classpath/
+.settings
+.classpath
 .metadata/
 target/
 /.idea/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,6 +1,6 @@
 .project
-.settings/
```
LosD: Isn't .settings a folder?

kaspersorensen: Yes you're right. That should have only been a change for `.classpath`.

LosD: Probably not anything that breaks stuff, though :)

LosD:
LGTM

PR #96, opened 2016-04-22
METAMODEL-244

kaspersorensen:
I wanted to give a shot at fixing METAMODEL-244 so here goes.

I made available a strategy interface for column naming, which I implemented in a few different ways, and created a default delegating strategy as well. It's implemented in the 'fixedwidth' and 'csv' module for now.

I also managed to fix something which I've encountered a few times before, but newer got around to report as an issue: The fact that it is annoying that you sometimes get duplicated column names for CSV files and such. That should rarely be allowed, so I implemented a strategy for renaming those duplicated column names automatically.

I'd like a review if anyone is willing. If you agree that the approach taken is good I will happily also apply it in the excel module. It would be quite easy I think actually.


core/src/main/java/org/apache/metamodel/schema/builder/ColumnNamingStrategy.java
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Provides the name to apply for a given column.
+     * 
+     * @param ctx
+     *            the context of the column naming taking place. This contains
+     *            column index, intrinsic name etc. if available.
+     * @return the name to provide to the column.
+     */
+    public String getNextColumnName(ColumnNamingContext ctx);
```
kaspersorensen: I'm wondering if this method should reside here or in some object that is scoped just for the individual table. Because right now the interface seems to be global fir many tables but implementations typically need state which is scoped only for the duration of naming a single table.

kaspersorensen:
Also implemented this for CSV module now. And added a 'session' abstraction to encapsulate the state that is needed while applying column names for a table.

albertostratio:
@kaspersorensen Do we need all the ColumnNamingStrategy interfaces and implementations stuff to just get the names of the columns. It seems like kinda over-engineering to me.

core/src/main/java/org/apache/metamodel/schema/naming/DefaultColumnNamingStrategy.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package org.apache.metamodel.schema.naming;
+
+import org.apache.metamodel.schema.builder.DelegatingIntrinsicSwitchColumnNamingStrategy;
+
+/**
+ * The default (in most cases) {@link ColumnNamingStrategy} to use when no other
+ * strategy is specified.
+ */
+public class DefaultColumnNamingStrategy extends DelegatingIntrinsicSwitchColumnNamingStrategy {
```
LosD: (repeat of #97 comment, just in the right place :))
That's a pretty weird way to make something a default to me.

This screams for some sort of proper factory.

core/src/main/java/org/apache/metamodel/schema/naming/UniqueColumnNamingStrategy.java
```diff
@@ -0,0 +1,62 @@
...
+            public String getNextColumnName(ColumnNamingContext ctx) {
+                final String intrinsicName = ctx.getIntrinsicColumnName();
+                boolean unique = names.add(intrinsicName);
+                if (unique) {
+                    return intrinsicName;
+                }
+
+                String newName = null;
+                for (int i = 2; !unique; i++) {
+                    newName = intrinsicName + i;
```
LosD: Wouldn't an underscore before be better? Some columns may already end with numbers.

That could lead to endless confusion, e.g. something like "is_iso9001" would become "is_iso90011", which is a different standard (no, I have no good reason why this exact example would ever happen, but it's easily a general problem :))

LosD:
Code mostly looks good. But it is quite a bit more complicated than I expected.

core/src/main/java/org/apache/metamodel/schema/naming/UniqueColumnNamingStrategy.java
kaspersorensen: Agreed, will add an underscore.

core/src/main/java/org/apache/metamodel/schema/naming/DefaultColumnNamingStrategy.java
kaspersorensen: Ok I will make a ColumnNamingStrategies class or such instead

kaspersorensen:
Changed as per review remarks. What do you think now?

LosD:
Looks good! Maybe I'd let the test be DelegatingIntrinsicSwitchColumnNamingStrategyTest. That it is the default is maybe less relevant for the test than what it really is testing. But that's up to you, not a deal breaker for me.

kaspersorensen:
Renamed it. Will merge it now then.

LosD:
Yep. A bit "dangerous" to still get it with `ColumnNamingStrategies.defaultStrategy()`, but nevermind.  :)

PR #98, opened 2016-04-26
ElasticSearch bulk support and mapping creation using Map

kaspersorensen:
This branch adds support for two things... 

 * Creation of elastic search mappings (document types) using a JSON object instead of array. This was discussed on the mailing list in thread "Approach to submitting ElasticSearch mapping".
 * Support for bulk operations in the REST elasticsearch connector

Sorry, it should have been two PRs, but I did it all at once - please let me know if you need it to be split up.

### DIFF
--- a/elasticsearch/common/src/main/java/org/apache/metamodel/elasticsearch/common/ElasticSearchUtils.java
+++ b/elasticsearch/common/src/main/java/org/apache/metamodel/elasticsearch/common/ElasticSearchUtils.java
@@ -86,7 +86,7 @@ public class ElasticSearchUtils {
         return getFilteredQuery("exists", fieldName);
     }
 
-    public static Object getMappingSource(final MutableTable table) {
+    public static Map<String, ?> getMappingSource(final MutableTable table) {
         if (table.getColumnByName(FIELD_ID) == null) {
             final MutableColumn idColumn = new MutableColumn(FIELD_ID, ColumnType.STRING).setTable(table).setPrimaryKey(
                     true);
@@ -96,8 +96,6 @@ public class ElasticSearchUtils {
         final Map<String, Map<String, String>> propertiesMap = new LinkedHashMap<>();
         
         for (Column column : table.getColumns()) {
-            // each column is defined as a property pair of the form: ("field1",
-            // "type=string,store=true")
             final String columnName = column.getName();
             if (FIELD_ID.equals(columnName)) {
                 // do nothing - the ID is a client-side construct
@@ -108,7 +106,6 @@ public class ElasticSearchUtils {
             final Map<String, String> propertyMap = new HashMap<>();
             final String type = getType(column);
             propertyMap.put("type", type);
-//            propertyMap.put("store", "true");
             
             propertiesMap.put(fieldName, propertyMap);
         }
--- a/elasticsearch/common/src/main/java/org/apache/metamodel/elasticsearch/common/ElasticSearchUtils.java
+++ b/elasticsearch/common/src/main/java/org/apache/metamodel/elasticsearch/common/ElasticSearchUtils.java
@@ -21,7 +21,10 @@ package org.apache.metamodel.elasticsearch.common;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 
 import org.apache.metamodel.query.FilterItem;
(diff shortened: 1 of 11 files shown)


kaspersorensen:
Correction:

It's actually also a third thing that is added, namely stripping of illegal field name characters, see METAMODEL-246

elasticsearch/common/src/main/java/org/apache/metamodel/elasticsearch/common/ElasticSearchUtils.java
```diff
@@ -95,17 +103,49 @@ public static QueryBuilder getExistsQuery(String fieldName) {
                 // do nothing - the ID is a client-side construct
                 continue;
             }
-            sourceProperties.add(columnName);
+            
+            final String fieldName = getValidatedFieldName(columnName);
+            final Map<String, String> propertyMap = new HashMap<>();
+            final String type = getType(column);
+            propertyMap.put("type", type);
+//            propertyMap.put("store", "true");
```
LosD: How come this is commented out? Don't we want to ensure storage?

elasticsearch/rest/pom.xml
```diff
@@ -27,7 +27,7 @@ under the License.
 	<modelVersion>4.0.0</modelVersion>
 
 	<properties>
-		<jest.version>0.1.7</jest.version>
+		<jest.version>2.0.2</jest.version>
```
LosD: Oh, nice bump :)

elasticsearch/common/src/main/java/org/apache/metamodel/elasticsearch/common/ElasticSearchUtils.java
```diff
@@ -78,15 +86,15 @@ public static QueryBuilder getExistsQuery(String fieldName) {
         return getFilteredQuery("exists", fieldName);
     }
 
-    public static List<Object> getSourceProperties(final MutableTable table) {
+    public static Object getMappingSource(final MutableTable table) {
```
LosD: Why not `Map` here?

elasticsearch/native/src/main/java/org/apache/metamodel/elasticsearch/nativeclient/ElasticSearchCreateTableBuilder.java
```diff
@@ -44,15 +44,20 @@ public ElasticSearchCreateTableBuilder(ElasticSearchUpdateCallback updateCallbac
...
+        final Object source = ElasticSearchUtils.getMappingSource(table);
 
         final ElasticSearchDataContext dataContext = getUpdateCallback().getDataContext();
         final IndicesAdminClient indicesAdmin = dataContext.getElasticSearchClient().admin().indices();
         final String indexName = dataContext.getIndexName();
 
         final PutMappingRequestBuilder requestBuilder = new PutMappingRequestBuilder(indicesAdmin)
                 .setIndices(indexName).setType(table.getName());
-        requestBuilder.setSource(sourceProperties.toArray());
+        if (source instanceof Map) {
```
LosD: I'm obviously missing something, but it seems to me to always be a `Map`?

LosD:
Code looks good to me, though there was a few surprises.

elasticsearch/common/src/main/java/org/apache/metamodel/elasticsearch/common/ElasticSearchUtils.java
kaspersorensen: Hmm I chose for Object because I felt that the data type was irrelevant - or even an implementation detail. The 'source' object is some kind of magical object anyway without much type safety. But on the other hand, it seems that the `source(Map)` method is different in behaviour from `source(Object)` so that's a good case for changing this method to return map.

elasticsearch/common/src/main/java/org/apache/metamodel/elasticsearch/common/ElasticSearchUtils.java
kaspersorensen: Ah good that you mentioned it. I couldn't find the "store" attribute in newer versions of the ElasticSearch REST documentation [1]. So I think it won't work anymore ... My guess (at least my evidence so far when working with this) is that there isn't an option anymore to not store field values.

[1] https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-put-mapping.html

kaspersorensen:
I've updated this PR. Maybe you can check it again @LosD ?

LosD:
LGTM! :)

PR #99, opened 2016-04-29
METAMODEL-244: Added support for Excel datastore as well

kaspersorensen:
I realised that although METAMODEL-244 was 'fixed', I had not applied the technique to the 'Excel' module. This PR does that.

### DIFF
--- a/couchdb/src/test/java/org/apache/metamodel/couchdb/CouchDbDataContextTest.java
+++ b/couchdb/src/test/java/org/apache/metamodel/couchdb/CouchDbDataContextTest.java
@@ -53,7 +53,8 @@ public class CouchDbDataContextTest extends CouchDbTestCase {
         super.setUp();
 
         if (isConfigured()) {
-            httpClient = new StdHttpClient.Builder().host(getHostname()).build();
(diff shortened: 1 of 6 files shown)


couchdb/src/test/java/org/apache/metamodel/couchdb/CouchDbDataContextTest.java
```diff
@@ -53,7 +53,8 @@ protected void setUp() throws Exception {
         super.setUp();
 
         if (isConfigured()) {
-            httpClient = new StdHttpClient.Builder().host(getHostname()).build();
+            final int timeout = 8 * 1000; // 8 seconds should be more than enough
```
LosD: `public static final SOCKET_TIMEOUT`?

(I have a great hate for socket timeout settings, because they always seems so arbitrary, and it's hard to find a good number, especially in our case where we really don't know what the use case will be... I guess that's a sign that we at some point should make these user configurable)

LosD: - D'oh, and this is a test. Scratch both parts of that comment :)

LosD:
LGTM!

PR #100, opened 2016-05-12
Do already-loaded check before retrieving connection.

LosD:
This avoids an issue with excessive retrieving/returning Connections introduced in METAMODEL-142

Fixes METAMODEL-252

Tested a table copy using DataCleaner on a SQL Server table of 8 varchar(max) columns with 500000 entries: Goes from 69 seconds to 33 seconds.


kaspersorensen:
+1 from me.

PR #101, opened 2016-05-16
Removed all deprecated API from MetaModel (aaaaah)

kaspersorensen:
This is a PR for the 5.x branch. I suggest getting rid of all the deprecated methods of metamodel for that release.


hsaputra:
@kaspersorensen just wondering we want to remove the deprecated methods in 5.x branch?
Shouldn't we just do it in master?

kaspersorensen:
We could do it in master, but I guess it seems nicer to only do that sort of thing when it aligns on a bigger release.

LosD:
Are all the deprecations old-ish? It might be a major release, but if we can avoid throwing out stuff that has been deprecated only a few months ago, it would be best.

kaspersorensen:
The most recent deprecations where introduced in MM 4.4 AFAIK. Those where related to the split of FunctionType into AggregateFunction and ScalarFunction interfaces. So the deprecations have been there since the release date of 4.4.0: 2015-10-12.

LosD:
Half a year is probably reasonable when we don't have an official policy about keeping depreciations for a certain amount of time.

hsaputra:
The only worry I have is that this change introduce API changes that could break compilation for minor release

kaspersorensen:
I agree Henry, that's why I wanted to put it on the 5.0 branch so that the expectation is also managed a bit :-)

hsaputra:
Maybe we should do shout out to dev@ list to check if anyone still in 5.0 and ok to add this breaking API changes?

kaspersorensen:
I'm not sure I understand what you mean Henry? 5.0 isn't out yet so I guess no one will be depending on it yet?

hsaputra:
Ah I am so sorry, I need more dose on caffeine =(
I apologize for the confusion.

What I meant was, would anyone in 4.0 still need it to be around for 5.0.

kaspersorensen:
Aha yes... Let's ask that.

PR #102, opened 2016-05-16
METAMODEL-241: Fixed

kaspersorensen:
This fix is implemented by further diving deep into the depths of OpenJDK. I'm not really fond of going down that route, but it seems to be the way to go if we really want to keep deserialization support.

Thanks to @LosD for the good hint on what had changed in OpenJDK by the way!


LosD:
Oh wow that is nasty. But if that's what it takes, that's what it takes. 

LGTM!

PR #103, opened 2016-05-20
FixedWidth EBCDIC support

jhorcicka:
METAMODEL-250 FixedWidth EBCDIC support

* Refactoring of FixedWidthReader. 
* BufferedReader was replaced by BufferedInputStream (support for files without new-line characters). 
* New boolean options "headerPresent" and "eolPresent". 
* Simple fixed-width EBCDIC file in test/resources.


LosD:
Travis failure seems unrelated (a failed Cassandra startup)

fixedwidth/src/main/java/org/apache/metamodel/fixedwidth/FixedWidthConfiguration.java
```diff
@@ -54,45 +56,49 @@ public FixedWidthConfiguration(int fixedValueWidth) {
...
 
     public FixedWidthConfiguration(int columnNameLineNumber, String encoding, int fixedValueWidth,
-            boolean failOnInconsistentLineWidth) {
+            boolean failOnInconsistentLineWidth, boolean headerPresent, boolean eolPresent) {
         this.encoding = encoding;
         this.fixedValueWidth = fixedValueWidth;
         this.columnNameLineNumber = columnNameLineNumber;
         this.failOnInconsistentLineWidth = failOnInconsistentLineWidth;
+		this.headerPresent = headerPresent;
+		this.eolPresent = eolPresent;
```
LosD: Something weird happening with the whitespace here

fixedwidth/src/main/java/org/apache/metamodel/fixedwidth/FixedWidthConfiguration.java
```diff
@@ -54,45 +56,49 @@ public FixedWidthConfiguration(int fixedValueWidth) {
...
     
     public FixedWidthConfiguration(int columnNameLineNumber, ColumnNamingStrategy columnNamingStrategy, String encoding,
-            int[] valueWidths, boolean failOnInconsistentLineWidth) {
+            int[] valueWidths, boolean failOnInconsistentLineWidth, boolean headerPresent, boolean eolPresent) {
         this.encoding = encoding;
         this.fixedValueWidth = -1;
         this.columnNameLineNumber = columnNameLineNumber;
         this.failOnInconsistentLineWidth = failOnInconsistentLineWidth;
+		this.headerPresent = headerPresent;
+		this.eolPresent = eolPresent;
```
LosD: And here

fixedwidth/src/main/java/org/apache/metamodel/fixedwidth/FixedWidthConfiguration.java
```diff
@@ -103,6 +109,8 @@ public FixedWidthConfiguration(String encoding, List<FixedWidthColumnSpec> colum
             valueWidths[i] = columnSpecs.get(i).getWidth();
         }
         this.failOnInconsistentLineWidth = failOnInconsistentLineWidth;
+		this.headerPresent = headerPresent;
+		this.eolPresent = eolPresent;
```
LosD: And here

fixedwidth/src/main/java/org/apache/metamodel/fixedwidth/FixedWidthReader.java
```diff
@@ -31,156 +31,261 @@
...
+                correctedResult[i] = result[i];
+            }
+
+            result = correctedResult;
+        }
+
+        return result;
+    }
+
+    private InconsistentValueWidthException buildInconsistentValueWidthException(String recordData, String[] result,
```
kaspersorensen: This method seems backwards to me. It is called "build[something]" but will usually (in the good cases) return null. And then it is evaluated later on to be thrown if it is not null. Why this pattern is needed I don't know. Seems a much more normal way to do it was to have something like a "validate[something]" which will throw the exception if it is desirable.

fixedwidth/src/test/java/org/apache/metamodel/fixedwidth/FixedWidthConfigurationTest.java
```diff
@@ -27,18 +27,18 @@
...
 		assertEquals(
 				"FixedWidthConfiguration[encoding=UTF8, fixedValueWidth=10, valueWidths=[], columnNameLineNumber=1, failOnInconsistentLineWidth=true]",
-				new FixedWidthConfiguration(1, "UTF8", 10, true).toString());
+				new FixedWidthConfiguration(1, "UTF8", 10, true, false, true).toString());
 	}
 
 	public void testEquals() throws Exception {
 		FixedWidthConfiguration conf1 = new FixedWidthConfiguration(1, "UTF8",
-				10, true);
+				10, true, false, true);
```
kaspersorensen: Please make sure there are overloaded constructors available for the old constructor signature. Then you also wont have to change this existing test.

fixedwidth/src/main/java/org/apache/metamodel/fixedwidth/FixedWidthConfiguration.java
```diff
@@ -33,66 +33,74 @@
...
-	private final int[] valueWidths;
-	private final int columnNameLineNumber;
-	private final boolean failOnInconsistentLineWidth;
-	private final ColumnNamingStrategy columnNamingStrategy;
+    private final String encoding;
+    private final int fixedValueWidth;
+    private final int[] valueWidths;
+    private final int columnNameLineNumber;
+    private final boolean failOnInconsistentLineWidth;
+    private final boolean headerPresent;
```
kaspersorensen: This new property is redundant (and thus potentially 'dirty') because `columnNameLineNumber` holds the same metadata. If it is 0 (`NO_COLUMN_NAME_LINE `), then it means that there is no header.

jhorcicka: I am not so sure about that. I can image it is confusing (let's create a better name then) but this is something else. In EBCDIC files there might be a "binary header" with some meta-data that we want to skip before we can read actual data. And the actual data can (perhaps) contain the "normal" header with column names. I am not sure if this can really happen since there are "format" files for structure and column names definition.  

Example:
1. HDR 2016-06-09 ... // this is "headerPresent" 
[2. NAME, SURNAME, YEAR, ...]? // this is "columnNameLineNumber"
3. John, Doe, 1980, ...

kaspersorensen: Ah right, but maybe in that case you would then simply set columnNameLineNumber to 2 because its on the second line?

LosD: I'm not sure, but I think the "lines"  are not real lines, but the EBCDIC header followed by lots of whitespace, and that it discards until it sees the real data

LosD: (might be whitespace in the length of a line though)

jhorcicka: @LosD: I do it generally by skipping whitespace until I reach data but in the case of "the one" file we had it was just the length of one record. 

@kaspersorensen: Yes, I could set columnNameLineNumber to 2 but I am not sure if "a common user" would know to do this too. Unless (s)he knows the details of the file which I assume we can not expect. On the other hand if there is "column-names-header" then (s)he has to be aware of it anyway to use it properly. I don't have a strong opinion here and will go with "whatever" you advice. :-)

kaspersorensen: I guess what I want to avoid is just that the two properties are redundant or can be conflicting. For instance of `columnNameLineNumber` is `0` and `headerPresent` is `true` then they communicate the opposite fact to the user.

So maybe it's rather just a matter of `headerPresent` needing a rename? Can we name it in a way where it's meaning is less about the header and more about the padded whitespace maybe?

jhorcicka: Yes, let's give it a better name. I am not too much into "padding" since some padding is actually in all rows. Basically this "binary header" contains some additional data about the file, so what about `metadataPresent`? Would that be clear to the users?

kaspersorensen: Hmm I'm really at a loss because deep inside I feel that this is a very particular thing to EBCDIC and therefore I cannot come up with a fitting generic name. And so maybe we should just give it a name that reveals that it is a particular thing. Such as `skipEbcdicHeader` or whatever. Or maybe it is less particular than that, but then I am still not sure what it is! :)

I would almost consider subtyping the whole configuration class to have a EbcdicConfiguration class instead of FixedWidthConfiguration.

jhorcicka: I would go with ```skipEbcdicHeader``` then. I also feel that EBCDIC is something special in the context of fixed-width but let's solve this within a new issue as a refactoring part. There already are other projects waiting for this to be merged and this kind of communication takes a lot of time. 

So, let's change the name to ```skipEbcdicHeader```, focus on other aspects of this PR and then merge it. Then we can start a new discussion about separation of EbcdicConfiguration. What do you think?

kaspersorensen: I would actually then prefer that we subclass it. Remove the 'final' part of the FixedWidthConfiguration and make an EbcdicConfiguration class that extends it. That way we don't introduce something into this class that is still not a very well-defined thing. For that you have my green :)

jhorcicka:
I separated EbcdicConfiguration (extends FixedWidthConfiguration) and EbcdicReader (extends FixedWidthReader).

kaspersorensen:
Test failed on Travis, but it seems unrelated.
(Although quite interesting test failure that I'm gonna report into JIRA to look at.)

kaspersorensen:
+1 from me, let's merge soon so that we get this into the next release - it's been dangling for too long :)

fixedwidth/src/main/java/org/apache/metamodel/fixedwidth/FixedWidthReader.java
```diff
@@ -18,78 +18,235 @@
...
+
+        for (ch = _stream.read(); !isEndingCharacter(ch); ch = _stream.read()) {
+            line.append((char) ch);
+        }
+
+        if (ch == CARRIAGE_RETURN) {
+            readLineFeedIfFollows();
+        }
+
+        return (line.length()) > 0 ? line.toString() : null;
```
jhorcicka: Here, the returned String is charset-less which is actually a bug for diacritics input. PR for this bug fix is here: https://github.com/apache/metamodel/pull/121

PR #104, opened 2016-05-20
Fix for encoding error when using InputStreams

bmehner:
Fixes https://issues.apache.org/jira/browse/METAMODEL-1086

### DIFF
--- a/csv/src/main/java/org/apache/metamodel/csv/CsvDataContext.java
+++ b/csv/src/main/java/org/apache/metamodel/csv/CsvDataContext.java
@@ -22,7 +22,6 @@ import au.com.bytecode.opencsv.CSVParser;
 import au.com.bytecode.opencsv.CSVReader;
 import java.io.*;
 import java.net.URL;
-import java.nio.charset.Charset;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


LosD:
Travis error seems unrelated.

csv/src/main/java/org/apache/metamodel/csv/CsvDataContext.java
```diff
@@ -244,7 +227,7 @@ private static File createFileFromInputStream(InputStream inputStream, String en
         file = fileCandidate;
 
         final BufferedWriter writer = FileHelper.getBufferedWriter(file, encoding);
-        final BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream));
+        final BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream, Charset.forName(encoding)));
```
kaspersorensen: For the sake of consistency, would it be nicer to use `FileHelper.getBufferedReader(inputStream, encoding)`?

kaspersorensen:
Sorry for the long wait - I didn't realize that you had changed the branch since my remark.

+1 and let's merge! :)

PR #105, opened 2016-05-23
Feature/small changes to fixed width

ClaudiaPHI:
Modified FixedWidthReader so that we can feed a line to a reader independently. 

The modification were necessary for integrating fixed width files on Hadoop: https://github.com/datacleaner/DataCleaner/issues/1343

### DIFF
--- a/fixedwidth/src/main/java/org/apache/metamodel/fixedwidth/FixedWidthReader.java
+++ b/fixedwidth/src/main/java/org/apache/metamodel/fixedwidth/FixedWidthReader.java
@@ -30,7 +30,7 @@ import java.util.List;
 /**
  * Reader capable of separating values based on a fixed width setting.
  */
-final class FixedWidthReader implements Closeable {
+final public class FixedWidthReader implements Closeable {
 
 	private final BufferedReader _reader;
 	private final int _fixedValueWidth;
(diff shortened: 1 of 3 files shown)


kaspersorensen:
I feel a bit like I am missing a unit test for this kind of usage.
An I wonder if that will reveal a slight abstraction issue... The reader today takes a Resource as a constructor argument. Would you have any resource to provide it with when you are yourself providing the lines of data?

ClaudiaPHI:
I am providing the lines of data. I need to be able to use this class without the BufferReader. I will add a unit test.

ClaudiaPHI:
I need to wait for this PR: https://github.com/apache/metamodel/pull/103 if I am to avoid conflicts.

kaspersorensen:
+0

Code is acceptable from me, but I think it would be better to separate "parser" from "reader" here. A reader is a stateful thing IMO - it has a pointer/cursor to some data that it is reading through and it is also closeable. The parsing part however can have a different scope or maybe even be completely stateless. So I would prefer if we make two different classes for it (the reader will probably use the parser, but not the other way around).

kaspersorensen:
OK let's merge this and maybe some day improve it as per my remarks.

PR #106, opened 2016-05-24
METAMODEL-1088 Fix aliases for MongoDB modules

ardlema:
This PR includes the changes to fix METAMODEL-1088 by checking whether the columns contain aliases and if so just delegating the query to the QueryPostprocessDataContext.

### DIFF
--- a/mongodb/mongo2/src/main/java/org/apache/metamodel/mongodb/mongo2/MongoDbDataContext.java
+++ b/mongodb/mongo2/src/main/java/org/apache/metamodel/mongodb/mongo2/MongoDbDataContext.java
@@ -280,20 +280,23 @@ public class MongoDbDataContext extends QueryPostprocessDataContext implements U
                 final List<FilterItem> whereItems = query.getWhereClause().getItems();
 
                 // if all of the select items are "pure" column selection
-                boolean allSelectItemsAreColumns = true;
+                boolean allSelectItemsAreColumnsWithoutAliases = true;
                 List<SelectItem> selectItems = query.getSelectClause().getItems();
 
                 // if it is a
                 // "SELECT [columns] FROM [table] WHERE [conditions]"
                 // query.
                 for (SelectItem selectItem : selectItems) {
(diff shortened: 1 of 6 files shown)


LosD:
I know nothing about Mongodb, but code looks good :)

It does uncover one (unrelated) thing, though: It seems Mongo2 & 3 modules are _very_ similar (a qucik glance shows no difference in your changes for 2 & 3), which makes me think more more stuff might belong in common?

ardlema:
Completely agree @LosD. When applying the changes I thought the same. I was about to start refactoring the code but I thought that I shouldn't do this within this PR.

I'd open a new jira to sort this out, what do you think?

LosD:
Yeah, that makes sense.

ardlema:
Jira created: [METAMODEL-1089](https://issues.apache.org/jira/browse/METAMODEL-1089)

Thank you @LosD

albertostratio:
OK to merge? Anyone?

LosD:
+1 from me, but again my MongoDB knowledge is pretty limited. :)

albertostratio:
Thank you! I'll wait another day for anyone who like to comment on this and then will merge it

kaspersorensen:
+1 from me too.

A side-note: Have anyone checked if the same fix can be applied to e.g. CouchDB or other data context implementations? I think a similar approach of using that `allSelectItemsAreColumns` flag is used in other places.

kaspersorensen:
Wait what ... Hmm I see that this is actually a fix that works functionally, but will mean that if you use an alias then the query will not be optimized at all. Is that what we want? I can live with it, since functionality is more important than non-functionals. But some people could take a big performance hit over this if they use an alias. Would be better to fix that too.

albertostratio:
I agree @kaspersorensen. I've just pushed some changes to optimize the query even though it contains aliases.

kaspersorensen:
Great, +1 from me!

albertostratio:
Cool! Thank you @kaspersorensen and @LosD. I'll merge this.

PR #107, opened 2016-06-03
METAMODEL-1093 close compiled ResultSets

LosD:
This makes sure that ResultSets are closed JdbcDataSet itselfs closes. Also refactors closing of Statements and ResultsSets to use try-with-resources as much as possible. This also moves the responsibility of closing Statements and ResultSets away from JdbcDataContext, which seems to belong more to the caller, since the context isn't the one to actually track them even though it creates them (and it is of course a requirement for using try-with-resources).

Fixes METAMODEL-1093

### DIFF
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataSet.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataSet.java
@@ -197,8 +197,12 @@ final class JdbcDataSet extends AbstractDataSet {
         if (_closed) {
             return;
         }
+
+        FileHelper.safeClose(_resultSet);
+
         if (_jdbcDataContext != null) {
-            _jdbcDataContext.close(_connection, _resultSet, _statement);
+            FileHelper.safeClose(_statement);
+            _jdbcDataContext.close(_connection);
         }
         if (_compiledQuery != null) {
             _compiledQuery.returnLease(_lease);
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcMetadataLoader.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcMetadataLoader.java
@@ -22,7 +22,6 @@ import java.sql.Connection;
 import java.sql.DatabaseMetaData;
 import java.sql.ResultSet;
 import java.sql.SQLException;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.Set;
 import java.util.StringTokenizer;
@@ -75,7 +74,7 @@ final class JdbcMetadataLoader implements MetadataLoader {
         try {
             loadTables(schema, connection);
         } finally {
-            _dataContext.close(connection, null, null);
+            _dataContext.close(connection);
         }
     }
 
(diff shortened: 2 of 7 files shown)


LosD:
Travis error seems unrelated (CouchDB timeout)

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcMetadataLoader.java
```diff
@@ -92,20 +91,27 @@ public void loadTables(JdbcSchema schema, Connection connection) {
         }
     }
 
+    private String getSchemaName(Schema schema) {
```
kaspersorensen: This method name here can be a bit confusing because there is both a concept of "schema" in JDBC and in MetaModel. And the unfortunate thing is that they aren't 100% aligned, which is basically what the _usesCatalogAsSchemas flag is used to determine.

I would rename it to "getJdbcSchemaName" to avoid that confusion.

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataSet.java
```diff
@@ -197,8 +197,12 @@ public void close() {
         if (_closed) {
             return;
         }
+
+        FileHelper.safeClose(_resultSet);
+
         if (_jdbcDataContext != null) {
-            _jdbcDataContext.close(_connection, _resultSet, _statement);
+            FileHelper.safeClose(_statement);
```
kaspersorensen: Shouldn't this be placed outside of the `if`, just like the closing of the ResultSet above?

kaspersorensen:
I read your JIRA issue description, and if you want to reproduce in a unittest, you _could_ do that in the MySQL integration test maybe?

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataSet.java
LosD: No, we can't close the statement unconditionally, it makes almost every single test fail. I'm not entirely sure why (didn't really pursue it, honestly).

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcMetadataLoader.java
LosD: Makes sense

LosD:
I would love to, but I'm almost sure it won't trigger the error. I tried making a test that almost copied exactly what the DataCleaner table lookup analyzer does, but still didn't seem to trigger the issue. Didn't do it inside the integration setup though, I'll see if that makes a difference.

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataSet.java
kaspersorensen: OK :)

LosD: Ah. I think I know: The statement for compiled queries can't be closed with the dataset, then next execution will fail.

LosD:
I added a test (and did some general fixing up... The test assumed that non-system tables besides Sakila existed, and updated the MySQL driver, since the old one had problems with milli/microsecond precision).

Unfortunately, my test does _not_ break on MySQL. I simply can't get it to fail. Still have no idea why it breaks in some situations and not others.

kaspersorensen:
But travis seems to fail somehow now?
```
Tests in error: 
  MysqlTest.testCompiledQueries:417->AbstractJdbIntegrationTest.getConnection:104 Â» IllegalState
```

LosD:
My bad, forget the silly integration test pattern we use :)

LosD:
(By the way could it be possible to get Travis to use some of these databases? Every time I use one of the integration tests, it seems to have bitrottet quite a bit because no one ever runs them)

kaspersorensen:
It would be nice yes. I know travis has support for various DBs like CouchDB and so on (which we use). Maybe take a look if MySQL is also there. If it is, let's make a story about implementing automatic tests of it.

LosD:
Just checked. It (and Postgres) is there. 

A danger may be memory use.

kaspersorensen:
I doubt that will be too bad. But anyways, will you create the issue, then you have my LGTM on this PR :)

LosD:
Here you go: https://issues.apache.org/jira/browse/METAMODEL-1096 :)

kaspersorensen:
LGTM! :)

PR #109, opened 2016-06-08
Improving writing to an excel file

ClaudiaPHI:
In DataCleaner we have a performance issue when trying to write an Excel file:  https://github.com/datacleaner/DataCleaner/issues/1370 

I removed the workaround to "first write to a temp file to avoid that workbook source is the same as the target (will cause read+write cyclic overflow)"
The "InMemoryResource" is not needed anymore after the change I have made.

### DIFF
--- a/excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
+++ b/excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
@@ -19,6 +19,7 @@
 package org.apache.metamodel.excel;
 
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.text.NumberFormat;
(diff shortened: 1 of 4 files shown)


excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -153,20 +155,15 @@ public static Workbook readWorkbook(ExcelDataContext dataContext) {
      * @param wb
      */
     public static void writeAndCloseWorkbook(ExcelDataContext dataContext, final Workbook wb) {
-        // first write to a temp file to avoid that workbook source is the same
-        // as the target (will cause read+write cyclic overflow)
```
kaspersorensen: I think the gist of this is just that you're disregarding this comment now? That would cause an old bug to reoccur.

full/src/main/java/org/apache/metamodel/DataContextFactory.java
```diff
@@ -571,6 +571,8 @@ public static UpdateableDataContext createMongoDbDataContext(String hostname, In
                 throw (RuntimeException) e;
             }
             throw new IllegalStateException(e);
+        } finally {
+            mongoClient.close();
```
kaspersorensen: It should not be closed. That would leave the created DataContext unusable.

kaspersorensen:
-1, I don't think these changes are viable. Please see review remark.

PR #110, opened 2016-06-13
Rest api - MetaModel-as-a-Service - draft application

kaspersorensen:
Hi all,

This is my attempt at building a 'MetaModel-as-a-Service' webapp. Although it is not yet as great as I imagine it should be before we release it, I do think there's a considerable amount of work here to go with, and that might trigger others who want to help with the finishing touches.

JIRA issue, for reference: https://issues.apache.org/jira/browse/METAMODEL-1097

This webapp allows the user to define data sources in a number of groups (in the code I used the word 'tenant' because I imagine usually each group would be a tenant of the application). The data source can then be explored, queried and updated using RESTful endpoints. For example:

 * `PUT /t1` would register a new tenant with name 't1'.
 * `PUT /t1/ds` would register a new data source with name 'ds'.
 * `GET /t1/ds` would show information about 'ds', including it's schema names.
 * `GET /t1/ds/s/public` would show information about the schema 'public'.
 * `GET /t1/ds/s/public/t/customers` would show information about the table 'customers' in schema 'public'.
 * `POST /t1/ds/s/public/t/customers` would insert data into the 'customers' table.
 * `GET /t1/ds/q?sql=SELECT+name+FROM+customers` would fire the given query to the data source.

(see the controllers package for all endpoints).

Here's what is missing (as far as I can think of) before I think it is _really_ releasable:

 - No formal interface documentation yet (I think we should add a swagger file).
 - No persistence of tenant or datasource information (I imagine we can have many ways of implementing this - the simplest would simply be a folder/file based persistence of tenant folders and data source folders / JSON files).
 - So far only a simple datasource definition reader for POJO data (but others should be very easy to add - I suggest adopting the `MetaModel-spring` module's approach to this).

Would love a review and discussion, and would invite anyone else to further contribute to making this releasable.


kaspersorensen:
Seems it has kept your (somewhat old) remarks from the early commits @LosD. Maybe you can do me a favor and delete them so we can have a fresh view on the PR? The EOLs where fixed and you're more than welcome to post other remarks too :)

LosD:
I've seen this before for EOLs. I guess it's because it confiders it another line

LosD:
By the way, API Blueprint seems to be a good alternative to Swagger. Then apiary.io can be used as a mock server.

LosD:
I'd maybe use POST for the SQL-like query, so UPDATE and INSERT can be performed safely.

kaspersorensen:
We do not allow UPDATE or INSERT in the query interface. It's literally just a facade for `DataContext.executeQuery(String)` (which is just for SELECT queries).

LosD:
That settles that, then ☺

kaspersorensen:
This branch has now been updated with the new SPI provided in METAMODEL-1099. Shall we merge this (into 5.x)? Please vote :)

kaspersorensen:
Closing this and #111 in order to create a new PR without all the commit-bound remarks.

PR #112, opened 2016-06-27
Added support for postgresql's "json" and "jsonb" data types -> MAP

kaspersorensen:
Hi all,

Here's a PR that adds MetaModel support for PostgreSQL's JSON based data types. I'm currently on a project where we use this data type and unfortunately MetaModel's support for it is very bad (breaks upon insert/update, queries/returns it as a string). I think that it makes most sense to treat the data type as a logical `MAP` type. Hope you agree.

To do this I had to do a refactor of moving static methods from `JdbcUtils` to non-static methods on the `IQueryRewriter` interface. That way those methods can be overridden and apply database-specific behaviour.

### DIFF
--- a/jdbc/pom.xml
+++ b/jdbc/pom.xml
@@ -75,7 +75,18 @@
 			<groupId>org.postgresql</groupId>
 			<artifactId>postgresql</artifactId>
 			<version>9.3-1104-jdbc4</version>
-			<scope>test</scope>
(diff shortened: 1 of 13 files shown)


kaspersorensen:
Fixed unittest, please review.

kaspersorensen:
Merging based on lazy consensus.

PR #113, opened 2016-07-04
Dynamic DataContext factory (MM 1099)

kaspersorensen:
This is my attempt at solving METAMODEL-1099. I've leveraged the standard [ServiceLoader](https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html) for registering resource factories and data context factories. And I implemented this pattern with the various standard resources as well as the CSV data context. In my opinion this approach is pretty good, but I'd like to hear your feedback!

If you think this approach is good, we should obviously expand it to fit all the datacontext implementations. But that would be a rather trivial task I think, once we agree on this.

### DIFF
--- a/core/src/main/java/org/apache/metamodel/util/ResourceUtils.java
+++ b/core/src/main/java/org/apache/metamodel/util/ResourceUtils.java
@@ -18,10 +18,30 @@
  */
 package org.apache.metamodel.util;
 
+import java.net.URI;
+
+import org.apache.metamodel.factory.ResourceFactoryRegistryImpl;
+import org.apache.metamodel.factory.ResourceProperties;
+import org.apache.metamodel.factory.SimpleResourceProperties;
+import org.apache.metamodel.factory.UnsupportedResourcePropertiesException;
+
 /**
  * Static utility methods for handling {@link Resource}s.
  */
 public class ResourceUtils {
+
+    public static Resource toResource(URI uri) {
+        return toResource(new SimpleResourceProperties(uri));
+    }
+
+    public static Resource toResource(String uri) {
+        return toResource(new SimpleResourceProperties(uri));
+    }
+
+    public static Resource toResource(ResourceProperties resourceProperties)
+            throws UnsupportedResourcePropertiesException {
+        return ResourceFactoryRegistryImpl.getDefaultInstance().createResource(resourceProperties);
+    }
 
     /**
      * Gets the parent name of a resource. For example, if the resource's
--- a/core/src/main/java/org/apache/metamodel/util/ResourceUtils.java
+++ b/core/src/main/java/org/apache/metamodel/util/ResourceUtils.java
@@ -30,14 +30,39 @@ import org.apache.metamodel.factory.UnsupportedResourcePropertiesException;
  */
 public class ResourceUtils {
 
-    public static Resource toResource(URI uri) {
+    /**
+     * Creates a Resource based on a URI
+     * 
+     * @param uri
+     * @return
+     * @throws UnsupportedResourcePropertiesException
+     *             if the scheme or other part of the URI is unsupported.
+     */
+    public static Resource toResource(URI uri) throws UnsupportedResourcePropertiesException {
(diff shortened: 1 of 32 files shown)


tomaszguzialek:
Looks solid.

kaspersorensen:
Ok I'm gonna continue down this path then ...

kaspersorensen:
Added a JDBC factory too and fixed a bunch of the TODOs in the code. I suggest taking the PR as a candidate for merging now as it is quite a lot of work to make all the factories in one PR and I'd love to offer those as independent starter issues for people new to MetaModel.

LosD:
`throws UnsupportedResourcePropertiesException` or not `throws UnsupportedResourcePropertiesException` seems more than a little bit chaotic. In some cases there are methods that calls a method which has `throws UnsupportedResourcePropertiesException` directly, without adding that to its own signature.

That is pretty confusing... In my mind, either `UnsupportedResourcePropertiesException` should be a checked exception (I know, I know... Will never happen), or not declared. This is just chaos.

core/src/main/java/org/apache/metamodel/util/ResourceUtils.java
```diff
@@ -18,11 +18,31 @@
...
+    public static Resource toResource(URI uri) {
+        return toResource(new SimpleResourceProperties(uri));
+    }
+
+    public static Resource toResource(String uri) {
+        return toResource(new SimpleResourceProperties(uri));
+    }
+
+    public static Resource toResource(ResourceProperties resourceProperties)
+            throws UnsupportedResourcePropertiesException {
```
LosD: Here's one of the examples I talked about. Why do this declared the throwing, when the overloaded methods (just above) does not, even though they just call this (along with a constructor)?

LosD:
Overall idea is brilliant (of course I instantly think "Hey, OSGi ServiceFactory!", but this is close enough :smile:).

I'm _very_ happy nothing ended up in MetaModel-full as initially suggested in METAMODEL-1099, as that is something I think we should deprecate because of all the dependency issues it instantly brings to anyone using it. This might even make MM-full less interesting.

core/src/main/java/org/apache/metamodel/util/ResourceUtils.java
kaspersorensen: Ah right, but that's IMO more an issue with the overloaded methods :)
Maybe it's a bit of a religious discussion, but I would refer to some good (IMO) stackoverflow answers on the topic here:

http://stackoverflow.com/questions/20871763/should-i-declare-unchecked-exceptions-in-the-throws-specification

http://stackoverflow.com/questions/25743574/java-best-practice-for-declaring-unchecked-exception

LosD: Joshua Block's _Effective Java_ item 62 is pretty clear: 
> Use the JavaDoc `@throws` tag to document each unchecked exception that a method can throw, but do _not_ use the `throws` keyword to include unchecked exceptions in the method declaration

Of course that doesn't really solve the "how far down the chain should you go", but at least it doesn't look like a checked exception.

Regarding how far to document these, I think it's a good thing to do for stuff like the almost direct overloads here, and also methods closely coupled tightly enough that it would make sense.

kaspersorensen: Ok so there are arguments on both sides here. But then I can throw another one: For consistency it makes sense to use the 'throws' keyword since that's what is being done in other core interfaces of MetaModel, including the most core of all: `DataContext` (has many `throws MetaModelException` declarations).

LosD: Hard to argue with consistency, no matter how bad that consistency is ☺

kaspersorensen:
Updated the ResourceUtils according to review feedback/discussion - added javadocs to explain why/when the exception can be expected.

kaspersorensen:
OK I will merge it in as per lazy consensus in terms of actual votes cast, and positive feedback remarks.

PR #114, opened 2016-07-11
Created a parser for fixed width

ClaudiaPHI:
Created a Parser class for FixedWidthReader.  I moved the parser method to it's own class.

### DIFF
--- a/fixedwidth/src/main/java/org/apache/metamodel/fixedwidth/FixedWidthLineParser.java
+++ b/fixedwidth/src/main/java/org/apache/metamodel/fixedwidth/FixedWidthLineParser.java
@@ -1,0 +1,111 @@
+package org.apache.metamodel.fixedwidth;
+
+import java.io.IOException;
+import java.text.CharacterIterator;
+import java.text.StringCharacterIterator;
+import java.util.ArrayList;
+import java.util.List;
+
+public class FixedWidthLineParser {
+ 
+    private final int _fixedValueWidth;
+    private final int[] _valueWidths;
+    private final boolean _failOnInconsistentLineWidth;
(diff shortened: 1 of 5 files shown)


LosD:
Travis is failing because you forgot the license header for the new class.

LosD:
I guess this makes no functional difference to anyone using the the old reader, it just opens up the parser for usage from outside the reader?

My only (minor) objection is that unit testing the parser should probably not be part of the test of old reader, but a separate test for the new class, and the old unbuffered test should still be in place to make sure that use hasn't been broken.

fixedwidth/src/main/java/org/apache/metamodel/fixedwidth/FixedWidthLineParser.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class FixedWidthLineParser {
+ 
+    private final int _fixedValueWidth;
+    private final int[] _valueWidths;
+    private final boolean _failOnInconsistentLineWidth;
+    private final int _expectedLineLength;
+    private final boolean _constantWidth;
```
kaspersorensen: The 5 above fields are all represented with one `FixedWidthConfiguration` object. It's not important to me, but I think it could reduce complexity for you to reuse that class instead of calling out these individual fields.

kaspersorensen:
LGTM

LosD:
LGTM!

ClaudiaPHI:
I have created the issue https://issues.apache.org/jira/browse/METAMODEL-1102 and added the improvements suggested by Kasper.

PR #115, opened 2016-07-12
Anchor WildcardPattern to start

LosD:
This makes sure the `WildcardPattern` anchors its test to the start of the string, not only the end.

This comes with a minor performance penalty (quick test said 5-10%), but I think correctness trumps speed here.

I would have liked to compare speed with pre-compiled regexes (`Pattern.quote()` the operand, replace `%` with `.*?` and anchor it with `^` and `$`), but if the LIKE operand is inside a `SelectItem`, that would mean compiling a new regex pattern for every row, which would probably be _very_ expensive, so I opted not to try. There seem to be some similar optimization of the IN operator, so I might be wrong, though.

Fixes METAMODEL-1103

### DIFF
--- a/core/src/main/java/org/apache/metamodel/util/WildcardPattern.java
+++ b/core/src/main/java/org/apache/metamodel/util/WildcardPattern.java
@@ -40,8 +40,12 @@ public final class WildcardPattern implements Serializable {
 	public WildcardPattern(String pattern, char wildcard) {
 		_pattern = pattern;
 		_wildcard = wildcard;
-		_startsWithDelim = (_pattern.charAt(0) == _wildcard);
-		_endsWithDelim = (_pattern.charAt(pattern.length() - 1) == _wildcard);
+		if(_pattern.isEmpty()){
+			_startsWithDelim = _endsWithDelim = false;
+		} else {
+			_startsWithDelim = _pattern.charAt(0) == _wildcard;
+			_endsWithDelim = _pattern.charAt(pattern.length() - 1) == _wildcard;
+		}
 	}
 
 	public boolean matches(String value) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/metamodel/util/WildcardPattern.java
```diff
@@ -32,13 +32,15 @@
...
+	private final boolean _startsWithDelim;
+	private final boolean _endsWithDelim;
 	private String _pattern;
 	private char _wildcard;
-	private boolean _endsWithDelim;
 
 	public WildcardPattern(String pattern, char wildcard) {
 		_pattern = pattern;
 		_wildcard = wildcard;
+		_startsWithDelim = (_pattern.charAt(0) == _wildcard);
```
kaspersorensen: This will throw IndexOutOfBoundsException if an empty string is passed as the pattern.

kaspersorensen:
I agree that this needs to be fixed and it seems like a good fix. Just a remark that we should also support empty string operands. Probably worth a unittest too :)

LosD:
Whoopsie, yes

LosD:
Empty strings now supported.

We're still pretty far from full LIKE support as we're missing single character (_), character lists ([abc] much like regex), negation ([!abc]) and ESCAPE. However those are simply missing features and not a  bug like this. I'll create a lower-priority issue with that.

LosD:
Ah, it seems only % and _ are standard SQL, according to [SQL-99 Complete, Really](https://mariadb.com/kb/en/sql-99/character-string-operations/#like-predicate). That probably makes the job much simpler.

kaspersorensen:
LGTM. And I see you've created an issue for the underscore thing, so let's merge this. Does _this_ PR have a issue though?

kaspersorensen:
Ah, METAMODEL-1103 ... I'll merge then.

PR #117, opened 2016-07-26
METAMODEL-1106: Added 'INDEX' (and other) keyword(s) to SqlKeywords

kaspersorensen:
Proposed fix for METAMODEL-1106


kaspersorensen:
Merging as per lazy concensus

PR #118, opened 2016-07-27
METAMODEL-1099 elasticsearch factories

kaspersorensen:
This PR adds some factories for the two ElasticSearch modules.


kaspersorensen:
Merging as per lazy concensus.

PR #111, opened 2016-06-14
RESTful API swagger docs

kaspersorensen:
This is a continuation of #110 which adds a Swagger definition file for the REST endpoints. I'd like a review of the swagger.yaml file contents please.

(For ease-of-reading I would recommend pasting it in to e.g. http://editor.swagger.io/)

PR #119, opened 2016-08-05
RESTful MetaModel

kaspersorensen:
Again an attempt to post a PR for a completely new webapp that offers MetaModel as a RESTful API.

Previously posted as #110 and #111


kaspersorensen:
Argh ... The old remarks are still being added. I will make a diff and make a new single-commit branch.

PR #120, opened 2016-08-05
RESTful MetaModel service-webapp

kaspersorensen:
Squashed all the commits of #110 and #111 to get the long review history cleared.


kaspersorensen:
I've made a couple of changes to this PR today and to make it easier for people to review the webapp not just from a code perspective, but also from a user perspective, I've published my local build as a docker image on Docker Hub:

https://hub.docker.com/r/kaspersor/metamodel/

Feedback very much appreciated, and especially some kind of idea of whether you think we should add this application to MetaModel in the first place?

kaspersorensen:
Bump :) I would like to see some progress on the review of this rather big feature. If anyone would give it a review it would be awesome. Or even a test using the before-mentioned docker image.

LosD:
I think the reason that you're missing reviews is exactly because it is so big. I've started a few times, and every time I've lost interest rather quickly.

It seems a bit like shooting birds with cannons. The tenant stuff is just too much.

I'd go for something simple, like literally just a REST API to a DataStore in a Docker image. Everything else seems way out of scope to me.

kaspersorensen:
Bummer ;-) But thanks for the honest feedback.
Functionally speaking I think it is necesary to have at least a collection of datastores. You may be right that the tenant concept is too much, but at least having the ability to create more than one datastore (in a single "collection"/"registry") is valuable so that you don't need to start new server processes just because you have new datastores. It will also enable composite datastores which wouldn't be possible I guess on a single-datastore instance).

LosD:
Yeah, that might make sense, though I'd only see an advantage in the non-Docker case (which is of course an entirely valid case... Easy to get Docker-tunnel-vision these days 😆). But in many cases I think there's often advantages to keeping datastores separate. For the Docker case, there is likely performance improvements to using Kubernetes or Swarm clustering.

It would be super nice if the API could somehow make multiple datastores possible (and I guess de/registering datastores, though I'd love to cut that lard as well and just register it/them on startup or image create/run in the Docker case), and still making single datastore access as simple as possible.

kaspersorensen:
The way I see it is that Docker apps or microservices should try to do just a few things, and do them great. But the question is "what is the thing" then, because I think your interpretation is a bit different than mine there. In my opinion the "thing it does" is present a unified/federated view of your databases. The individual databases will exist elsewhere (maybe as other docker apps, but that's besides the point) but the thing that MM brings to the table is a way to discover them all and use them via a single interface that applies to them all. A common denominator (with all the bad stuff that that means), and a unification layer too. If that is the "thing it does" then it is really still a microservice approach to the webapp to have the collection since if it didn't have a collection there would only be a common denominator of one, and that wouldn't improve anything really?

LosD:
Hmmmm, good point.

The thing is that if it is in a cluster, it's hard to get the docker image on the same instance (or even rack) as all the DBs it provides access to when there's many of them, and that means that for chatty apps the indirection of MM can become quite expensive.

OTOH, chatty apps are often a symptom of bad app or DB design (though there is of course cases where it is unavoidable).

kaspersorensen:
OK so I have been going over this in my mind again and again a few times. Couldn't quite make myself remove the tenant concept either :-P At least I feel it needs a bit more thought. One of the things I like about it in a URL pattern like `/{tenant}/{datastore}`is that it allows for simple authorization schemes to grant access to `/{some-tenant}/*` and thereby make someone a tenant administator. Similarly of course a particular datastore could be authorized by using the URL pattern. Not having any such "outer collection name" means that setting up access rules becomes a lot more difficult. Whether it is tenant or some other grouping term I don't mind so much. It just feels like a big aid in organizing the catalog of datastores.

kaspersorensen:
Closing this PR as we now have a new repo for MetaModel Membrane.

PR #121, opened 2016-08-05
FixedWidthReader diacritics bug fix

jhorcicka:
Refactoring of FixedWidthReader in PR https://github.com/apache/metamodel/pull/103 introduced a bug related to charset (which was not used). So, input containing diacritic characters was not read correctly. 

JUnit FixedWidthReaderTest was updated with charset tests.


fixedwidth/src/main/java/org/apache/metamodel/fixedwidth/FixedWidthReader.java
```diff
@@ -85,6 +92,15 @@ public FixedWidthReader(InputStream stream, String charsetName, int[] valueWidth
         _expectedLineLength = expectedLineLength;
     }
 
+    private void initReader() {
+        try {
+            InputStreamReader inputStreamReader = new InputStreamReader(_stream, _charsetName);
```
jhorcicka: This was the missing part in the previous version, _charsetName was not used anywhere.

LosD:
This seems to be based on your old branch, instead of on mm:master.

It might be a good idea to make a new PR based directly on mm:master so there's no guesswork on which commits are new.

jhorcicka:
I closed this PR because of a bad source, I will create a new one.

PR #122, opened 2016-08-05
METAMODEL-1109: FixedWidthReader diacritics bug fix 

jhorcicka:
Refactoring of FixedWidthReader in PR #103 introduced a bug related to charset (which was not used). So, input containing diacritic characters was not read correctly.

JUnit FixedWidthReaderTest was updated with charset tests.

### DIFF
--- a/fixedwidth/src/main/java/org/apache/metamodel/fixedwidth/EbcdicReader.java
+++ b/fixedwidth/src/main/java/org/apache/metamodel/fixedwidth/EbcdicReader.java
@@ -26,6 +26,8 @@ import java.io.IOException;
  */
 class EbcdicReader extends FixedWidthReader {
 
+    private final BufferedInputStream _stream;
+    private final String _charsetName;
     private final boolean _skipEbcdicHeader;
     private final boolean _eolPresent;
     private boolean _headerSkipped;
@@ -33,6 +35,8 @@ class EbcdicReader extends FixedWidthReader {
     public EbcdicReader(BufferedInputStream stream, String charsetName, int[] valueWidths,
             boolean failOnInconsistentLineWidth, boolean skipEbcdicHeader, boolean eolPresent) {
         super(stream, charsetName, valueWidths, failOnInconsistentLineWidth);
+        _stream = stream;
+        _charsetName = charsetName;
         _skipEbcdicHeader = skipEbcdicHeader;
         _eolPresent = eolPresent;
     }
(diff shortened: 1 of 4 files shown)


fixedwidth/src/main/java/org/apache/metamodel/fixedwidth/FixedWidthReader.java
```diff
@@ -85,6 +92,15 @@ public FixedWidthReader(InputStream stream, String charsetName, int[] valueWidth
         _expectedLineLength = expectedLineLength;
     }
 
+    private void initReader() {
+        try {
+            InputStreamReader inputStreamReader = new InputStreamReader(_stream, _charsetName);
```
jhorcicka: This was the missing part in the previous version, _charsetName was not used anywhere.

fixedwidth/src/main/java/org/apache/metamodel/fixedwidth/FixedWidthReader.java
```diff
@@ -43,6 +47,7 @@
     private final boolean _constantWidth;
     private volatile int _rowNumber;
     protected final BufferedInputStream _stream;
+    protected Reader _reader;
```
kaspersorensen: I would suggest to keep having final fields only, and since `_stream` and `_charset` are no longer really used, I'd suggest to remove those and instead make the `initReader` return the created Reader with the needed arguments, like this:

```
_reader = initReader(stream, charsetName);
```

jhorcicka: Done.

fixedwidth/src/main/java/org/apache/metamodel/fixedwidth/EbcdicReader.java
```diff
@@ -26,13 +26,17 @@
...
+    private final String _charsetName;
     private final boolean _skipEbcdicHeader;
     private final boolean _eolPresent;
     private boolean _headerSkipped;
     
     public EbcdicReader(BufferedInputStream stream, String charsetName, int[] valueWidths,
             boolean failOnInconsistentLineWidth, boolean skipEbcdicHeader, boolean eolPresent) {
         super(stream, charsetName, valueWidths, failOnInconsistentLineWidth);
+        _stream = stream;
+        _charsetName = charsetName;
```
kaspersorensen: These two don't seem to ever be used?

jhorcicka: But they are, lines: 61, 70, 76.

kaspersorensen: Ah yes, it was fooling me because it used to be the super-class' fields and now they've moved here... Makes sense.

PR #123, opened 2016-08-08
METAMODEL-1111 Empty string treatment in Oracle

jhorcicka:
Fixes Oracle '' = null problem. In OracleQueryRewriter empty string operand is reaplaced by null.


kaspersorensen:
I wonder if this rule applies to all String-based data types in Oracle. Do we have any documentation to back it up? I could find references of this anomaly for `VARCHAR` and `VARCHAR2`, so maybe we need also to add a check on the data type to ensure that we don't rewrite it wrongly for other data types.

jhorcicka:
According to https://docs.oracle.com/cd/B28359_01/server.111/b28318/datatype.htm#i3253 I tried CHAR, NCHAR, CLOB, NCLOB, VARCHAR2 and NVARCHAR2. In these cases when an empty string was assigned to the column, null was stored instead. 

From what I read I think currently there is a general rule '' = null. It might change in the future when Oracle tries to standardize VARCHAR.

kaspersorensen:
OK, let's merge this then.

PR #125, opened 2016-08-16
Use ColumnNamingStrategy argument in CsvConfiguration constructor

arjansh:
Fixes METAMODEL-1113.

Added unit tests to the different types of datastores which can make use of the ColumnNamingStrategies mechanism and fixed the CsvConfiguration constructor which takes a ColumnNamingStrategy as an arguments but doesn't assign it to its columnNamingStrategy field.


kaspersorensen:
LGTM!
(ouch what a lame bug - I feel a bit silly for probably being the one who introduced it)

PR #126, opened 2016-08-18
Make it possible to use column naming strategy for EBCDIC files

arjansh:
Fixes METAMODEL-1114.

Introducing a new EbcdicConfiguration constructor which takes ColumnNamingStrategy as an argument.


kaspersorensen:
The build failure is unrelated and strange. The code LGTM.

PR #127, opened 2016-08-28
Adding 2 more columns END and TABLE

laxmilal-menaria:
Added 2 More columns in Keywords List, I think we have lot of SQL keywords based on this documentation,

https://msdn.microsoft.com/en-us/library/ms189822.aspx

### DIFF
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/SqlKeywords.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/SqlKeywords.java
@@ -58,6 +58,8 @@ class SqlKeywords {
         KEYWORDS.add("DELETE");
         KEYWORDS.add("AND");
         KEYWORDS.add("OR");
+        KEYWORDS.add("END");
+        KEYWORDS.add("TABLE");
     }
 
     public static boolean isKeyword(String str) {


jdbc/src/main/java/org/apache/metamodel/jdbc/SqlKeywords.java
```diff
@@ -58,6 +58,8 @@
         KEYWORDS.add("DELETE");
         KEYWORDS.add("AND");
         KEYWORDS.add("OR");
+        KEYWORDS.add("END");
```
kaspersorensen: If we add "END" then we should probably also add "BEGIN" :)

laxmilal-menaria: Yes, we can add this too, but what happens if we add all keywords in one shot, is any performance degrade?
Adding columns one by one is not a good approach.
Thoughts?

Thanks,
Laxmilal Menaria


Sent from my iPhone




On Mon, Aug 29, 2016 at 10:07 PM +0530, "Kasper Sørensen" <notifications@github.com> wrote:












In jdbc/src/main/java/org/apache/metamodel/jdbc/SqlKeywords.java:
> @@ -58,6 +58,8 @@
>          KEYWORDS.add("DELETE");
>          KEYWORDS.add("AND");
>          KEYWORDS.add("OR");
> +        KEYWORDS.add("END");



If we add "END" then we should probably also add "BEGIN" :)

kaspersorensen: I don't worry at all about the performance here because it is still a pretty small set and it happens just once. I worry a bit about maintaining this keyword list though, but that's a conversation that maybe goes beyond the scope of this little fix. I will merge it in.

PR #128, opened 2016-09-06
Demonstrated issue with AVG function and fixed/adapted in JDBC module

kaspersorensen:
I was looking at METAMODEL-193 because I found it a bit strange that it was reported that e.g. `AVG(an-integer-column)` would always return an integer. After some digging around I realized that this is common RDBMS behaviour. But common or not, it is inconsistent with all the other MetaModel connectors.

So I thought to fix it so that there is consistency: I made it so that the `AVG` function will return a double no matter what. This requires the native SQL to be rewritten slightly to make the RDBMS pick it up. So my fix was to rewrite it into `AVG(1.0*an-integer-column)`. Granted this is a pretty dirty rewrite. So I am curious for review remarks and open for improvement ideas.

Note that this PR does NOT fix METAMODEL-193, but it fixes one of the use-cases causing that issue to be relevant at all.

### DIFF
--- a/core/src/main/java/org/apache/metamodel/query/AverageAggregateFunction.java
+++ b/core/src/main/java/org/apache/metamodel/query/AverageAggregateFunction.java
@@ -18,6 +18,7 @@
  */
 package org.apache.metamodel.query;
 
+import org.apache.metamodel.schema.ColumnType;
 import org.apache.metamodel.util.AggregateBuilder;
 
 public class AverageAggregateFunction extends DefaultAggregateFunction<Double> {
@@ -34,4 +35,8 @@ public class AverageAggregateFunction extends DefaultAggregateFunction<Double> {
         return new AverageAggregateBuilder();
     }
 
+    @Override
+    public ColumnType getExpectedColumnType(ColumnType type) {
+        return ColumnType.DOUBLE;
+    }
 }
--- a/core/src/test/java/org/apache/metamodel/query/AverageAggregateFunctionTest.java
+++ b/core/src/test/java/org/apache/metamodel/query/AverageAggregateFunctionTest.java
@@ -18,28 +18,20 @@
  */
 package org.apache.metamodel.query;
 
-import org.apache.metamodel.schema.ColumnType;
-import org.apache.metamodel.util.AggregateBuilder;
+import static org.junit.Assert.assertEquals;
 
-public class SumAggregateFunction extends DefaultAggregateFunction<Double> {
-    
-    private static final long serialVersionUID = 1L;
+import org.apache.metamodel.MockDataContext;
+import org.apache.metamodel.data.DataSet;
+import org.junit.Test;
 
-    @Override
-    public String getFunctionName() {
-        return "SUM";
-    }
+public class AverageAggregateFunctionTest {
 
-    @Override
-    public AggregateBuilder<Double> createAggregateBuilder() {
-        return new SumAggregateBuilder();
-    }
-
-    @Override
-    public ColumnType getExpectedColumnType(ColumnType type) {
-        if (type == ColumnType.BIT || type == ColumnType.INTEGER || type == ColumnType.BIGINT) {
-            return ColumnType.BIGINT;
+    @Test
+    public void testEvaluateAvgOfIntegersWithQueryPostprocessor() throws Exception {
+        final MockDataContext dc = new MockDataContext("sch", "tbl", "foo");
+        try (final DataSet ds = dc.query().from("tbl").select("AVG(foo)", "SUM(foo)", "COUNT(*)").execute()) {
+            ds.next();
+            assertEquals("Row[values=[2.5, 10.0, 4]]", ds.getRow().toString());
         }
-        return ColumnType.DOUBLE;
     }
 }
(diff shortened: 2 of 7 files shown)


LosD:
AVG is supposed to return the same output type as input type (or, at least the same class, e.g. an integer type with tinyint, int arguments and floating point type with float or double arguments), not always an input type. That also seems like the only sane thing to do.

LosD:
If we do it differently in other connectors, we really need to fix them, not break SQL.

kaspersorensen:
Do you mean "not always a double type" (instead of integer)? Or else I don't understand what you're saying?

Are you simply saying that "this is the rule of SQL, so we should follow that", or are you saying that you think the rule is good? I would question the latter, but OK I might be wrong in asserting this issue in the first place. To me it just seems a whole lot more natural and correct to have an average-function return a decimal type of some sort.

LosD:
Yeah I wasn't exactly clear, sorry about that. The SQL rule is that output matches input (or at least a type that is close) and I think that's the only sane thing to do when there is no way to specify what output type is wanted.

kaspersorensen:
Hmm it seems to me that it is just "a rule" but not always "the right rule". For instance, MongoDB seems to return decimals when needed [1] which I would personally also expect from SQL, had I not recently learned that SQL didn't.

It's sorta like asking: If you where to write a `Math.avg(int[] numbers)` function, what would it return? And I would instinctively always say ´double´ to that. Now there may be alternative view or special needs, requiring maybe a special variant for higher and lower precision.

[1] https://docs.mongodb.com/manual/reference/operator/aggregation/avg/

LosD:
While not a super example, since there's not really a general math library with an average, but most of them go pretty far not to change type on you, and I would certainly expect the same with an average. Especially since you can't predict the same result of a float average casted to an int, as of an int average. Even worse if you have decimals and want to make an average of them since you wouldn't use decimals if you didn't care about precision and predictably in the first place.

LosD:
Here's a quick test of the difference between float and long averages:
https://gist.github.com/LosD/d6d3bd6b2615375d840dcded45e9b898

The bigger the difference between sum and count, the bigger the chance. Using the values above (random number of any possible long / random positive integer) I get ~0.00028% failures. Limiting count to 0 < count < 100000, I get ~2.26% failures. Limiting the sum to only int values, I get 0% failures.

kaspersorensen:
Hmm, so it appears to me that there are pros and cons of either way of thinking about it. I am trying to juggle a few concerns: (1) consistency within MetaModel and (2) meeting users expectations (which we have demonstrated can differ).

Maybe the best would even be to have two FunctionTypes for it then? Or maybe even three, a la , `AVG_SAME_TYPE` (will retain int->int), `AVG_DECIMAL` (will always return a decimal type), `AVG_NATIVE` (will do whatever the native backing store prefers). At the same time that doesn't really sound very appealing somehow :)

LosD:
This has been laying around for a while. I guess some kind of decision has to be made on it sooner or later.

If we need options, maybe we should have per-context flags? I see other places where that would make sense (e.g. in #165). Of course, we should still settle on sane defaults, so it doesn't really end the discussion, but apparently there's different needs and expectations. To me, the default should always be keeping the correct type. When we're getting the tools we need with METAMODEL-192/METAMODEL-193, we should try to uphold type consistency. It's then easy to cast to a floating point type if you want it to be rounded, while it's impossible to undo the damage from an automatic conversion.

kaspersorensen:
I updated your gist. I think there is still a good case for using decimal avg's ... Here's my output, indicating that your test is more often wrong than right when applying more decimal precision and real rounding instead of int-division (which I believe always rounds down):

```
false alarms: 2693
errors: 149, total tests = 1000000000, percentage: 0.000015%
```

kaspersorensen:
Yes, I should have of course tested that to explain what I mean. But a simple
```
System.out.println("An avg: " + (199 / 100));
```
outputs
```
An avg: 1
```

LosD:
Part of the point is it _should_ always round down when you're using integers. You're changing the type _and_ the behaviour.

Anyway, it still doesn't fix the base problem with unwanted type changes, and a BigDecimal is _at least_ 32-_byte_, besides being an even weirder type to get back in many cases.

kaspersorensen:
I'm not saying that I want to return a BigDecimal. Rather I'm saying that the test isn't correct in my opinion. Rounding a double down to the nearest integer is doable when wanted. But the other direction isn't possible. So in my opinion the more natural data type to return is a double (or some other decimal type). I guess with my example of `199/100 == 1`, I'm simply arguing that it's silly to return the same data type. Which is why I guess e.g. MongoDB also returns a double. I haven't researched other databases yet, but I suspect that this aspect of SQL is something that isn't generally appreciated or consistently implemented because it seems so odd.

kaspersorensen:
I've been looking a bit more, and it seems that ANSI SQL does not specify anything here. PostgreSQL and MS SQL use the behaviour you describe, whereas Oracle and MySQL return floating points. Furthermore I checked ElasticSearch, which also returns floating points.

LosD:
Of course the other direction is possible. Just cast to a double first. Changing automatically is a destructive operation, which there will be no way to avoid.

ANSI SQL says "don't lose precision", which is only possible if you aren't using long (Since double only has a 52-bit mantissa). Then you could make long a special case. And decimal is a special case as well. Making 32-bit and lower ints the only data types that does not return its own datatype. And then we're really in inconsistent-land.

kaspersorensen:
I'm closing this PR since I guess the discussion has gone dead. Not sure what to do about the precision issue here, but I think a discussion/vote would be more appropriate and then we may revive (or not) this PR.

PR #129, opened 2016-09-07
Fix DefaultCompiledQuery.copyFilterItem

aportosa:
Previous version was missing the use of logical operator while calling FilterItem constructor


kaspersorensen:
Ouch, yes. LGTM. Would be good to have a small unittest (and that would be easy to make here) though.

LosD:
Travis error seems unrelated. Seems like a Travis issue (Some Ruby thing gone wrong).

kaspersorensen:
Yes, let's merge this. If anyone also feels like making a tiny unittest just to confirm the copyFilterItem behaviour I would applaude it :)

PR #130, opened 2016-09-19
Added primary key lookup optimization for Apache Cassandra module

kaspersorensen:
I think the PR title says it all ... I was working a bit with Cassandra and found that it was easy (and really missing) in our connector to add optimized support for primary key lookup queries.

### DIFF
--- a/cassandra/src/test/java/org/apache/metamodel/cassandra/CassandraDataContextTest.java
+++ b/cassandra/src/test/java/org/apache/metamodel/cassandra/CassandraDataContextTest.java
@@ -32,6 +32,7 @@ import javax.swing.table.TableModel;
 import org.apache.metamodel.data.DataSet;
 import org.apache.metamodel.data.DataSetTableModel;
 import org.apache.metamodel.data.FilteredDataSet;
+import org.apache.metamodel.data.InMemoryDataSet;
 import org.apache.metamodel.query.Query;
 import org.apache.metamodel.query.parser.QueryParserException;
 import org.apache.metamodel.schema.ColumnType;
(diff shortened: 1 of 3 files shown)


cassandra/src/test/java/org/apache/metamodel/cassandra/CassandraDataContextTest.java
```diff
@@ -113,7 +114,7 @@ public void testSchemaAndSimpleQuery() throws Exception {
 
     @Test
     public void testWhereColumnEqualsValues() throws Exception {
-        DataSet ds = dc.query().from(testTableName).select("id").and("title").where("id").isEquals(firstRowId)
+        DataSet ds = dc.query().from(testTableName).select("id").and("title").where("title").isEquals(firstRowTitle)
```
LosD: Was the original test broken by the change?

LosD:
Code LGTM, but I know nothing about Cassandra.

A small worry is the change of a seemingly unrelated test method.

cassandra/src/test/java/org/apache/metamodel/cassandra/CassandraDataContextTest.java
kaspersorensen: No but the query no longer represented a "filtered dataset" type query because it got optimized. In fact the new test method is identical to the old test method, except that the assertion of it being a `InMemoryDataSet` is there to show the different paths that the two queries take depending on what column is being filtered on.

LosD: Ahhh yes. Makes sense.

LosD:
Everything LGTM now, but it would be nice with a review from someone that knows Cassandra.

kaspersorensen:
Maybe @albertostratio ? :)

kaspersorensen:
OK I am sorry that no one is adding their 10 cents, but given the time it has been for review, I am going to merge (I did test it myself on actual Cassandra databases).

PR #131, opened 2016-09-20
Improved the query builder API by overloading groupBy and having methods

kaspersorensen:
A few minor improvements to the query builder API. I ran into these today while building a bit more complex queries than usual and I was missing the method support for it in MetaModel.

See syntax examples.


LosD:
Travis error seems unrelated (apparently a crashing ElasticSearch... We've also seen quite a few problems like this in DataCleaner lately, with services or the Travis image itself crashing/hanging... Maybe they're having trouble at the moment?)

LosD:
LGTM (besides the full-file reformats, but that probably doesn't come as a surprise :))

PR #132, opened 2016-10-05
MSSQL datetime format language neutral

Juslwk:
Currently, creating a date time condition filter, the format is hardcoded yyyy-MM-dd
yyyy-MM-dd for US language setting
yyyy-dd-MM for British settings
yyyy-MMM-dd or yyyyMMdd for universal settings
we would need to change the hardcoded part to yyyyMMdd or yyy-MMM-dd so that it would work on any language settings.

### DIFF
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/SQLServerQueryRewriter.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/SQLServerQueryRewriter.java
@@ -112,7 +112,7 @@ public class SQLServerQueryRewriter extends DefaultQueryRewriter {
 
                 final Date date = (Date) operand;
 
-                final DateFormat format = DateUtils.createDateFormat("yyyy-MM-dd HH:mm:ss");
+                final DateFormat format = DateUtils.createDateFormat("yyyyMMdd HH:mm:ss");
                 final String dateTimeValue = "CAST('" + format.format(date) + "' AS DATETIME)";
 
                 sb.append(dateTimeValue);
--- a/jdbc/src/test/java/org/apache/metamodel/dialects/SQLServerQueryRewriterTest.java
+++ b/jdbc/src/test/java/org/apache/metamodel/dialects/SQLServerQueryRewriterTest.java
@@ -87,7 +87,7 @@ public class SQLServerQueryRewriterTest extends TestCase {
                         .toDate("2014-06-28 14:06:00")));
 
         assertEquals(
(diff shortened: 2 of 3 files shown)


kaspersorensen:
The code looks good to me, but I did not verify on any actual DB. Anyone with a MS SQL db that could integration test this?

Juslwk:
For the salesforce change, should I have made a new pull request instead of merging with this? so sorry my mistake... I'm not so familiar with github... 
Do I revert and then make a new branch on my github and then create a new pull request??

kaspersorensen:
Ideally you would create another PR, but given the tiny nature of the change, I think this is probably OK.

But I do have a question on it: Have you tested it? Is currency values in deed returned as double values? Do they properly aggregate in a MAX statement or similar? The code of it seems fine, but I would judge it based on integration test results.

Juslwk:
Hi, yes currency is returned as double. This is the documentation from salesforce:
[Object Reference for Salesforce and Force.com - Currency](https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/field_types.htm#i1435541)
I've tested all the aggregation on our own program, connecting to Salesforce via metamodel, and looks good. I didn't do your integration test though.. I'm not sure why but my tests always fails at cassandra, saying it cant start up the server within the specified time, so most of the time I just skip tests and build, then test on our own program.

LosD:
I believe the integration tests @kaspersorensen is talking about are the one in the JDBC module which are configured by copying the `example-metamodel-integrationtest-configuration.properties` file to `~/metamodel-integrationtest-configuration.properties`, then uncommenting and configuring the `jdbc.sqlserver.` properties. Please note that you will need the AdventureWorks database.

However, I wouldn't bother, unless you also feel like fixing the tests. There's failures relating to tables not being in the "correct" order in the schema, times being compared using exact string matches (SQL Server timestamps can change by +/- 3ms), and other failures<sup>1</sup>.

The good news is that, testWorkingWithDates was happy when I tried, so I guess that means that your change didn't break anything.

It's weird that Cassandra fails for you, by the way. My only issue is that the test dumps a `.toDelete` file in the Cassandra folder that I need to nuke for the license check to stay happy.

<sup>1</sup> One of them might be a real issue, though. `new Query().select("Name")` fails with "Not capable of parsing SELECT token: Name") on line 185. Doesn't seem related to your change, though.

Juslwk:
@LosD Thanks for testing the testWorkingWithDates for us, I'm glad I didnt break anything. 
I don't know why Cassandra fails for me too. It fails on a clean fork as well! It also creates a .toDelete here.

kaspersorensen:
OK cool, I will merge.

PR #133, opened 2016-10-10
Don't get killed by Hive SQLExceptions when finding default schema name

LosD:
This will allow Hive to throw its exceptions while we're trying to determine the default schema name, without that failing.

Also fixes a problem where columns types without a width would fail by specifying maximum allowed column width for VARCHAR types. (Maybe I should have fixed this in another PR, but it broke the Hive integration test, which drove me mad... Let me know if I should split them).

Fixes METAMODEL-1119

### DIFF
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/HiveQueryRewriter.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/HiveQueryRewriter.java
@@ -35,6 +35,10 @@ public class HiveQueryRewriter extends DefaultQueryRewriter {
         if (columnType == ColumnType.INTEGER) {
             return "INT";
         }
+        // Hive does not support VARCHAR without a width, nor VARCHAR(MAX).
+        if (columnType == ColumnType.STRING && columnSize == null) {
+            return super.rewriteColumnType(columnType, 65535);
+        }
         return super.rewriteColumnType(columnType, columnSize);
     }
     
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/HiveQueryRewriter.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/HiveQueryRewriter.java
@@ -35,8 +35,14 @@ public class HiveQueryRewriter extends DefaultQueryRewriter {
         if (columnType == ColumnType.INTEGER) {
             return "INT";
         }
+
+        if(columnType == ColumnType.STRING) {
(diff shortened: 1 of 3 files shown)


jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/HiveQueryRewriter.java
```diff
@@ -35,6 +35,10 @@ public String rewriteColumnType(ColumnType columnType, Integer columnSize) {
         if (columnType == ColumnType.INTEGER) {
             return "INT";
         }
+        // Hive does not support VARCHAR without a width, nor VARCHAR(MAX).
+        if (columnType == ColumnType.STRING && columnSize == null) {
```
kaspersorensen: Maybe checking with STRING isn't quite enough here. I could imagine a lot of Hive users using the more specific VARCHAR column type.

LosD: Ah yes. Will fix. I'll also look for other VARCHAR "convertees" (I'm not actually sure if other non-VARCHAR types are in danger of the same issue... I'll see if I can find anything).

LosD: Hmmm... It actually seems that Hive has a literal STRING type without length, that some actually recommends.

We've seen other cases of problems with STRING before (in SQL Server), but I can't remember the details. Do you remember if it is something to worry about, @kaspersorensen?

kaspersorensen: No I wouldn't worry about it. There was a bug once about not rewriting STRING properly but I think that's just history.

LosD:
Okay, that should fix it.

BTW, another curious thing with Hive on JDBC: Hive inserts are _very_ slow, since it's starting a MapReduce job for every insert... So you'd expect them to prioritize the batch feature? Nope. No support. Every... Single... Insert... Becomes... It's... Own... Job... I think the integration test job uses around 10-15 seconds per insert on my laptop (granted, with Hive running in the MapR sandbox on VirtualBox, but the VM overhead shouldn't really be noticable).

I have a feeling that JDBC isn't really prioritized, so we might at some point see if there's another more low-level API to connect to Hive.

On the other hand, querying is fast, so in that regard it's fine.

kaspersorensen:
LGTM, will merge.

PR #134, opened 2016-10-25
Fixed toString() method of insert-into-table builder

kaspersorensen:
I came across a little issue today while debugging some code that used MetaModel. The toString() method of an INSERT builder had no comma-delimitation of the values in the INSERT statement. Obviously this is only used for toString() representation and not passed on to the database, but still it looks bad and it shouldn't ;)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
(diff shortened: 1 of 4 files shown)


kaspersorensen:
I'll merge based on lazy consensus.

LosD:
Weird. I added a longer comment that essentially said "code looks good, but this is horrible as it's trying to fix 3 things without even referencing a single issue... And updating RAT is not fixing the actual issue", but apparently something went wrong.

kaspersorensen:
Sorry about that. Would you like to undo any of it?

LosD:
Nah, too late. Just wanted to voice my concern over PRs that are not bound to (single) issues becoming a habit :slightly_smiling_face:

kaspersorensen:
Noted, and agree!

PR #135, opened 2016-10-28
METAMODEL-1127: Fixed setting of null Map on postgres

kaspersorensen:
Fixes https://issues.apache.org/jira/browse/METAMODEL-1127


LosD:
LGTM!

PR #136, opened 2016-11-10
METAMODEL-1128: Size set and enabled scrolling

kaspersorensen:
Here's my stab at a simple fix for METAMODEL-1128.


LosD:
Code LGTM, but haven't tested.

Should we make a story specifically about looking into using sliced scrolls for performance? I'm not sure how much it'll take, or how big the benefit would be.

LosD:
Tested using DataCleaner. Seems to work fine, and reasonably performant; ~6 seconds to do value distribution on 100K records (value distribution run on two columns)... The test set I used is available here: http://ikeptwalking.com/elasticsearch-sample-data/

One thing I didn't test in depth (since I don't know the dataset) was the correctness of the data retrieved. However, _id keys were unique, so I'm pretty that there's at least no problems with getting same data twice.

LosD:
Should we merge, or wait for reporter of METAMODEL-1128 to give feedback?

Juslwk:
LGTM! I've tested a bit on my own data set, works great!

kaspersorensen:
Great, let's merge it then.

PR #137, opened 2016-12-09
Metamodel 1132 sql native paging

rposkocil:
fixes https://issues.apache.org/jira/browse/METAMODEL-1132

### DIFF
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/AbstractQueryRewriter.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/AbstractQueryRewriter.java
@@ -464,15 +464,16 @@ public abstract class AbstractQueryRewriter implements IQueryRewriter {
     protected boolean isSupportedVersion(String databaseProductName, int databaseVersion) {
 
         if(databaseProductName.equals(_dataContext.getDatabaseProductName())
-                && databaseVersion <= databaseVersionToInt(_dataContext.getDatabaseVersion())) {
+                && databaseVersion <= getDatabaseMajorVersion(_dataContext.getDatabaseVersion())) {
             return true;
         }
         return false;
     }
 
-    private int databaseVersionToInt(String version) {
+    private int getDatabaseMajorVersion(String version) {
         int firstDot = -1;
         if(version != null) {
+            version = version.replaceAll("[^0-9.]+", "");
             firstDot = version.indexOf('.');
         }
         if(firstDot >= 0) {
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/AbstractQueryRewriter.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/AbstractQueryRewriter.java
@@ -460,4 +460,25 @@ public abstract class AbstractQueryRewriter implements IQueryRewriter {
         }
         return resultSet.getObject(columnIndex);
     }
+
+    protected boolean isSupportedDatabase(String databaseProductName, int databaseVersion) {
+
+        if(databaseProductName.equals(_dataContext.getDatabaseProductName())
+                && databaseVersion <= databaseVersionToInt(_dataContext.getDatabaseVersion())) {
+            return true;
+        }
+        return false;
+    }
+
+    private int databaseVersionToInt(String version) {
+        int firstDot = -1;
+        if(version != null) {
+            firstDot = version.indexOf('.');
+        }
+        if(firstDot >= 0) {
+            return Integer.valueOf(version.substring(0, firstDot));
+        } else {
+            return 0;
+        }
+    }
 }
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/AbstractQueryRewriter.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/AbstractQueryRewriter.java
@@ -461,7 +461,7 @@ public abstract class AbstractQueryRewriter implements IQueryRewriter {
         return resultSet.getObject(columnIndex);
     }
 
-    protected boolean isSupportedDatabase(String databaseProductName, int databaseVersion) {
+    protected boolean isSupportedVersion(String databaseProductName, int databaseVersion) {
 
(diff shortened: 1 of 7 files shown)


jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/OracleQueryRewriter.java
```diff
@@ -18,17 +18,19 @@
...
 
 /**
  * Query rewriter for Oracle
  */
-public class OracleQueryRewriter extends DefaultQueryRewriter {
+public class OracleQueryRewriter extends OffsetFetchQueryRewriter {
 
     public OracleQueryRewriter(JdbcDataContext dataContext) {
-        super(dataContext);
+        super(dataContext, 11);
```
LosD: Probably want to use a constant here to document the magic number, like `FIRST_FETCH_SUPPORTING_VERSION` (or maybe something a little less clumsy than what I came up with :smile:).

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/AbstractQueryRewriter.java
```diff
@@ -460,4 +460,25 @@ public Object getResultSetValue(ResultSet resultSet, int columnIndex, Column col
         }
         return resultSet.getObject(columnIndex);
     }
+
+    protected boolean isSupportedDatabase(String databaseProductName, int databaseVersion) {
```
LosD: Shouldn't this be limited to `isSupportedVersion`? I mean, hopefully the concrete rewriters know what product they are.

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/OracleQueryRewriter.java
LosD: BTW, I thought it was only supported from version 12c in Oracle?

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/SQLServerQueryRewriter.java
```diff
@@ -31,15 +33,10 @@
...
+public class SQLServerQueryRewriter extends OffsetFetchQueryRewriter {
 
     public SQLServerQueryRewriter(JdbcDataContext dataContext) {
-        super(dataContext);
-    }
-
-    @Override
-    public boolean isMaxRowsSupported() {
-        return true;
+        super(dataContext, 11);
```
LosD: Also magic number here

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/SQLServerQueryRewriter.java
```diff
@@ -56,7 +53,7 @@ protected String rewriteSelectClause(Query query, SelectClause selectClause) {
         String result = super.rewriteSelectClause(query, selectClause);
 
         Integer maxRows = query.getMaxRows();
-        if (maxRows != null) {
+        if (maxRows != null && !result.contains("FETCH")) {
             if (query.getSelectClause().isDistinct()) {
```
LosD: Could the same limitation apply to OFFSET/FETCH?

LosD:
BTW, not a blocker for me, but you could consider to use a separate class instead of inheritance for this. It's not really a problem now, but if we want to use other features that are shared between some database engines, this design will break down rather quickly.

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/AbstractQueryRewriter.java
```diff
@@ -460,4 +460,25 @@ public Object getResultSetValue(ResultSet resultSet, int columnIndex, Column col
...
+    protected boolean isSupportedVersion(String databaseProductName, int databaseVersion) {
+
+        if(databaseProductName.equals(_dataContext.getDatabaseProductName())
+                && databaseVersion <= databaseVersionToInt(_dataContext.getDatabaseVersion())) {
+            return true;
+        }
+        return false;
+    }
+
+    private int databaseVersionToInt(String version) {
```
kaspersorensen: I'd call this method something like `getDatabaseMajorVersion()` just because it will return the _major_ version element only, not something that represents the 'real' version.

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/AbstractQueryRewriter.java
```diff
@@ -460,4 +460,25 @@ public Object getResultSetValue(ResultSet resultSet, int columnIndex, Column col
...
+        int firstDot = -1;
+        if(version != null) {
+            firstDot = version.indexOf('.');
+        }
+        if(firstDot >= 0) {
+            return Integer.valueOf(version.substring(0, firstDot));
```
kaspersorensen: This could throw an exception if the version from the DB is something like `R12.0`.

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/AbstractQueryRewriter.java
```diff
@@ -460,4 +460,26 @@ public Object getResultSetValue(ResultSet resultSet, int columnIndex, Column col
...
+    }
+
+    private int getDatabaseMajorVersion(String version) {
+        int firstDot = -1;
+        if(version != null) {
+            version = version.replaceAll("[^0-9.]+", "");
+            firstDot = version.indexOf('.');
+        }
+        if(firstDot >= 0) {
+            return Integer.valueOf(version.substring(0, firstDot));
```
kaspersorensen: Not sure if we want to make this more "defensive" in terms of programming. I can't imagine it really, but if the string argument is just "." then this substring will return empty string and that will throw exception when parsing as integer. It's not really something I think is likely to happen, but just saying :-)

kaspersorensen:
LGTM

JoosjeBoon:
Shall we pull then? I want to test this using the DH connector. Thanks guys!

LosD:
Merged :)

PR #138, opened 2016-12-26
METAMODEL-1133: Made PojoDataContext thread-safe with CopyOnWrite list

kaspersorensen:
My Suggested fix for METAMODEL-1133.


LosD:
Errr... Apparently I am not. Having some permission issues from my home computer to the Apache Git repo.

PR #140, opened 2017-01-24
Dynamo DB support

kaspersorensen:
This PR adds Amazon DynamoDB support to MetaModel. A few notes on the implementation:

 * I'm personally fairly happy with the read/discovery/query part of things. As much as possible is leveraged from Amazon's SDK, but since DynamoDB is essentially schemaless, there's also the usual option of supplying `SimpleTableDef` objects.
 * Creating tables is a bit funky. There are options in creating a DynamoDB table that I haven't seen elsewhere, such as the throughput capacity flags. I added system properties for those. There is also in the SDK options for creating secondary indices, but I didn't find a good way to map that to MetaModel during CREATE TABLE. So I left it out.
 * The testing is pretty sparse right now, and you need a set of Amazon credentials to make what is there work. Not sure if there is a mock service available somehow, or an embedded version of DynamoDB.

JIRA Issue: METAMODEL-1136


dynamodb/src/main/java/org/apache/metamodel/dynamodb/DynamoDbDataContext.java
```diff
@@ -0,0 +1,320 @@
...
+
+    public DynamoDbDataContext(AmazonDynamoDB client) {
+        this(client, null, false);
+    }
+
+    public DynamoDbDataContext(AmazonDynamoDB client, SimpleTableDef[] tableDefs) {
+        this(client, tableDefs, false);
+    }
+
+    private DynamoDbDataContext(AmazonDynamoDB client, SimpleTableDef[] tableDefs, boolean shutdownOnClose) {
```
LosD: Not a biggie in any way, and probably a standard pattern in a lot of MM contexts, but I'm not really a fan of creating the client outside, then closing it inside the class.

I'd much rather use events/callbacks for that sort of thing. Might be the old C++ [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) pattern still making me cringe when I see the responsibility of closing resources passed to other objects. Yes, I feel 300 years old when mentioning stuff like that. :)

dynamodb/src/main/java/org/apache/metamodel/dynamodb/DynamoDbDataContext.java
```diff
@@ -0,0 +1,320 @@
...
+    public void close() {
+        if (_shutdownOnClose) {
+            _dynamoDb.shutdown();
+        }
+    }
+
+    @Override
+    protected Schema getMainSchema() throws MetaModelException {
+        final Map<String, SimpleTableDef> tableDefs = new HashMap<>();
+        for (SimpleTableDef tableDef : _tableDefs) {
```
LosD: I'll only mention it here, but all your enhanced for loops could have `final`s.

(as well as all or most of your method signatures, though I'm not sure I can convince you to do that :))

dynamodb/src/main/java/org/apache/metamodel/dynamodb/DynamoDbDataContext.java
```diff
@@ -0,0 +1,320 @@
...
+
+            // add primary keys
+            addColumnFromKeySchema("Primary index", tableDescription.getKeySchema(), table, true);
+
+            // add attributes from global and local indices
+            final List<GlobalSecondaryIndexDescription> globalSecondaryIndexes = tableDescription
+                    .getGlobalSecondaryIndexes();
+            if (globalSecondaryIndexes != null) {
+                for (GlobalSecondaryIndexDescription globalSecondaryIndex : globalSecondaryIndexes) {
+                    addColumnFromKeySchema(globalSecondaryIndex.getIndexName(), globalSecondaryIndex.getKeySchema(),
```
LosD: This pattern repeats quite a bit. Would it make sense to extract a method for it?

dynamodb/src/main/java/org/apache/metamodel/dynamodb/DynamoDbDataContext.java
```diff
@@ -0,0 +1,320 @@
...
+            if (table.getColumnByName(attributeName) == null) {
+                final String keyType = keySchemaElement.getKeyType();
+                final MutableColumn column = new MutableColumn(attributeName, table).setPrimaryKey(primaryKey);
+                appendRemarks(column, indexName + " member ('" + keyType + "' type)");
+                table.addColumn(column);
+            }
+        }
+    }
+
+    private void appendRemarks(MutableColumn column, String remarks) {
```
LosD: Could be static, if I'm not blind

dynamodb/src/main/java/org/apache/metamodel/dynamodb/DynamoDbDataContext.java
```diff
@@ -0,0 +1,320 @@
...
+        final String existingRemarks = column.getRemarks();
+        if (existingRemarks == null) {
+            column.setRemarks(remarks);
+        } else {
+            column.setRemarks(existingRemarks + ", " + remarks);
+        }
+    }
+
+    private ColumnType toColumnType(String attributeName, String attributeType) {
```
LosD: This method seems like a prime candidate to go to your `DynamoDbUtils` class

dynamodb/src/test/java/org/apache/metamodel/dynamodb/DynamoDbDataContextIntegrationTest.java
```diff
@@ -0,0 +1,210 @@
...
+        props.load(new FileReader(file));
+
+        final String accessKey = props.getProperty("dynamodb.accessKey");
+        final String secretKey = props.getProperty("dynamodb.secretKey");
+
+        Assume.assumeNotNull(accessKey, secretKey);
+
+        final AWSCredentials credentials = new BasicAWSCredentials(accessKey, secretKey);
+        final AWSCredentialsProvider credentialsProvider = new AWSStaticCredentialsProvider(credentials);
+        client = AmazonDynamoDBClientBuilder.standard().withRegion(Regions.US_WEST_2).withCredentials(
```
LosD: Region should probably be configurable.

dynamodb/src/main/java/org/apache/metamodel/dynamodb/DynamoDbDataContext.java
kaspersorensen: Unfortunately the GlobalSecondaryIndexDescription and LocalSecondaryIndexDescription classes don't share a common super-class, so I'm not sure if it can be made more general.

dynamodb/src/main/java/org/apache/metamodel/dynamodb/DynamoDbDataContext.java
kaspersorensen: I guess I agree ... This is actually there to prevent that we will close the client inside the class. It seems fairly similar to what other MM connectors do though.

kaspersorensen:
Addressed your remarks.

dynamodb/src/main/java/org/apache/metamodel/dynamodb/DynamoDbTableCreationBuilder.java
```diff
@@ -0,0 +1,111 @@
...
+
+        // await the table creation to be "done".
+        {
+            String tableStatus = createTableResult.getTableDescription().getTableStatus();
+            while (TableStatus.CREATING.name().equals(tableStatus)) {
+                logger.debug("Waiting for table status to be ACTIVE. Currently: {}", tableStatus);
+                try {
+                    Thread.sleep(300);
+                } catch (InterruptedException e) {
+                }
```
LosD: Whoops, just saw this: This seems to the wrong way to handle InterruptedException? Someone is trying to stop the thread, so you should exit the operation, not ignore and press on.

kaspersorensen: But it would be someone "interrupting our sleep", not interrupting the thread in general, would it? Basically the table creation on Amazon takes some time - I've seen it taking from 1 second to 10 seconds. But since you can't do anything with the table in that timespan, I wanted a sleeping checker to wait before we return. An interruption of that ... Hmm I can't even see what that would mean :)

LosD: No, it means "stop what you are doing". http://www.nurkiewicz.com/2014/05/interruptedexception-and-interrupting.html?m=1

http://daniel.mitterdorfer.name/articles/2015/handling-interruptedexception/

When you see Java programs stalling when you press cancel (hello, Eclipse)? That's often because someone mistreated the InterruptException.

LosD: The best thing you can do if you can't stop executing, e.g. because of atomicity concerns is something like:
```java
            String tableStatus = createTableResult.getTableDescription().getTableStatus();
            boolean interrupted = false
            while (TableStatus.CREATING.name().equals(tableStatus)) {
                logger.debug("Waiting for table status to be ACTIVE. Currently: {}", tableStatus);
                try {
                    Thread.sleep(300);
                } catch (InterruptedException e) {
                    interrupted = true;
                }
                tableStatus = client.describeTable(tableName).getTable().getTableStatus();
            }
            if (interrupted) {
                 Thread.currentThread().interrupt();
            }
```

(Approach taken from https://www.ibm.com/developerworks/library/j-jtp05236/#N10186)

kaspersorensen: Got it ... So I implemented it a bit like that, except I did it at the UpdateCallback level, allowing for the complete update script to finish (not just the "create table" part of it).

LosD:
Code LGTM now, but for some files it seems like Git suddenly considers them changed (since commit 	e76cd23), even though I can't see any change, not even white space changes. Not entirely sure what's going on.

For some of the text files, line endings has also suddenly changed.

I guess it doesn't matter for mergeability, but it would sure be nice to find out what causes weird stuff like that.

LosD:
Hmmm... Maybe it's all line-ending changes, GitHub just only shows the change for .txt files?

kaspersorensen:
Hmm that's pretty odd. The .gitattributes file should force it to keep the same line endings.

kaspersorensen:
Argh I can't get it to not do it. But if you add `?w=1` to the diff URL then it shows the patch without whitespace changes.
I'd like to just add the PR based on a diff download and then not worry about this since the "merging" is going to happen with a separate/squased commit to master anyway.

LosD:
Yeah let's go ahead. LGTM! 

Unfortunately you'll have to do the merge yourself, or wait until Monday. My home computer and the  Apache Git repo has some weird hate for each other.

PR #142, opened 2017-05-01
Deprecated Ref, Func and Predicate in favor of java.util.function

kaspersorensen:
"Ref" to be replaced by Supplier.
"Func" to be replaced by Function.
"Predicate" to be replaced by Predicate.
"Action" to extend Consumer (allowing for throwing checked exceptions).
"UncheckedFunc to extend Function.

Also added `@SuppressWarning` on method that the Java 8 compiler is
emitting warnings for while the code being alright.

I think I still need a JIRA issue for this, but any feedback on the code front is very welcome.

### DIFF
--- a/core/src/main/java/org/apache/metamodel/CompositeDataContext.java
+++ b/core/src/main/java/org/apache/metamodel/CompositeDataContext.java
@@ -27,6 +27,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Function;
 
 import org.apache.metamodel.data.DataSet;
 import org.apache.metamodel.query.FromItem;
@@ -34,10 +35,9 @@ import org.apache.metamodel.query.Query;
 import org.apache.metamodel.schema.CompositeSchema;
 import org.apache.metamodel.schema.Schema;
 import org.apache.metamodel.schema.Table;
-import org.apache.metamodel.util.Func;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
+//
 /**
  * DataContext for composite datacontexts. Composite DataContexts wrap several
  * other datacontexts and makes cross-datastore querying possible.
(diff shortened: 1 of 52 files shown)


core/src/main/java/org/apache/metamodel/CompositeDataContext.java
```diff
@@ -27,17 +27,17 @@
...
 import org.apache.metamodel.query.FromItem;
 import org.apache.metamodel.query.Query;
 import org.apache.metamodel.schema.CompositeSchema;
 import org.apache.metamodel.schema.Schema;
 import org.apache.metamodel.schema.Table;
-import org.apache.metamodel.util.Func;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
+//
```
tomaszguzialek: I guess this is some leftover...

kaspersorensen: Hmm yes, odd, I don't recall why though, but will remove shortly.

LosD:
Aren't all the `@SuppressWarning("resource")` hiding actual bad code?

I remember looking into our `Closeable` usage before and being horrified.

kaspersorensen:
Oh, I guess they may be eclipse-specific, although I hope not. I did evaluate each case of adding the annotation, and felt that they where all safe cases. But in general I agree that I prefer to fix the cause of the warning. Sometimes the compilers/tools analysis is wrong though; typically when you wrap a closeable in another closeable, you shouldn't close the inner object because that would make the wrapper object unusable (an example of this in MetaModel is around `DataSet` implementations that wrap other `DataSet`s).

kaspersorensen:
Any final verdicts from anyone?

LosD:
Haven't checked line-by-line, but approach LGTM, and happy to see standard classes replacing our own.

kaspersorensen:
I'm gonna take that as a go-ahead. But thinking about it a bit more, I guess it shouldn't be merged to master, but rather the `5.x` branch since this is going to require Java 8.

I'll prepare the branch to be `5.x` compatible.

PR #144, opened 2017-05-10
METAMODEL-1139: Use Java 8 functions instead of our own Ref, Action and Func types

kaspersorensen:
This time the same as #142  but now with an official issue number (https://issues.apache.org/jira/browse/METAMODEL-1139) and suggestion to merge to the `5.x` branch


kaspersorensen:
As #142 was reviewed and accepted, I feel inclined to just merge this as it's so similar. But for good record I can leave it open for the day or so?

PR #145, opened 2017-05-26
METAMODEL-1141-Parse-RFC4180-compliant-CSV

LosD:
This makes it possible to parse RFC4180-compliant CSV, where the escape character is a repeated double quote.

OpenCSVs RFC-4180 parser doesn't enforce other RFC-4180 rules (quote character must be double quote, separator must be comma and EOL must be CRLF), so the check to use the RFC-4180 parser is simply if the escape is the same as the quote character.

This also updates OpenCSV to newest version, and fix a problem in a testfile uncovered by that update.

Fixes METAMODEL-1141


csv/src/test/resources/tickets.csv
```diff
@@ -10,7 +10,7 @@ One way of improving this could be through caching. Another way could be through
 3,DataCleaner 1.5 Release,202,Pattern Finder improvements,DataCleaner-core,None,enhancement,darrenH,assigned,2008-08-19T04:27:12Z+0200,2008-09-16T09:21:56Z+0200,"__Pattern Finder suggestions__:
  * have an option to ignore repeating spaces (so {{{\"aaa aaaaa\" and \"aaa         aaaaa\"}}} are counted as one pattern. 
  * have an option to ignore case, and a different option to preserve case.
- * have an option to treat all 'special' characters as one pattern (so \"aaa*\", \"aaa/\", \"aaa\\" etc' are counted as one pattern, maybe denoted \"aaaS\").
+ * have an option to treat all 'special' characters as one pattern (so \"aaa*\", \"aaa/\", \"aaa\" etc' are counted as one pattern, maybe denoted \"aaaS\").
```
kaspersorensen: Good catch. This was actually exported directly from an old bugtracker called "trac". So maybe it was a bug in trac that caused this issue, but nevertheless I think the fix is good :-)

kaspersorensen: Good catch. This was actually exported directly from an old bugtracker called "trac". So maybe it was a bug in trac that caused this issue, but nevertheless I think the fix is good :-)

kaspersorensen:
Looks good to me!

ClaudiaPHI:
LGTM! Tested and it works fine !

PR #146, opened 2017-06-08
METAMODEL-63: UnionDataSet

kaspersorensen:
Here's my attempt at designing/solving METAMODEL-63.

Ideally we would support UNION also as part of our Query language. But that seems like a huge change and a much more low-hanging fruit would be to allow unions via a special `DataSet` implementation.

I also introduced a new interface, `InvokableQuery`. I've often had the need to pass around both DataContext and Query (because those two are needed in combination to produce the `DataSet`). The new interface allows passing these around in a more coherent and encapsulated way.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,7 @@
 ### Apache MetaModel 5.0
 
  * [METAMODEL-6] - Added update summary containing information about changes on returning UpdateableDataContext.executeUpdate(..)
- * [METAMODEL-63] - Added UnionDataSet, a general purpose utility for doing client-side unions.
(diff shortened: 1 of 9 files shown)


kaspersorensen:
I'd love any feedback before I go ahead and merge this based on lazy concensus?

LosD:
A bit sad to see it implemented without a proper UNION, but still seems like a nice step on the way.

Code LGTM

kaspersorensen:
I agree, full UNION support would be very cool. I would encourage anyone with an interest in that to use this new class as a building block to implement UNION.

PR #149, opened 2017-07-21
Feature/jdbc composite relations

tomatophantastico:
A fix for METAMODEL-1145.

The test might not be in the best location.


kaspersorensen:
The solution looks good to me, but it's not really following our code conventions. Because of the small size, I can fix that up before patching, so no problem. But I see you're taking on a few other PRs too, so maybe it's good for you to pick up our code formatting rules etc. to avoid issues.

PR #150, opened 2017-07-21
Feature/quote columns

tomatophantastico:
Working on a problem similar to  METAMODEL-1143 ,this is my solution for HSQLDB and select queries.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -32,6 +32,10 @@ under the License.
 			<artifactId>slf4j-api</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>com.google.guava</groupId>
+			<artifactId>guava</artifactId>
+		</dependency>
+		<dependency>
 			<groupId>org.slf4j</groupId>
 			<artifactId>slf4j-nop</artifactId>
 			<scope>test</scope>
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/DefaultQueryRewriter.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/DefaultQueryRewriter.java
@@ -101,7 +101,7 @@ public class DefaultQueryRewriter extends AbstractQueryRewriter {
         return super.rewriteColumnType(columnType, columnSize);
     }
 
-    private boolean needsQuoting(String alias, String identifierQuoteString) {
+    public boolean needsQuoting(String alias, String identifierQuoteString) {
         boolean result = false;
         if (alias != null && identifierQuoteString != null) {
             if (alias.indexOf(identifierQuoteString) == -1) {
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/DefaultQueryRewriter.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/DefaultQueryRewriter.java
@@ -101,7 +101,7 @@ public class DefaultQueryRewriter extends AbstractQueryRewriter {
         return super.rewriteColumnType(columnType, columnSize);
     }
 
-    public boolean needsQuoting(String alias, String identifierQuoteString) {
+    protected boolean needsQuoting(String alias, String identifierQuoteString) {
         boolean result = false;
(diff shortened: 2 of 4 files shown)


jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/DefaultQueryRewriter.java
```diff
@@ -101,7 +101,7 @@ public String rewriteColumnType(ColumnType columnType, Integer columnSize) {
         return super.rewriteColumnType(columnType, columnSize);
     }
 
-    private boolean needsQuoting(String alias, String identifierQuoteString) {
+    public boolean needsQuoting(String alias, String identifierQuoteString) {
```
kaspersorensen: I would suggest `protected` scope instead of `public`

jdbc/src/test/java/org/apache/metamodel/jdbc/HsqldbTest.java
```diff
@@ -417,4 +418,98 @@ public void testInterpretationOfNull() throws Exception {
         Connection conn = DriverManager.getConnection("jdbc:hsqldb:mem:interpretation_of_null", USERNAME, PASSWORD);
         JdbcTestTemplates.interpretationOfNulls(conn);
     }
+    
+    public void testCapitalization() throws Exception{
+      try(Connection connection = DriverManager.getConnection("jdbc:hsqldb:mem:different_types_insert", USERNAME,PASSWORD);){
```
kaspersorensen: Coding conventions/formatting rule: There should be a space between "try" and "(".

(same rule is relevant other places in the diff, this is just an example)

jdbc/src/test/java/org/apache/metamodel/jdbc/HsqldbTest.java
```diff
@@ -417,4 +418,98 @@ public void testInterpretationOfNull() throws Exception {
...
+        assertEquals(idCol, dcon.getDefaultSchema().getTable(0).getColumn(0).getName());
+        
+        dcon.query().from(mixedcap).select(idCol,nameCol,emailCol).execute();
+        
+//        Query query = new Query();
+//        String subselect = "(select * from \""+mixedcap+"\")";
+//        query.from(subselect);
+//        query.select("test");
+//        
+//        dcon.executeQuery(query);
```
kaspersorensen: I would leave out / remove commented code

core/pom.xml
```diff
@@ -32,6 +32,10 @@ under the License.
 			<artifactId>slf4j-api</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>com.google.guava</groupId>
+			<artifactId>guava</artifactId>
```
kaspersorensen: We should only add the dependency where necesary. You are using guava in the JDBC module, so you should add it as a dependency there instead of the `core` module.

kaspersorensen:
The change looks good to me, but I had a few review remarks that I'd prefer us taking care of first.

core/pom.xml
tomatophantastico: i add it in an other pull request to core, so i added it here too, otherwise it would not compile.

jdbc/src/test/java/org/apache/metamodel/jdbc/HsqldbTest.java
tomatophantastico: removed it.

This code should demonstrate some other issue i had, but it should better be placed in an other test

jdbc/src/test/java/org/apache/metamodel/jdbc/HsqldbTest.java
tomatophantastico: beautified the test

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/DefaultQueryRewriter.java
tomatophantastico: agreed

kaspersorensen:
Patched (but I moved the Guava dependency to the jdbc module).

PR #152, opened 2017-07-23
Added all recurring devs and contributors to pom.xml

kaspersorensen:
I was looking at our pom.xml file and stumbled across the developers/contributors section. It seemed very outdated to me so I took the liberty of updating it based on the contributors that GitHub could tell me about. This seems much more representative and up to date. If I missed anyone, I apologize. Would feel happy to add others too :-)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -105,18 +105,52 @@ under the License.
 	</organization>
 	<developers>
 		<developer>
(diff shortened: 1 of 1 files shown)


kaspersorensen:
Merging as per lazy consensus

PR #147, opened 2017-06-13
Merge `5.x` into master branch

kaspersorensen:
I've just raised a VOTE for releasing version `5.0-RC2`. With that milestone I think that we're getting close to a final release of 5.x and as preparation for that we should re-integrate the branch back into the master branch.

This PR does just that :-)

PR #153, opened 2017-07-28
Collections instead of Arrays 

tomatophantastico:
This solves METAMODEL-7

This is a major API change, proceed with caution.

Lists are used in cases where order is important (schemas, tables, columns)
Collection is used, when no natural order is obvious (Relations)

Feedback is appreciated.


core/src/main/java/org/apache/metamodel/CompositeDataContext.java
```diff
@@ -190,9 +183,7 @@ public Schema getSchemaByNameInternal(String name) throws MetaModelException {
                 }
             }
         }
-        String[] result = set.toArray(new String[set.size()]);
-        Arrays.sort(result);
-        return result;
+        return new ArrayList<>(set);
```
LosD: Doesn't order still need to be preserved here, at least to keep the current behavior the same?

core/src/main/java/org/apache/metamodel/DeleteAndInsertBuilder.java
```diff
@@ -103,7 +103,10 @@ private Row update(final Row original) {
         final DataContext dc = _updateCallback.getDataContext();
         final Table table = getTable();
         final List<FilterItem> whereItems = getWhereItems();
-        final DataSet dataSet = dc.query().from(table).select(table.getColumns()).where(whereItems).execute();
+        final DataSet dataSet = dc.query()
+                .from(table)
+                .select(table.getColumns().toArray(new Column[table.getColumns().size()]))
```
LosD: This is a bit of a shame. No good way to avoid the extra `toArray()`? It seems you added an overriden method to `Query` that does the same, but I might be missing something here.

core/src/main/java/org/apache/metamodel/convert/Converters.java
```diff
@@ -188,7 +191,7 @@ private static void autoDetectConvertersInternally(DataContext dataContext, Tabl
             detectors.put(column, new ColumnTypeDetector());
         }
 
-        Query query = dataContext.query().from(table).select(columns).toQuery();
+        Query query = dataContext.query().from(table).select(columns.toArray(new Column[columns.size()])).toQuery();
```
LosD: Same extra toArray here (just marking them in case I'm right that we can avoid them).

core/src/main/java/org/apache/metamodel/convert/Converters.java
```diff
@@ -136,15 +135,19 @@ public static DataContext addTypeConverters(DataContext dataContext, Map<Column,
...
-            autoDetectConvertersInternally(dataContext, table, tableColumns, sampleSize, result);
-        }
+        columns.stream()
+                .filter(col -> col.getType() != null)
+                .filter(col -> col.getType().getSuperType().equals(SuperColumnType.LITERAL_TYPE))
+                // group by table
+                .collect(Collectors.toMap(Column::getTable,Arrays::asList))
+                //and detect it
+                .forEach((tab, cols) ->{
+                    autoDetectConvertersInternally(dataContext,tab,cols,sampleSize,result);
```
LosD: Extremely minor, but this doesn't look like it needs braces?

core/src/main/java/org/apache/metamodel/data/AbstractRowBuilder.java
```diff
@@ -68,7 +71,7 @@ public AbstractRowBuilder(Column[] columns) {
 
     @Override
     public final Row toRow() {
-        return new DefaultRow(new SimpleDataSetHeader(_columns), _values);
+        return new DefaultRow(new SimpleDataSetHeader(Arrays.stream(_columns).map(SelectItem::new).collect(Collectors.toList())), _values);
```
LosD: Looks like it might be an advantage to make `_columns` a list here, instead. Of course that means a whole lot more changes in the class, so it certainly also makes sense not to.

core/src/main/java/org/apache/metamodel/CompositeDataContext.java
tomatophantastico: This is a Treeset now, see line 175, order is insertion order.

The real Question here is: should getSchemas() be List or Collection?

core/src/main/java/org/apache/metamodel/DeleteAndInsertBuilder.java
tomatophantastico: right, at first i did not want to change the query interface, but latter added during the refactor, because it was really missing.

Updated it

core/src/main/java/org/apache/metamodel/data/AbstractRowBuilder.java
tomatophantastico: Array and List have a slight different semantic.
My reasoning was, if i can grasp the code within a minute, i changed internal methods, otherwise preserve it.

This is especially the case with the XML-Dom-related stuff.

tomatophantastico:
w.r.t. guavas Lists.newArrayList() 
I think i used it mainly in the testcases, purely for syntactic convenience.

Internally ArrayLists should be used as it allows for more freedom on how to create the objects, and then wrap with Collections.unmodifiableList, as for example done in ImmutableTable.

core/src/main/java/org/apache/metamodel/convert/Converters.java
tomatophantastico: true

core/src/main/java/org/apache/metamodel/data/AbstractRowBuilder.java
LosD: Makes a lot of sense. We can always improve later

core/src/main/java/org/apache/metamodel/CompositeDataContext.java
LosD: Ah, missed that. Perfect!

Not sure about List vs. Collection. I think I personally prefer List to avoid wrapping it before working with it, but it's not a huge deal either way.

LosD:
I don't really agree on "ArrayLists should be used", especially for single-element lists, as a singletonList does not need a backing array, but it's not really a blocker for me.

tomatophantastico:
Sorry, this was more of the reasoning behind my bulk changes & the changes in the (im)mutable table class, the implementor should of course pick what fits the scenario best.

kaspersorensen:
I also wanna applaude this big effort :-) But like @LosD I am a bit concerned with adding Guava to the core module because that means our API and really core classes essentially depends on Guava. I think it would be best to keep it out because we're really dragging in transitive dependencies for all our users otherwise.

tomatophantastico:
IIRC, guava is only in the test cases, so i would just set the scope to test

kaspersorensen:
Great. That makes sense. I expect to take a closer look at the diff sometime this weekend.

kaspersorensen:
I see a lot of unused imports on this branch:
image

kaspersorensen:
Other than the compiler warnings, the code looks good to me.

tomatophantastico:
I manually rearranged the imports.

To make things easier improve quality i also created a checkstyle integration in https://github.com/tomatophantastico/metamodel/tree/feature/checkstyle

As i did not base the checkstyle branch on '5.x.x' and instead on top of this branch, there is currently no separate pull request.

But i think it is a good idea to have some (more) automated testing.

kaspersorensen:
Can you open a PR for the checkstyle improvement then also? :-)

Btw. the 5.x branch is no longer in use really. We've merged it into master as of #147.

I will take a look at this PR again as soon as I get a chance. Unfortunately I'm not getting a lot of time these past days, but will try to take aside some time.

kaspersorensen:
I've only got more compiler warnings that I'd like to fix before I give this my +1.
I've submitted those in a PR for you: https://github.com/tomatophantastico/metamodel/pull/1

kaspersorensen:
+1

kaspersorensen:
I'll go ahead and merge this, good work. Thank you for the considerable contribution.

PR #154, opened 2017-08-04
Checkstyle intergration (and collections)

tomatophantastico:
This pull is on top of pull request #153.

I opened a new pull request as this is technically a new feature, #153 is however included in this one. It is not as large as it seems. Only the last 4 commits are relevant.

Integrates checkstyle into the maven test phase: so when ever mvn test is called, checkstyle is run and fails on violations.

I only activated three rules:

unused imports -> self-explanatory
import order -> enforce eclipse ordering, therefore smaller commits, when using different IDEs
licence -> rat checks if present, this one however enforces 1:1 identity with the context of the text file. I found and corrected some deviations


kaspersorensen:
can you merge in the latest master so that the diff will become reviewable here? :-)

PR #156, opened 2017-08-07
Checkstyle integration

tomatophantastico:
Integrates checkstyle into the maven test phase: so when ever mvn test is called, checkstyle is run and fails on violations.

This should be subject to discussion, especially with more controversial code styling rules (tabs/spaces anyone?)

I only activated three rules:

unused imports -> self-explanatory
import order -> enforce eclipse ordering, therefore smaller commits, when using different IDEs
licence -> rat checks if present, this one however enforces 1:1 identity with the context of the text file. I found and corrected some deviations


LosD:
Travis failure seems unrelated. Cassandra startup timed out.

LosD:
I'm a big fan of using CheckStyle (but not that there's apparently no GitHub integration to be found. :disappointed:), so I'm all for it. Also for more aggressive rules.

tomatophantastico:
travis can be used to enforce checkstyle, actually, by integrating it into the maven build, commits are checked (afterwards)

tomatophantastico:
More interestingly: Which rules should be employed?
What are the current rules?
Is a switch to google/sun rules an option?

LosD:
About Travis for CheckStyle, I know (we use it like that in DataCleaner), but a proper Github integration would be much better, IMHO. More visible and less annoying ☺️

Unfortunately, I haven't found anything. 

Regarding rules, I always thought that the standard sets were pretty weird. For building them, it's rather hard to come up with a good set that CheckStyle likes, and works well with most editor's formatters. Especially indentation is a bastard, e.g last time I looked, CheckStyle had major problems with Lambda expressions, and getting Eclipse, CheckStyle and IDEA to agree on trinary expressions and where to break was exceptionally hard.

core/src/main/java/org/apache/metamodel/UpdateSummaryBuilder.java
```diff
@@ -1,4 +1,5 @@
 /**
+ * Licensed to the Apache Software Foundation (ASF) under one
```
kaspersorensen: Hmm I'm surprise that our current license-check (RAT) plugin doesn't catch this. But definately important to get fixed.

pom.xml
```diff
@@ -330,6 +330,27 @@ under the License.
...
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-checkstyle-plugin</artifactId>
+				<version>2.17</version>
+				<executions>
+					<execution>
+						<id>test</id>
+						<phase>test</phase>
+						<configuration>
+							<configLocation>style/checkstyle.xml</configLocation>
```
kaspersorensen: The folder name "style" doesn't look quite right to me. Actually I think that unless there are other files than this file, we should just put it in the root folder.

kaspersorensen: Eh ok ... I should have read a bit further! :-) I think we should rename the folder to "checkstyle" then.

As for the license header check - we already have the `apache-rat-plugin` which also does some of this. Could seem that it's less complete though. We might want to sync them up to avoid double work.

style/checkstyle.xml
```diff
@@ -0,0 +1,46 @@
...
+specific language governing permissions and limitations
+under the License.
+-->
+<!DOCTYPE module PUBLIC
+    "-//Puppy Crawl//DTD Check Configuration 1.2//EN"
+    "http://www.puppycrawl.com/dtds/configuration_1_2.dtd">
+
+<module name="Checker">
+	<module name="SuppressionFilter">
+    <property name="file" value="style/checkstyle_suppressions.xml"/>
```
kaspersorensen: Mixed indentation here. Which begs the question: Can checkstyle check the indentation of xml files? :-)

kaspersorensen:
My opinion of checkstyle is a bit mixed. Essentially I feel that there is a pitfall which is that it makes it hard for people to contribute good code if their tool has very minor discrepancies with the code style. That said, I think a minimum set is very nice to have in place...

One of the things I notice here though is the import statement ordering. I believe that 99% of our codebase today is sorted using eclipse's standard settings. If we require some other ordering then my next question is: How do we explain to contributors (myself included) how to set up their IDE. This should essentially be documented, probably in the CONTRIBUTE.md file.

pom.xml
tomatophantastico: w.r.t. foldername: 
i'd prefer naming after function and not implementation, but not really an issue to.
Is the default eclipse formatter used by everyone? Custom files might go here as well.

core/src/main/java/org/apache/metamodel/UpdateSummaryBuilder.java
tomatophantastico: Surprised me as well, initally i put this check in place only for testing purposes, but differences in multiple locations were found, i commited it.
As checkstyle does a 1:1 check, also the /** vs /* issue raised in membrane can be centrally defined here.

tomatophantastico:
minimum set is a good idea. I used intellij idea and i am not to terribly familiar with it yet, so this contributed to not so clean commits.
Can we collect some issues regarding formatting that recently came up in commits? The ones i got, are basically already in this pull request.

pom.xml
```diff
@@ -330,6 +330,27 @@ under the License.
...
 			</plugin>
 
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-checkstyle-plugin</artifactId>
+				<version>2.17</version>
+				<executions>
+					<execution>
+						<id>test</id>
+						<phase>test</phase>
```
tomatophantastico: There is no need to attach it to the test phase, if you think this scares of contributions it can be moved to mvn verify and adapt the travis script, so it will show up in the pull request.

Thinking about it, this is maybe the way to go.

.travis.yml
```diff
@@ -10,12 +10,15 @@ before_install:
...
 
+script: 
+  - mvn verify
+
 services:
   - couchdb
   - mongodb
   
 after_success:
-  - mvn test javadoc:javadoc
```
tomatophantastico: I removed the call to mvn test, as test is in verify

style/checkstyle.xml
```diff
@@ -18,29 +18,47 @@ specific language governing permissions and limitations
...
-	<module name="Header">
-	    <property name="headerFile" value="style/licence_header.txt"/>
-	    <property name="fileExtensions" value="java"/>
-	</module>
+    <module name="Header">
+        <property name="headerFile" value="style/licence_header.txt"/>
+        <property name="fileExtensions" value="java"/>
+    </module>
+    
+    <module name="FileTabCharacter">
```
tomatophantastico: The contribute.md states that spaces are used instead of tabs.
A lot of files do not stick to this convention.

kaspersorensen: Cool, but now the build fails then so somehow we need to fix it. Either you go ahead and fix it (you certainly have my blessing) or we comment out this rule and make it a separate issue to fix.

LosD:
@kaspersorensen Regarding explaining settings, we could just export the formatting rules from the IDEs, and explain in CONTRIBUTE.md to use those. At least Eclipse and IntelliJ has that option.

The hard part is making them agree 100%... E.g. Eclipse and IDEA has rather different ways of formatting ternary operators, especially when there's also method chains involved. And Eclipse has this lame-o thing with grouping top-level domains (org, com, dk, etc.) in own sort groups (seemingly in reverse alphabetical order, just because), instead of just java and javax. IDEA doesn't have that option, so you need to specify those groups manually. Eclipse also has this (very nice) feature with separate line lengths for JavaDoc and code which IDEA doesn't have.

We ran into other things for DataCleaner, but those are the ones I remember best. We ended up with these two formatters:
https://github.com/datacleaner/DataCleaner/blob/master/Formatter-Eclipse.xml
https://github.com/datacleaner/DataCleaner/blob/master/Formatter-IDEA.xml

They're not perfect (and I don't think they can be, unless the Eclipse and IDEA folks sit together and decide to make the formatters compatible), but they're reasonable. Most files has no changes when formatted by one and then the other, and for those that does is usually only a line or two.

(If we end up using this as a "format everything" commit, can we make sure to squash it, and make _very_ clear in the commit message that there's no functional changes in it (and of course make sure that's true.). It can be quite a PITA to do annotations and other history browsing otherwise?)

kaspersorensen:
Or we could omit having any rules about imports? I mean, it's never in the imports section that I find anything hard to compare when looking at diffs. So even if it goes forwards-and-backwards all the time, I don't personally care much.

Other rules that I can think of:

* Requiring a license header is legally required, so can't be discussed really :-)
* Enforcing the same indentation is something I really want checkstyle to enforce. Thumbs up for automating that fully.
* Requiring `final` is nice, but takes a lot of effort to retrofit into all the places where it doesn't exist. I would also argue that it's less important in many places, e.g. test code and such. For me it is a perfectly fine thing to just take on a code review basis.
* Maximum argument count is an annoying type of check IMO. Especially when building immutable classes that often take a lot of constructor arguments. And I like immutability much more than checkstyle :-) So I would vote against such a check.
* Method and class complexity rules I also dislike. I've seen too many cases where I felt that it made the code more complex to split it into tons of private methods and stuff like that. I would be fine with a warning or something like that, but essentially this comes down to a human judgement call where we have to evaluate "is this method appropriate in complexity to the issue it is solving".

LosD:
I disagree about the last two points. Both are clear indicators of misdesigned classes. If a constructor or method requires that many arguments, the class represents too much, or the method has too many jobs.

Same with complexity. A properly designed class does not need complex methods.

Of course there are exceptions. But in those cases it can be disabled using comments (but there should be a good argument for doing that). 

The best argument against it is that there's a lot of terrible code (when looking at with argument list length and complexity googles), so a lot of exceptions are needed. And since we're way too generous with public classes, all core classes are essentially API classes, so we can't even fix those step by step, but have to change them big-bang style, which scares the hell out of me. OTOH, it's nice to highlight classes that needs work when a chance shows itself. But I must admit we never enabled it in DataCleaner for that exact reason.

kaspersorensen:
I understand your point and respectfully disagree :-) BUT I do think it is worthy of a VOTE on the mailing list! So how about we organize a vote where we list these various checks and let people vote +1 or -1 on the individual checks?

LosD:
Makes sense! 

Though some of them are hard. E.g. indentation because checkstyle doesn't understand lambda expressions, so you can't get it to agree with the formatters. So I'd have to vote no there, even if I agree in principle. And it would require that we get IDEs and checkstyle to agree. I'd absolutely refuse to contribute PRs if it requires me to install Eclipse to make them compliant.

LosD:
(Of course, there's a chance that checkstyle fixed the lambda thing since I checked)

tomatophantastico:
First, at this moment i do not see a merit in voting about complexity rules:
As long as there is no one out there who wants to do refactor those massive amounts if code to fit those rules, there is no need to codify it. 
Reworks of code, i guess, should always improve their readability.

Second, the JDBC test is failing again, locally this works, any idea what may cause this?

Third, the lambdas (@LosD ): This seems to be fixed.
However, as the Contribute.md states indention is 4 spaces, they are indented the same level as code blocks. IMHO this is ok, but not grate and not the default behaviour of most IDE. In the code you can find different indentions (4 & 6 spaces).

Indention is currently just a warning, so you might want to check yourself what you prefer.

kaspersorensen:
Sorry for not paying attention for a long time. I checked the Travis build error and it seems to be some general JVM crash. Not sure what happened, but I doubt it is build related at all. Maybe the VM that Travis provisioned had some issue.

Can you make another commit to the branch, that will trigger a new build on Travis. I'll also do a local build just to verify.

LosD:
No need to do a commit, just log in to Travis and trigger it.

kaspersorensen:
I don't think we can just do that. ASF owns the Travis instance and the builds are only automatically triggered AFAIK.

LosD:
Ah, that might be.

kaspersorensen:
Hey @tomatophantastico, all,

I made https://github.com/tomatophantastico/metamodel/pull/2 some time ago to help you fix some of the merge conflicts and minor issues. Are you still around? Can you take a look? Or else I think we need to close this PR.

PR #159, opened 2017-08-25
METAMODEL-1160 and METAMODEL-1163: Deserialization of legacy schema objects

kaspersorensen:
Fixes both METAMODEL-1160 and METAMODEL-1163.


LosD:
Not really a readable PR, lots of unnecessary reformatting going on.

kaspersorensen:
Oh damn, that's a tabs vs spaces thing... The old version of those files seems to have tabs, contrary to our guidelines.

Is it possible to read if you add the w=1 param like this:
https://github.com/apache/metamodel/pull/159/files?w=1
?

Or should I reproduce the tabs in those files?

LosD:
I know, but you shouldn't be doing whole-file reformats at all. Even with w=1, there's still big reformats.

LosD:
We can do like we did in DataCleaner: Have one whole project commit that does whole-project reformat. It's still a bit annoying history-wise, but since it will be clearly marked as a reformatting commit, it's easy for anyone to see that they just need to go back one more commit to look for functional changes.

When it's done in commits like this one, it's really, really annoying to hunt through the history for causes of bugs.

kaspersorensen:
Superceeded by #160

PR #160, opened 2017-08-30
METAMODEL-1160 and METAMODEL-1163: Fixed

kaspersorensen:
Re-did the same changes as in #159 but this time without all the reformatting of complete classes :-)


core/src/main/java/org/apache/metamodel/schema/MutableRelationship.java
```diff
@@ -134,4 +139,17 @@ private MutableRelationship(List<Column> primaryColumns, List<Column> foreignCol
 		return _foreignColumns;
 	}
 
+    private void readObject(ObjectInputStream stream) throws IOException, ClassNotFoundException {
+        final GetField getFields = stream.readFields();
+        Object primaryColumns = getFields.get("_primaryColumns", null);
+        Object foreignColumns = getFields.get("_foreignColumns", null);
+        if (primaryColumns instanceof Column[] && foreignColumns instanceof Column[]) {
```
LosD: Are these always tied together, or could one be an array and the other not?

kaspersorensen: In our history it has only been either-or. Of course someone (slightly crazy) could of course change one of them back to an array and then we'd have an interesting new issue to work on :-) But I argue that the only scenario worth worrying about now is that it's the singular "old class format" where they both where arrays.

LosD: In that case, LGTM! I'll merge

PR #163, opened 2017-10-01
METAMODEL-1165: Fixed queries w/o table names (syntax: "FROM tables[0]")

kaspersorensen:
Fixes https://issues.apache.org/jira/browse/METAMODEL-1165 by adding a "tables[0]" syntax support

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@
  * [METAMODEL-1151] - Added DataContextFactory classes for instantiating DataContexts of many types based on properties.
  * [METAMODEL-1160] - Fixed bug when deserializing v4.x CsvTable objects
  * [METAMODEL-1163] - Fixed bug when deserializing v4.x MutableRelationship and ImmutableRelationship objects
+ * [METAMODEL-1165] - Added support for querying by table index instead of name, ie. "FROM tables[0]".
 
 ### Apache MetaModel 4.6.0
 
--- a/core/src/main/java/org/apache/metamodel/AbstractDataContext.java
+++ b/core/src/main/java/org/apache/metamodel/AbstractDataContext.java
@@ -387,7 +387,7 @@ public abstract class AbstractDataContext implements DataContext {
         if (tableName == null) {
             return null;
         }
-
+        
         final String[] tokens = tokenizePath(tableName, 2);
         if (tokens != null) {
             Schema schema = getSchemaByToken(tokens[0]);
--- a/core/src/main/java/org/apache/metamodel/schema/AbstractSchema.java
+++ b/core/src/main/java/org/apache/metamodel/schema/AbstractSchema.java
@@ -22,6 +22,8 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import org.apache.metamodel.util.EqualsBuilder;
@@ -37,6 +39,8 @@ public abstract class AbstractSchema implements Schema {
     private static final long serialVersionUID = 1L;
 
     private static final Logger logger = LoggerFactory.getLogger(AbstractSchema.class);
+
+    private static final Pattern TABLE_ARRAY_TOKEN_PATTERN = Pattern.compile("^tables\\[(\\d+)\\]$");
 
     @Override
     public final String getQuotedName() {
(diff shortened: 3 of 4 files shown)


kaspersorensen:
Before I go ahead and merge this based on Lazy Concensus, I'd like to ask one more time if anyone has any feedback on the new feature. To me it seems pretty cool, but others may feel otherwise.

@LosD maybe a comment? Or @ardlema?

LosD:
On the way home now, will take a proper look when I'm home.

LosD:
Took a look (phone code-browsing was not as bad as I thought it would be) . Code looks fine, but the whole idea is pretty terrible. We have plenty of problems with datastores being vulnerable to order change (just look at our integration tests where half the issues are due to order being assumed), and this will make it worse.

As I read from the issue, its MEANT for single table datastores, but that's not the implementation. This _will_ be used for more and it_will_ break when they update their databases.

I'd either drop it completely or find a format that can only be used for single-table datastores, not even for multi table datastores that currently contains a single table.

kaspersorensen:
If I was to paraphrase then you're saying that just because not every `List` is ordered, you shouldn't have a `get(int)` method? I don't think that's reasonable :-) I agree that it might not be what you want to use in many (most) situations. But if you know that you're looking for "table no. X" then it adds a small convenience. Or maybe you know the amount of tables, but not their names, and you want to fire a query to select the count of all tables?

LosD:
Every list IS ordered. That's part of the contract of the List  interface...

LosD:
Anyway, up to you. I still think it's a horrible idea, a small convenience is no worth it for a major source of bugs in client code... Even bugs that will show itself at unexpected times (again, try to run our Integration tests. They're mostly broken due to having silly assumptions that this would encourage more of, and DB upgrades reordering tables and schemas).

kaspersorensen:
I don't think that the right way forward is to let it be "up to me" or anyone else for that matter. We should find some kind of concensus or agreement, at least between the people who engage in MetaModel's direction. So from that angle, I think we should either discuss more or drop the feature :-) Just to confirm then - you would be more in favor of introducing a keyword specifically for single-table datastores?

LosD:
Something like that sounds like a better idea to me, yes. 

Suggestions:
1. Keep the format, but reject it for all multi-table Datastores. Might of course be a little weird, interface-wise (an array selector that only works when there's no array :slightly_smiling_face:).
2. If the parser can handle it, just allow single-table data to skip the FROM. A bit odd-looking to me though.
3. Allow something like "default_table" as well as the actual table name for single-table datastores.
4. Require the FROM-clause, but ignore the table name for single-table datastores.

tomatophantastico:
IMHO: i prefer the query language to be free of "special constructs" as much as possible. 

I encoutered a similar problem as described in mm-1165 and my workaround was to construct the query programatically.
But i was not able to externalize the SQL string with CSVs, as the schema depends on the serialisation of the data, e.g. the table name depends on the file name.

To break this dependency and allow reuse of SQL, a better idea would be to break the tablename/filename connection and allow either 
- a custom table name on constructing single table data contexts
- a better default value on constructing single table data contexts
-  an alias in the data context.

What is the reason using the filename for tables? To avoid clashes in the composite DataContext?

kaspersorensen:
I'm not opposed to _also_ adding better support for choosing your own table names and such, but yes, the idea behind using the filename as table name was to allow expansion of the logical DataContext either via CompositeDataContext or maybe by a future implementation where a folder of files could be opened as one schema.

When you mention "alias in the data context", you mean another Table with TableType=ALIAS? I can see how that would make it really clear. OTOH that would then have to be implemented on a data context basis rather than as a generic construct in the parser. If I can draw a parallel to `SELECT *` then I would say that this is a parsing/query level convenience that I find really useful whenever I'm feeling lazy and don't want to figure out column names. I am kind of hoping for something similar just for table names.

tomatophantastico:
To me, there are two use cases / requirements here:

1. Make queries transferable
2. Ease of exploring a schema and its data via SQL

w.r.t 1)
Yes, i think the data contexts are the place where this should be implemented. As most (all?) non-jdbc datacontexts are extending QueryPostProcessingContext, this could perhaps be done without too many changes.

i generally think it is not a good idea to deal with a table "Backup_23423_hash_by_Jörg_do_not_modify.csv" in the query, especially if it is an export of the lets say person table. Having an alias would clear up the semantics of the query.

"select * from person" will work on the source system, "select * from tables[0]" not. 


w.r.t. 2)
Just iterating over tables without having the table name, does not really make sense to me. How should the result of such a query look like?

Most DBMs implement this via custom commands (list tables/documents...) or using tables (SELECT table_schema,table_name FROM information_schema.tables).

Maybe we can cook up sth. more convenient here. But i would prefer to change the semantics of SQL.

kaspersorensen:
OK I guess I can get behind using an ALIAS table too. My only worries are around (1) user interfaces such as DataCleaner's which may now become cluttered with additional tables and (2) composite datacontexts built from e.g. 2 csv files, which would then have two ALIAS tables of same name. But both concerns can be fixed, I think, by filtering out irrelevant table types.

PR #164, opened 2017-10-18
METAMODEL-86: Allow transactional querying in JDBC update scripts

kaspersorensen:
Here's my attempt at solving METAMODEL-86. I wasn't sure how to test it so ... It's not really tested, except existing tests should pass obviously. Review accordingly, please :-)

### DIFF
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcUpdateCallback.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcUpdateCallback.java
@@ -34,6 +34,7 @@ import org.apache.metamodel.create.TableCreationBuilder;
 import org.apache.metamodel.delete.RowDeletionBuilder;
 import org.apache.metamodel.drop.TableDropBuilder;
 import org.apache.metamodel.insert.RowInsertionBuilder;
+import org.apache.metamodel.jdbc.JdbcUtils.JdbcActionType;
 import org.apache.metamodel.schema.Schema;
 import org.apache.metamodel.schema.Table;
 import org.apache.metamodel.update.RowUpdationBuilder;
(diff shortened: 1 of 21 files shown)


kaspersorensen:
Update: I found a way to unit test this and a good thing I did too. The approach I had built wasn't really working since the query method I used closed connection. That's fixed now. Furthermore the unit test revealed that such update scripts can easily cause deadlocks at the DB level. These will be rolled back, but the client needs a good way to catch them, which is why I added the `RolledBackUpdateException`.

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcUpdateCallback.java
```diff
@@ -156,20 +158,25 @@ public final TableCreationBuilder createTable(Schema schema, String name) throws
...
+        return (JdbcDataContext) super.getDataContext();
     }
 
     // override the return type to the more specific subtype.
     @Override
-    public final JdbcDataContext getDataContext() {
-        return (JdbcDataContext) super.getDataContext();
+    public final DataContext getDataContext() {
+        final Connection connection = getConnection();
+        return new JdbcUpdateCallbackDataContext(getJdbcDataContext(), connection);
```
jakubneubauer: Ok, this hack took me a while before I got the idea :-) Clever trick how to make the associated DataContext use our connection

kaspersorensen: Clever good or clever bad? :-) Are we good to merge you think?

jakubneubauer:
Now I checked the rest. Also executed few tests on real database:
- The connection obtained from the UpdateScript used for queries is the same as used for updates - OK
- Transaction is not committed during the UpdateScript duration nor after updates/inserts/deletes nor after the queries.
- On the other getting connection on DataContext outside of UpdateScript returns new connection - OK
- Throwing any exception from the update script causes rollback.

Code looks fine, I recommend merge :-)

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcUpdateCallback.java
jakubneubauer: it is fine I think, it was just neutral observation ;-)

PR #165, opened 2017-10-20
METAMODEL-1165: Add alias default table

kaspersorensen:
Another approach to METAMODEL-1165. This time with the use of alias tables. Following the feedback from @LosD and @tomatophantastico from my previous PR on the same issue.

### DIFF
--- a/core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
+++ b/core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
@@ -76,11 +76,17 @@ public abstract class QueryPostprocessDataContext extends AbstractDataContext im
     public static final String SYSTEM_PROPERTY_CREATE_DEFAULT_TABLE_ALIAS = "metamodel.alias.default.table";
     public static final String INFORMATION_SCHEMA_NAME = "information_schema";
 
-    private final Map<Column, TypeConverter<?, ?>> _converters;
+    private final Map<Column, TypeConverter<?, ?>> converters;
+    private final boolean defaultTableEnabled;
 
     public QueryPostprocessDataContext() {
+        this(true);
+    }
+
+    public QueryPostprocessDataContext(boolean defaultTableEnabled) {
         super();
-        _converters = new HashMap<Column, TypeConverter<?, ?>>();
+        this.defaultTableEnabled = defaultTableEnabled;
+        this.converters = new HashMap<Column, TypeConverter<?, ?>>();
     }
 
     @Override
@@ -385,7 +391,7 @@ public abstract class QueryPostprocessDataContext extends AbstractDataContext im
 
             // conversion is done at materialization time, since it enables
             // the refined types to be used also in eg. where clauses.
-            dataSet = new ConvertedDataSetInterceptor(_converters).intercept(tableDataSet);
+            dataSet = new ConvertedDataSetInterceptor(converters).intercept(tableDataSet);
         }
 
         return dataSet;
@@ -451,8 +457,8 @@ public abstract class QueryPostprocessDataContext extends AbstractDataContext im
 
         if (name == null || name.equalsIgnoreCase(mainSchemaName)) {
             final Schema mainSchema = getMainSchema();
-            final boolean createAliasTable =
-                    Boolean.parseBoolean(System.getProperty(SYSTEM_PROPERTY_CREATE_DEFAULT_TABLE_ALIAS, "true"));
+            final boolean createAliasTable = defaultTableEnabled
+                    && Boolean.parseBoolean(System.getProperty(SYSTEM_PROPERTY_CREATE_DEFAULT_TABLE_ALIAS, "true"));
             if (createAliasTable) {
                 return DefaultTableAliasedSchema.wrapIfAppropriate(mainSchema);
             }
@@ -587,7 +593,7 @@ public abstract class QueryPostprocessDataContext extends AbstractDataContext im
      */
     @Override
     public void addConverter(Column column, TypeConverter<?, ?> converter) {
-        _converters.put(column, converter);
+        converters.put(column, converter);
     }
 
     /**
--- a/core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
+++ b/core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
@@ -77,15 +77,23 @@ public abstract class QueryPostprocessDataContext extends AbstractDataContext im
     public static final String INFORMATION_SCHEMA_NAME = "information_schema";
 
     private final Map<Column, TypeConverter<?, ?>> converters;
-    private final boolean defaultTableEnabled;
+    private final boolean singleTableDatastore;
 
     public QueryPostprocessDataContext() {
         this(true);
     }
 
-    public QueryPostprocessDataContext(boolean defaultTableEnabled) {
+    /**
+     * 
+     * @param singleTableDatastore a flag that, if set to true, indicates that this DataContext contains just a single
+     *            table. This information will be used to optimize and provide convenience for the implementation. An
+     *            additional {@link Table} of type {@link TableType#ALIAS} with the name "default_table" will be
+     *            automatically added in addition to the single table. That again makes for convenient querying of the
+     *            single table using a predictable name/alias.
+     */
+    public QueryPostprocessDataContext(boolean singleTableDatastore) {
         super();
-        this.defaultTableEnabled = defaultTableEnabled;
+        this.singleTableDatastore = singleTableDatastore;
         this.converters = new HashMap<Column, TypeConverter<?, ?>>();
     }
 
@@ -457,7 +465,7 @@ public abstract class QueryPostprocessDataContext extends AbstractDataContext im
 
         if (name == null || name.equalsIgnoreCase(mainSchemaName)) {
             final Schema mainSchema = getMainSchema();
-            final boolean createAliasTable = defaultTableEnabled
+            final boolean createAliasTable = singleTableDatastore
                     && Boolean.parseBoolean(System.getProperty(SYSTEM_PROPERTY_CREATE_DEFAULT_TABLE_ALIAS, "true"));
             if (createAliasTable) {
                 return DefaultTableAliasedSchema.wrapIfAppropriate(mainSchema);
--- a/core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
+++ b/core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
@@ -50,6 +50,7 @@ import org.apache.metamodel.query.SelectClause;
 import org.apache.metamodel.query.SelectItem;
 import org.apache.metamodel.schema.Column;
 import org.apache.metamodel.schema.ColumnType;
+import org.apache.metamodel.schema.DefaultTableAliasedSchema;
 import org.apache.metamodel.schema.MutableColumn;
 import org.apache.metamodel.schema.MutableRelationship;
 import org.apache.metamodel.schema.MutableSchema;
@@ -63,17 +64,16 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * Abstract DataContext for data sources that do not support SQL queries
- * natively.
+ * Abstract DataContext for data sources that do not support SQL queries natively.
  * 
- * Instead this superclass only requires that a subclass can materialize a
- * single table at a time. Then the query will be executed by post processing
- * the datasets client-side.
+ * Instead this superclass only requires that a subclass can materialize a single table at a time. Then the query will
+ * be executed by post processing the datasets client-side.
  */
 public abstract class QueryPostprocessDataContext extends AbstractDataContext implements HasReadTypeConverters {
 
     private static final Logger logger = LoggerFactory.getLogger(QueryPostprocessDataContext.class);
 
+    public static final String SYSTEM_PROPERTY_CREATE_DEFAULT_TABLE_ALIAS = "metamodel.alias.default.table";
     public static final String INFORMATION_SCHEMA_NAME = "information_schema";
 
     private final Map<Column, TypeConverter<?, ?>> _converters;
@@ -112,7 +112,7 @@ public abstract class QueryPostprocessDataContext extends AbstractDataContext im
         if (singleFromItem && noGrouping) {
 
             final FromItem fromItem = query.getFromClause().getItem(0);
-            final Table table = fromItem.getTable();
+            final Table table = MetaModelHelper.resolveTable(fromItem);
             if (table != null) {
 
                 // check for SELECT COUNT(*) queries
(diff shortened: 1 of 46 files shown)


kaspersorensen:
Any reactions on this approach @tomatophantastico or @LosD? Following our previous discussion ... this should be the way, right?

LosD:
Can't look too much now (on the way to work), but looks good. A bit worried about the size of something that to me is pretty insignificant, though.

kaspersorensen:
I also thought the change would be a bit easier. But we've never had anything like it before so nothing was really prepared for it.

I think actually this is a rather significant feature. The inconvenience of not being able to easily formulate a simple "SELECT * FROM table" on a CSV file is really annoying. - Especially if we look to the Membrane webapp where I could totally see a use case in using this type of query. After all, when you as a HTTP client only have SQL-like strings to query with, convenience and syntax sugar becomes very important. And it makes the Postman test cases a lot more portable too ;-)

LosD:
Oh man, I never got back to this. Will take a full look tonight.

Anyway, yeah might be more significant than I thought.

core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
```diff
@@ -63,24 +64,29 @@
...
 
-    private final Map<Column, TypeConverter<?, ?>> _converters;
+    private final Map<Column, TypeConverter<?, ?>> converters;
+    private final boolean defaultTableEnabled;
 
     public QueryPostprocessDataContext() {
+        this(true);
+    }
+
+    public QueryPostprocessDataContext(boolean defaultTableEnabled) {
```
LosD: I'd maybe consider another argument and field name, like `singleTableDatastore`. The current name creates an obligation to have a default table, while `singleTableDatastore` would allow tweaking what actually happens in the situation (e.g. if we opt for context flags, like suggested in #128).

I guess a `singleTableDatastore` flag would also allow us to implement optimizations/simplifications later, as we know what to expect. We could also enforce the single table, and maybe save a context developer from surprises and headaches :).

core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
```diff
@@ -128,7 +134,7 @@ public DataSet executeQuery(final Query query) {
                                 logger.debug(
                                         "DataContext did not return any count query results. Proceeding with manual counting.");
                             } else {
-                                List<Row> data = new ArrayList<Row>(1);
+                                final List<Row> data = new ArrayList<Row>(1);
```
LosD: Unrelated change.

core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
```diff
@@ -151,8 +157,8 @@ public DataSet executeQuery(final Query query) {
                                 if (table != null) {
                                     if (isMainSchemaTable(table)) {
                                         final Object operand = whereItem.getOperand();
-                                        final Row row = executePrimaryKeyLookupQuery(table, selectItems, column,
-                                                operand);
+                                        final Row row =
+                                                executePrimaryKeyLookupQuery(table, selectItems, column, operand);
```
LosD: Unrelated change.

core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
```diff
@@ -192,8 +198,8 @@ public DataSet executeQuery(final Query query) {
 
         // we can now exclude the select items imposed by the WHERE clause (and
         // should, to make the aggregation process faster)
-        workSelectItems = CollectionUtils.concat(true, selectItems, groupBySelectItems, havingSelectItems,
-                orderBySelectItems);
+        workSelectItems =
+                CollectionUtils.concat(true, selectItems, groupBySelectItems, havingSelectItems, orderBySelectItems);
```
LosD: Unrelated change.

core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
```diff
@@ -216,8 +222,7 @@ public DataSet executeQuery(final Query query) {
     }
 
     /**
-     * Determines if all the select items are 'simple' meaning that they just
-     * represent scans of values in columns.
+     * Determines if all the select items are 'simple' meaning that they just represent scans of values in columns.
```
LosD: Unrelated change.

core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
```diff
@@ -235,38 +240,28 @@ private boolean isSimpleSelect(SelectClause clause) {
     }
 
     /**
-     * Executes a simple count query, if possible. This method is provided to
-     * allow subclasses to optimize count queries since they are quite common
-     * and often a datastore can retrieve the count using some specialized means
-     * which is much more performant than counting all records manually.
+     * Executes a simple count query, if possible. This method is provided to allow subclasses to optimize count queries
+     * since they are quite common and often a datastore can retrieve the count using some specialized means which is
+     * much more performant than counting all records manually.
```
LosD: Unrelated change.

core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
```diff
@@ -235,38 +240,28 @@ private boolean isSimpleSelect(SelectClause clause) {
...
      * 
-     * @param table
-     *            the table on which the count is requested.
-     * @param whereItems
-     *            a (sometimes empty) list of WHERE items.
-     * @param functionApproximationAllowed
-     *            whether approximation is allowed or not.
+     * @param table the table on which the count is requested.
+     * @param whereItems a (sometimes empty) list of WHERE items.
+     * @param functionApproximationAllowed whether approximation is allowed or not.
```
LosD: Unrelated change.

core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
```diff
@@ -235,38 +240,28 @@ private boolean isSimpleSelect(SelectClause clause) {
...
 
     /**
-     * Executes a query which obtains a row by primary key (as defined by
-     * {@link Column#isPrimaryKey()}). This method is provided to allow
-     * subclasses to optimize lookup queries since they are quite common and
-     * often a datastore can retrieve the row using some specialized means which
-     * is much more performant than scanning all records manually.
+     * Executes a query which obtains a row by primary key (as defined by {@link Column#isPrimaryKey()}). This method is
+     * provided to allow subclasses to optimize lookup queries since they are quite common and often a datastore can
+     * retrieve the row using some specialized means which is much more performant than scanning all records manually.
```
LosD: Unrelated change.

core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
```diff
@@ -235,38 +240,28 @@ private boolean isSimpleSelect(SelectClause clause) {
...
-     * @param selectItems
-     *            the items to select from the lookup query.
-     * @param primaryKeyColumn
-     *            the column that is the primary key
-     * @param keyValue
-     *            the primary key value that is specified in the lookup query.
+     * @param table the table on which the lookup is requested.
+     * @param selectItems the items to select from the lookup query.
+     * @param primaryKeyColumn the column that is the primary key
+     * @param keyValue the primary key value that is specified in the lookup query.
```
LosD: Unrelated change.

core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
```diff
@@ -319,13 +314,13 @@ protected DataSet materializeFromItem(final FromItem fromItem, final List<Select
 
             // materialize left side
             final List<SelectItem> leftOn = Arrays.asList(fromItem.getLeftOn());
-            fromItemDataSets[0] = materializeFromItem(fromItem.getLeftSide(),
-                    CollectionUtils.concat(true, selectItems, leftOn));
+            fromItemDataSets[0] =
+                    materializeFromItem(fromItem.getLeftSide(), CollectionUtils.concat(true, selectItems, leftOn));
```
LosD: Unrelated change.

core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
```diff
@@ -319,13 +314,13 @@ protected DataSet materializeFromItem(final FromItem fromItem, final List<Select
...
 
             // materialize right side
             final List<SelectItem> rightOn = Arrays.asList(fromItem.getRightOn());
-            fromItemDataSets[1] = materializeFromItem(fromItem.getRightSide(),
-                    CollectionUtils.concat(true, selectItems, rightOn));
+            fromItemDataSets[1] =
+                    materializeFromItem(fromItem.getRightSide(), CollectionUtils.concat(true, selectItems, rightOn));
```
LosD: Unrelated change.

core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
```diff
@@ -385,8 +380,8 @@ protected DataSet materializeTable(final Table table, final List<SelectItem> sel
 
         final DataSet dataSet;
         if (INFORMATION_SCHEMA_NAME.equals(schemaName)) {
-            DataSet informationDataSet = materializeInformationSchemaTable(table,
-                    buildWorkingSelectItems(selectItems, whereItems));
+            DataSet informationDataSet =
+                    materializeInformationSchemaTable(table, buildWorkingSelectItems(selectItems, whereItems));
```
LosD: Unrelated change.

core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
```diff
@@ -63,24 +64,29 @@
...
  */
 public abstract class QueryPostprocessDataContext extends AbstractDataContext implements HasReadTypeConverters {
 
     private static final Logger logger = LoggerFactory.getLogger(QueryPostprocessDataContext.class);
 
+    public static final String SYSTEM_PROPERTY_CREATE_DEFAULT_TABLE_ALIAS = "metamodel.alias.default.table";
     public static final String INFORMATION_SCHEMA_NAME = "information_schema";
 
-    private final Map<Column, TypeConverter<?, ?>> _converters;
+    private final Map<Column, TypeConverter<?, ?>> converters;
```
LosD: Unrelated change.

core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
```diff
@@ -63,24 +64,29 @@
...
+    private final boolean defaultTableEnabled;
 
     public QueryPostprocessDataContext() {
+        this(true);
```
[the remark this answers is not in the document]
kaspersorensen: I wanted to make sure that I carefully considered every single case of extending the base class. So to begin with I removed the default constructor so that I could check all the compile errors and make sure that every extension of QueryPostprocessDataContext was correctly setting the flag. After that I reintroduced the default constructor to have backwards compatibility.

core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
kaspersorensen: I did it for consistency sake actually. Since I introduced a new field (without the underscore, which I know you don't like) I felt like renaming the other fields in the class to also not have underscores would be appropriate.

core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
kaspersorensen: Fixed. Also added some javadoc to the constructor to explain what it means.

kaspersorensen:
Are you sure you want me to revert those "unrelated changes"? I only formatted files that I did a lot of rework of and if you ask me the diff is rather easy to read through here in GitHub. 99% of it is just line length in javadoc remarks and it highlights if any wording has changed, too. It just feels overly difficult to make grand changes to a class while not formatting it to a consistent form. No?

core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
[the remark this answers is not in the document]
LosD: Fair enough.

LosD:
In the one I looked at (QueryPostProcessessDataContext.java), the actual changes were around 10-15 lines (a little more with the since-introduced JavaDoc), out of 175 lines changed total, in a 700 lines class. A quick glance seems to repeat those numbers for many other classes.

Reformatting is fine.... For the method you're changing, even only for the lines around if it's a few lines in a larger method (of course those shouldn't exist... But they do 🙂) Anything else is just wasting the reviewers time on trying to find out what is actual changes, as well as making history traversal an even bigger pain. `?w=1` helps, but not much when lines gets broken up (as they do most of the time). The noise also makes it easier for problems to slip through, as it's _so_ easy to miss an actual change in what looks like yet another reformat.

You can't even look for highlights, as a moved part will also be highlighted (this is with `?w=1`):
image

I'll continue, but please don't do that again if you expect people to spend their spare time reviewing it.

core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
```diff
@@ -458,15 +459,18 @@ protected String getDefaultSchemaName() throws MetaModelException {
...
+        if (name == null && mainSchemaName != null) {
             return null;
         }
 
-        if (name.equalsIgnoreCase(mainSchemaName)) {
-            return getMainSchema();
+        if (name == null || name.equalsIgnoreCase(mainSchemaName)) {
+            final Schema mainSchema = getMainSchema();
+            final boolean createAliasTable = singleTableDatastore
+                    && Boolean.parseBoolean(System.getProperty(SYSTEM_PROPERTY_CREATE_DEFAULT_TABLE_ALIAS, "true"));
```
LosD: Ah, there's my flag! :)

core/src/test/java/org/apache/metamodel/QueryPostprocessDataContextTest.java
```diff
@@ -60,6 +60,49 @@
     private final Table table1 = schema.getTableByName(TABLE_CONTRIBUTOR);
     private final Table table2 = schema.getTableByName(TABLE_ROLE);
 
+    public void testSchemaTraversalWithAliasTable() {
+        final MockUpdateableDataContext dc = new MockUpdateableDataContext();
+        
+        final Column column = dc.getColumnByQualifiedLabel("foo");
+        assertEquals("table", column.getTable().getName());
+    }
+
```
LosD: I'm a bit missing a test to check that the SYSTEM_PROPERTY_CREATE_DEFAULT_TABLE_ALIAS flag actually properly disables the alias table without breaking stuff.

A bit the same with contexts that uses super(false)... Just to make sure that everything is (and keeps being) consistent.

kaspersorensen: Added those tests now :-)

kaspersorensen:
Thank you for the review. And sorry for the needless diffs. But now that things are straightened out, I'll merge ;-)

PR #167, opened 2017-11-14
METAMODEL-1169 Preserve milliseconds when rewriting query with Date operand (master)

jakubneubauer:


LosD:
Code LGTM, but this seem like a change that requires an update to the integration tests. However, they might be quite broken, so it might not make much of a difference, but at least we should take a look.

jakubneubauer:
tests were fixed.

jakubneubauer:
To the integration tests:
I found these int. tests related to date/time handling:

- SQLServerJtdsDriverTest.testWorkingWithDates and it is working.
- SQLServerJtdsDriverTest.ignoreTestTimestampValueInsertSelect is now commented out with description that this test was failing and actually somehow wrong

So it looks like it doesn't require any change.

LosD:
Merged

PR #170, opened 2017-11-27
METAMODEL-1171 fixed parsing of date/time value in queries

jakubneubauer:


kaspersorensen:
I'll merge apply this patch to `master` as it seems like a good improvement. But ideally I'd prefer it to have a unit test that included the parsing of a complete query, not just the formathelper's method.

PR #171, opened 2017-11-30
METAMODEL-1173: Fixed issue with evaluating ScalarFunctions in WHERE.

kaspersorensen:
This is a fix for METAMODEL-1173. It may seem a quite indirect fix in the sense that it uses a (new) SUBSTRING function to reproduce the issue, but from a testing perspective I needed a scalar function that was easier to work with on test data, and that took function parameters. Since we didn't have that, I found that the easiest way to test it was actually to introduce a new scalar function. It may be somewhat of a side-effect, but hopefully an agreeable/good one?


kaspersorensen:
Fixed build issue.

LosD:
Code LGTM, but not tested (at least by me, @Anand49 did you test). 

I won't have a chance to test or merge until tomorrow at the earliest, so someone else might want to do that.

kaspersorensen:
The code does include an additional test, `QueryPostprocessDataContextTest.testColumnOnlyUsedInScalarFunctionInWhereClause()`

LosD:
Oh damn, forgot about this. Will check tonight.

I meant "not tested" as in "I didn't do a manual test" :)

LosD:
Played around with it a bit in a scratch file and a Dockerized Mongo. WHEREs with map keys seems to work quite nicely (though I was quite hampered from my limited knowledge of Mongo), also without the explicit selection.

However, A SELECT with a SUBSTRING seems broken, or I'm using it wrong,

A `SELECT SUBSTRING(item, 1, 4) from inventory` with the values "journal" "mat" and "mousepad" should have yielded "jour", "mat" and "mous" (or "ourn", "at" and "ouse" if it's supposed to be 0-indexed... Which it looks like in code and tests, but seems quite odd in an SQL-like query), but instead yielded "ournal, "at" and "ousepad", seemingly ignoring the second limit. I saw something similar when trying to use it in there WHERE clause.

Here's the quite simple code snippet:
``` Java
import org.apache.metamodel.DataContext;
import org.apache.metamodel.data.DataSet;
import org.apache.metamodel.mongodb.mongo3.MongoDbDataContext;
import org.apache.metamodel.util.SimpleTableDef;

import com.mongodb.MongoClient;

class Test {
    public static void main(String[] args) {
        final DataContext mongoDbDataContext =
                new MongoDbDataContext(new MongoClient("192.168.99.100").getDatabase("test"),
                        new SimpleTableDef("inventory",
                                new String[] { "item", "qty", "tags", "size.h", "size.w", "size.uom" }));

        final DataSet ds = mongoDbDataContext.executeQuery("SELECT SUBSTRING(item, 1, 4) from inventory");
        ds.forEach(System.out::println);
    }
}
```
Values are those inserted here: https://docs.mongodb.com/manual/tutorial/insert-documents/#insert-multiple-documents


BTW, for Mongo3, the integration test is a single line from working, so maybe add a few tests around it to that one.

kaspersorensen:
Oh, I'll look at ensuring that substring actually works. 

What do you mean by "for Mongo3 the integration test is a single line from working" ?

kaspersorensen:
Hmm I "fixed" substring, at least ensured that it's returning the right results, similar to Java's `String.substring` method. But it looks like you're right with regards to the interpretation of SUBSTRING parameters in SQL. At least I looked at PostgreSQL, SQL Server and MySQL, and they all have variances of parameters, but one syntax in common: `SUBSTRING(str, start, count)` where `start` is 1 based start index and `count` is "number of characters", not end index. I guess we should comply with that?

LosD:
With "for Mongo3 the integration test is a single line from working", I meant that the MongoDB integration test aaaaaaalmost works when enabled, except for a single assert (`org.apache.metamodel.mongodb.mongo3.MongoDbDataContextTest` line 262), so it might be a good idea to keep it up to date, and add a test for the map issue, since that's one of the DBs where it matters.

Yeah, at least to me it would make most sense to use the behavior of the language we're emulating. We _could_ do it different when using the Java interface, but that would also somehow be weird, consistency-wise. Anyway, none of it is really a blocker, there's arguments for all approaches.

LosD:
Changes looks really good, but Travis is angry. Will look into it when I have a chance (might not be today)

LosD:
A bit irrelevant to this PR, but I _love_ getting integration tests into the Travis build. If we can stagger them a bit, I guess we can do all database engines that are available on Docker? Which even includes SQL Server these days.

kaspersorensen:
Darn it ... Seems we have something rotten in our handling of scalar functions. Have been trying to sort out that issue all night now, and still not fixed it. In short, we evaluate scalar functions in two different ways, (1) via `ScalarFunctionDataSet` and (2) via `FilteredItem`. I've been trying to consolidate it to _only_ be via (1), but there are still failing tests and stuff. Putting this PR on hold...

LosD:
Oh man

kaspersorensen:
There! I believe it's all fixed now. Took me some real digging around and eventually found that due to a bad `SelectItem.equals(Object)` method, the two scalar function SelectItems was reduced to one. Ended up adding a whole slew of unit tests to finally identify that this (and probably a few other tweaks) was the root cause.

If you'd like, I can change the PR to be more minimalistic. But I kind of like every little tweak that was done TBH, so this here is my suggested patch.

core/src/main/java/org/apache/metamodel/QueryPostprocessDataContext.java
```diff
@@ -406,18 +409,15 @@ protected DataSet materializeTable(final Table table, final List<SelectItem> sel
     }
 
     private List<SelectItem> buildWorkingSelectItems(List<SelectItem> selectItems, List<FilterItem> whereItems) {
-        final List<SelectItem> primarySelectItems = new ArrayList<>(selectItems.size());
-        for (SelectItem selectItem : selectItems) {
```
kaspersorensen: Just to say it ... This is one of the things that I am really pleased to get rid of. Because the `MetaModelHelper.getFiltered` and `MetaModelHelper.getSelection` methods are now taking care of scalar functions, stuff like this for-loop is no longer needed.

core/src/main/java/org/apache/metamodel/data/SimpleDataSetHeader.java
```diff
@@ -104,12 +104,6 @@ public int indexOf(SelectItem item) {
             }
             i++;
         }
-        
-        final boolean scalarFunctionQueried = item.getScalarFunction() != null;
-        if (scalarFunctionQueried) {
-            final SelectItem itemWithoutFunction = item.replaceFunction(null);
-            return indexOf(itemWithoutFunction);
```
kaspersorensen: Similarly, this special if-block for scalar functions was kind of a bad code smell that's now gone :-)

LosD:
Hasn't looked super thoroughly at it, but mostly LGTM.

A few thoughts: Not really a fan of MM helper getting bigger, when I think it should basically die in a fire, replaced with more specialised classes, but oh well. If it helps here, we can always refactor into a sensible model later if so decided.

There's a lot of changes to Mongo. Could it be that we need to do similar things to other databases that supports map-like queries?

LosD:
Okay, saw that wrong. Beyond the tests, there's actually NO changes in Mongo. Just you and your annoying formatter fingers :)

LosD:
LGTM

LosD:
Works!

LosD:
Merged

PR #172, opened 2017-12-07
METAMODEL-1173: Same as #171, but with Travis CI integration tests

kaspersorensen:
Same as #171, but with a tweak to the travis CI setup to enable running of MongoDB integration tests.
Doing this in a separate branch in case the Travis changes really doesn't work ;-)


kaspersorensen:
Ok that seemed to work - closing this in order to just have #171

PR #174, opened 2017-12-12
METAMODEL-1173: Added CouchDB integration test in add. to MongoDB

kaspersorensen:
Here you go @LosD :-) It's in deed a good idea to add test coverage for +1 database when it comes to METAMODEL-1173 :-)


LosD:
Beautiful! Merging

PR #176, opened 2018-01-09
METAMODEL-1177: Replaced null TableType with TableType.TABLE.

kaspersorensen:


kaspersorensen:
Merging this based on lazy concensus

PR #177, opened 2018-01-24
Upgrade Elasticsearch dependency from 1.5.0 to 5.6.3

arjansh:
Biggest changes:
- Node client is no longer supported. If you want to connect as a node client, you can "Start locally a Coordinating Only Node and then simply create a TransportClient in your application which connects to this Coordinating Only Node."
- Elasticsearch no longer supports the removal of Mappings from indices and existing type and field mappings can no longer be updated. As a result of this, we can no longer support drop table functionality in MetaModel.
- The io.searchbox jest client has been replaced by Elasticsearch's High level rest client. As a result of which the complete module has been rewritten and some code could be moved to the common module, because it now is the same in both the native and rest client modules.
- The EmbeddedElasticSearchServer has been removed from the test classes, because it is no longer possible to spawn a node in that manner. For the native client, it has been replaced by an Elasticsearch provided test server and client by extending the ESSingleNodeTestCase class, within the Rest client we now spawn a Docker image which runs an Elasticsearch cluster.


elasticsearch/common/src/main/java/org/apache/metamodel/elasticsearch/AbstractElasticSearchDataContext.java
```diff
@@ -0,0 +1,149 @@
...
+import org.apache.metamodel.schema.Schema;
+import org.apache.metamodel.util.SimpleTableDef;
+import org.elasticsearch.common.unit.TimeValue;
+
+public abstract class AbstractElasticSearchDataContext extends QueryPostprocessDataContext implements DataContext,
+        UpdateableDataContext {
+
+    public static final TimeValue TIMEOUT_SCROLL = TimeValue.timeValueSeconds(60);
+
+    protected final Object elasticSearchClient;
```
kaspersorensen: There doesn't seem to be any reason to keep this in the abstract class any more, since the `Object` type removes any kind of possibility of reusing it. It's used differently in the two subtypes, so I suggest they each just have a field for the client object with a type that they need.

elasticsearch/rest/pom.xml
```diff
@@ -52,40 +51,117 @@ under the License.
...
+										<ports>
+											<port>9200:9200</port>
+											<port>9300:9300</port>
+										</ports>
+
+										<env>
+												<ES_JAVA_OPTS>-Xms1g -Xmx1g</ES_JAVA_OPTS>
+												<cluster.name>docker-cluster</cluster.name>
+												<bootstrap.memory_lock>true</bootstrap.memory_lock>
+											    <xpack.security.enabled>false</xpack.security.enabled>
```
kaspersorensen: indentation issues here.

elasticsearch/rest/pom.xml
```diff
@@ -52,40 +51,117 @@ under the License.
...
+									<goal>verify</goal>
+								</goals>
+							</execution>
+						</executions>
+					</plugin>
+
+					<plugin>
+						<groupId>io.fabric8</groupId>
+						<artifactId>docker-maven-plugin</artifactId>
+						<version>0.15.16</version>
```
kaspersorensen: It's a fairly old version of this plugin that you're using here. Maybe we should (1) use the latest just because Docker things like this seems to move fast and (2) add it's version to <pluginManagement> in the root pom.

arjansh:
I suggest we do a release candidate after this PR has been merged.

PR #178, opened 2018-01-28
A new module for Apache Kafka "querying" using MetaModel

kaspersorensen:
Hi all,

I've been playing around with this idea for a while and decided to try it out ... I'm curious what you think!

This is a new MM module which allows you to query Apache Kafka. By "query" I mean: Each topic is represented as a table. And fields in here include `partition`, `offset`, `key`, `value` and `timestamp`.

When querying, the MM module creates a new unique Kafka consumer. This works, but I have one worry, which I don't really know how to mitigate: Consumers are registered in Kafka and cannot be deleted without pulling some tricks that go all the way into Zookeeper. This worries me because then every query fired will leave behind a new consumer ID in Zookeeper. If you use this module a lot then, then it becomes a lot of garbage.

Except for the issue described above, I think this works quite well. Queries are very fast and I've also added ability for the module to insert stuff.

### DIFF
--- a/kafka/src/test/java/org/apache/metamodel/kafka/KafkaDataContextTest.java
+++ b/kafka/src/test/java/org/apache/metamodel/kafka/KafkaDataContextTest.java
@@ -1,3 +1,21 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
 package org.apache.metamodel.kafka;
 
 import java.util.ArrayList;
--- a/kafka/src/test/java/org/apache/metamodel/kafka/KafkaDataContextTest.java
+++ b/kafka/src/test/java/org/apache/metamodel/kafka/KafkaDataContextTest.java
@@ -25,7 +25,7 @@ public class KafkaDataContextTest extends EasyMockSupport {
 
     @Test
     public void testGetSchemaInfo() {
-        final ConsumerFactory consumerFactory = createMock(ConsumerFactory.class);
+        final ConsumerAndProducerFactory consumerFactory = createMock(ConsumerAndProducerFactory.class);
 
         replayAll();
 
@@ -39,7 +39,7 @@ public class KafkaDataContextTest extends EasyMockSupport {
 
     @Test
     public void testQueryWithoutOptimization() {
-        final ConsumerFactory consumerFactory = createMock(ConsumerFactory.class);
+        final ConsumerAndProducerFactory consumerFactory = createMock(ConsumerAndProducerFactory.class);
         @SuppressWarnings("unchecked")
         final Consumer<String, String> consumer = createMock(Consumer.class);
 
@@ -94,7 +94,7 @@ public class KafkaDataContextTest extends EasyMockSupport {
 
     @Test
     public void testQueryOptimizationByPartition() {
-        final ConsumerFactory consumerFactory = createMock(ConsumerFactory.class);
+        final ConsumerAndProducerFactory consumerFactory = createMock(ConsumerAndProducerFactory.class);
         @SuppressWarnings("unchecked")
         final Consumer<String, String> consumer = createMock(Consumer.class);
 
--- a/kafka/src/test/java/org/apache/metamodel/kafka/KafkaDataContextTest.java
+++ b/kafka/src/test/java/org/apache/metamodel/kafka/KafkaDataContextTest.java
@@ -1,0 +1,154 @@
+package org.apache.metamodel.kafka;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Supplier;
+
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+import org.apache.kafka.common.PartitionInfo;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.metamodel.DataContext;
+import org.apache.metamodel.data.DataSet;
+import org.easymock.Capture;
+import org.easymock.EasyMock;
+import org.easymock.EasyMockSupport;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class KafkaDataContextTest extends EasyMockSupport {
+
+    @Test
+    public void testGetSchemaInfo() {
+        final ConsumerFactory consumerFactory = createMock(ConsumerFactory.class);
+
+        replayAll();
+
+        final Supplier<Collection<String>> topicSupplier = () -> Arrays.asList("foo", "bar");
+        final DataContext dc = new KafkaDataContext<>(String.class, String.class, consumerFactory, topicSupplier);
+
+        verifyAll();
+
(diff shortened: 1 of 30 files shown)


kaspersorensen:
A bit of clarification to answer my own question/worry: It seems that the `Consumer.unsubscribe(..)` call is the best thing I can reasonably do. This seems to remove any Zookeeper entry for the consumer group. I think. At least on my own system where I'm testing, I can't see any trace of my consumers in Zookeeper. But the things that I don't think we can ever get rid of are the entries/records in the consumer_offsets topic that is internal to Kafka.

I've been thinking that the best possibility would be to try and alleviate the problem by reusing existing Consumer group IDs. Implemented by using some pool of group IDs where leases for a group ID are granted when querying and returned when the DataSet is closed. But then again, that would make concurrent queries (or even worse - queries from multiple processes or nodes) very hard to manage because we would never know for sure that another MetaModel client wasn't _also_ firing a query with the same Consumer group ID. So my conclusion so far is that it isn't really improvable, except maybe by giving some hooks back to the user in case he really worries about storing too much consumer group information.

tomatophantastico:
Hi,

I am not sure, if i get the targeted use case here right, so intuitively i would hesitate to use this MM module for querying the Kafka, because: 

1) In Kafka message values are binary values (right?) and will often carry complex messages. While you can choose for simple/primitive types appropriate deserializers, for more complex (json, xml, avro, kryo, ....) this is not possible. So actually an nesting of datacontexts would be great here, though i do not think this is possible.

2) W.r.t. the subscriber issue: 
While it may work, your approach materializes all message data in main memory, for each query, each time. 
This can get very expensive, for both the Kafka server and the consumer.
If continually used, this should definitely by a subscriber / database.

kafka/src/test/java/org/apache/metamodel/kafka/KafkaDataContextTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+        replayAll();
+
+        final Supplier<Collection<String>> topicSupplier = () -> Arrays.asList("myTopic");
+        final DataContext dc = new KafkaDataContext<>(String.class, String.class, consumerFactory, topicSupplier);
+
+        final DataSet dataSet = dc.query().from("myTopic").select("partition", "offset", "value").where("key").eq(
+                "key2").execute();
+        Assert.assertTrue(dataSet.next());
+        Assert.assertEquals("Row[values=[1, 2, value2]]", dataSet.getRow().toString());
```
tomatophantastico: Please do not validate ResultSets via String serialization.
Without an explicit order by, the query processor should also be free to change ordering.

kaspersorensen: Actually the order here is given by the mock Consumer object above in the code. It’s just an imitation/mock scenario of how the records may be delivered to the client. So in this case I think the asserted order also makes sense to include because we assert that they are all there and that they’re streamed out and not reordered.

kaspersorensen:
To address your two discussion points:

1. Yes it is very typical to store some data structure in Kafka. I definitely think that a natural improvement here would be to extend the module here with more pluggable serialization support. I wonder if the already existing Converters utils would work. I think I’ll try that first.
Btw this is not very different from how our Cassandra module works today since Cassandra also exposes just the stored bytes as the base of its API.

2. Certainly some queries would not work well here. Using ORDER BY on a stream means that we have to exhaust the stream first. But OTOH many queries will work quite well and that’s what I was looking for. It does do streaming of the records out to the client, so it won’t have to materialize everything in memory to serve those queries.

kafka/src/test/java/org/apache/metamodel/kafka/KafkaDataContextTest.java
tomatophantastico: Correct me, if i am wrong: 
The apis the test should be programmed to not make any of those guarantees on both the serialization format and order.

When i modified the nested loop join i ran into a lot of those test. They failed as they assumed a specific order, so i rewrote working code in order to not change too many test.

In case of this particular test this is not a problem, as this is just a projection/selection, but in case of other operators, this can easily be  a problem, so i would avoid this kind of testing.

tomatophantastico:
wrt 1) That unfolding would imho add a lot of benefit. in the case of kafka this will be especially difficult, as virtually any kind of data can be encountered. iirc in case of MongoDB the compromise of dealing with a lack of schema is to scan a certain amount of documents per collection, so maybe a similar approach can be taken here.

wrt 2) My objection here is that kafka is build to scale, up to the TBs of data. While not the full dataset has to be held in memory at a time, all data to be filtered has to be transferred and materialized put into local memory, right? So, if there are 10^9 messages on the server, and you want to repeatedly query for a certain key, you'll have to copy all 10^9 messages every time.
Your concern  was that you might bog down kafka with too many subscriber ids. I just want to point out that this is probably a much bigger performance issue.

tomatophantastico:
wrt 2) My comment sounds a bit too negative, it should read it more like: I wouldn't worry to much about the subscriber, just make the adapters behaviour clear and continue. Having more adapters is always great.

kafka/src/test/java/org/apache/metamodel/kafka/KafkaDataContextTest.java
kaspersorensen: You're not wrong :-) Kafka does give a few guarantees - the main one being that within the same partition, the offsets will be continuously incrementing. So a bit of ordering can be guaranteed, but obviously not across multiple partitions.

In this case I want to ensure that all my records are in deed there on the output. Do you think there is a nicer way that I can do that? I'm open for doing it differently, but this seems to me to be the most readable and easy way to do it.

kaspersorensen:
wrt 1) OK this is a sure improvement point. Do you think it's required at this point, or can we use this module without it, and add this feature in a later PR. I might just go ahead and do it, but I don't feel like committing to it right now since I don't know when my time permits doing it.

wrt 2) Again I would argue that the same holds true for e.g. Cassandra. Some of the databases we support are simply not always well fit for a SQL like abstraction which MetaModel is. But OTOH there are also plenty of cases (including a project that I'm on at work) where Kafka is used not because it's massive volume, but because it has other compelling features (replication, load balancing, auto purging). In such a case I really almost don't mind if I end up scanning the complete stream to find the record that I'm looking for. It's such a rare event and not critical in sense of performance, but giving me the MetaModel module provides me with an easy abstraction to implement it.

tomatophantastico:
wrt 1) no, not required. if you got a use case, then this is fine with me.  I just think, for getting the data from the value, you still need  a lot of domain knowledge. I just wondered if there is a declarative way to make the value accessible. 

wrt 2) i disagree that this holds for cassandra, as cassandra features a rich query api (and cql) and beyond that a some sort of schema. in Kafka the message carries in most cases a binary blob and you will need to iterate over all of them. 
It is true that for all postprocessingdatacontexts (virtually all but JDBC) join and other  operations are resource intensive, as they are performed in memory.

The method you propose is fine for the infrequent case, small data set and  simple lookups. Metamodel otoh features a rich query api, so performance issues can surface quickly.
As this is the case with other systems, i suppose this is ok. Kafka just struck as it is a mix of remote data locality, potentially huge size and lack of query capabilities.

LosD:
I've been reading this a bit, and can't say that I have much technical input, as I've never really used Kafka, but it sounds a bit like something we could try as a kind of experimental module? It's often hard to judge the value of something like this unless more people has had the chance to use it and give feedback.

Of course it should come with disclaimers and urge people to give that quite precious feedback.

kaspersorensen:
I like that idea myself. I notice that many projects (such as Hadoop, Kafka, Guava, Spark) have a `@InterfaceStability` annotation. We could do the same as them and mark this one as "Evolving" or so.

Taking that idea a step further ... We might even want to consider annotating all our DataContexts with some informational annotations to let people know about the Query characteristics which we currently just [capture in the wiki](https://cwiki.apache.org/confluence/display/METAMODEL/Query+execution+strategies). Then we could certainly tag classes like this with some annotations that would explain how the SQL analogy is constrained to more or less a subset of meaningful queries.

I would definately volunteer to doing that work. But does that then mean we merge this PR, or do you think it needs to be further considered?

kaspersorensen:
It's been a while and we didn't really do anything. For now I've added a copy of the `InterfaceStability` annotations to the project and marked up `KafkaDataContext` as `Unstable`. Does this satisfy the PR enough to be included you think @losd and @tomatophantastico ?

ardlema:
Hi all, I've just seen this PR and think that this is quite interesting. I've been working with kafka for a while and will definitely give a try to this when gets merged.

So here is my +1 to go ahead with this and mark it (somehow) as experimental.

LosD:
Looks good!

kaspersorensen:
Merging it into master now then.

PR #181, opened 2018-06-05
Don't use Object.finalize()

kaspersorensen:
Since JDK9 the `Object.finalize()` method has been marked as `@Deprecated`. In this PR I removed references to `Objects`s `finalize` method, and tried to get rid of our own `finalize` method where possible without introducing potential issues.

Oh and there was a small junit dependency inconsistency.

I guess I just decided to weed out a couple of warnings really :-)


kaspersorensen:
Build fixed when I merged in master.

kaspersorensen:
Self-merging this one based on lazy concensus.

PR #182, opened 2018-06-06
Add Create, Update, Delete functionality for HBase

GerardDellemann:
Create: It's possible to create tables and insert rows.
Update: Insert and update are the same in HBase.
Delete: It's possible to drop a table and delete rows based on the rowKey.


hbase/src/main/java/org/apache/metamodel/hbase/HBaseUpdateCallback.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Initiates the building of a row insertion operation.
+     * @param table Table to get inserts.
+     * @param columns List of {@link HBaseColumn} to insert on.
+     * @return {@link HBaseRowInsertionBuilder}
+     * @throws IllegalArgumentException The table must be an {@link HBaseTable} and the columns list can't be null or empty
+     */
+    public HBaseRowInsertionBuilder insertInto(final Table table, final List<HBaseColumn> columns) {
```
kaspersorensen: I don't think it makes much sense to expose this method which is not part of the interface, but to make the `insertInto(Table)` method throw `UnsupportedOperationException`. If the user cannot use the interface, then why even implement it? The user will have to do something quite awkward like:

```
dataContext.executeUpdate(cb -> {
  HBaseUpdateCallback hbc = (HBaseUpdateCallback) cb;
  // something goes here
});
```
Normally we even keep the UpdateCallback implementations non-public because we don't want the users to know about it, just the interface.

core/src/main/java/org/apache/metamodel/insert/AbstractRowInsertionBuilder.java
```diff
@@ -44,6 +44,12 @@ public AbstractRowInsertionBuilder(U updateCallback, Table table) {
         _table = table;
     }
 
+    public AbstractRowInsertionBuilder(final U updateCallback, final Table table, final List<Column> columns) {
```
kaspersorensen: I don't think this constructor will be needed if we resolve my first discussion point. And because of that, I prefer to not add it, because not having it keeps the check in place that we don't suddenly start requiring a different method signature wherever this constructor is called.

hbase/src/main/java/org/apache/metamodel/hbase/HBaseColumn.java
```diff
@@ -0,0 +1,124 @@
...
+
+import org.apache.metamodel.schema.ColumnType;
+import org.apache.metamodel.schema.ColumnTypeImpl;
+import org.apache.metamodel.schema.MutableColumn;
+import org.apache.metamodel.schema.SuperColumnType;
+import org.apache.metamodel.schema.Table;
+
+public final class HBaseColumn extends MutableColumn {
+    public static final ColumnType DEFAULT_COLUMN_TYPE_FOR_ID_COLUMN = new ColumnTypeImpl("BYTE[]",
+            SuperColumnType.LITERAL_TYPE);
```
kaspersorensen: Should probably be `SuperColumnType.BINARY_TYPE`

kaspersorensen: Actually you may want to just use `ColumnType.BINARY` instead of creating a new ColumnTypeImpl here.

hbase/src/main/java/org/apache/metamodel/hbase/HBaseColumn.java
```diff
@@ -0,0 +1,124 @@
...
+
+    public HBaseColumn(final String columnFamily, final String qualifier, final Table table) {
+        this(columnFamily, qualifier, table, -1);
+    }
+
+    public HBaseColumn(final String columnFamily, final Table table, final int columnNumber) {
+        this(columnFamily, null, table, columnNumber);
+    }
+
+    public HBaseColumn(final String columnFamily, final String qualifier, final Table table, final int columnNumber) {
```
kaspersorensen: This is a lot of constructors :-) I would only add the ones you use yourself, and then make the class non-public and final.

hbase/src/main/java/org/apache/metamodel/hbase/HBaseCreateTableBuilder.java
```diff
@@ -0,0 +1,98 @@
...
+        super(updateCallback, schema, name);
+        if (!(schema instanceof MutableSchema)) {
+            throw new IllegalArgumentException("Not a mutable schema: " + schema);
+        }
+        this._columnFamilies = columnFamilies;
+    }
+
+    @Override
+    public Table execute() {
+        if (_columnFamilies == null || _columnFamilies.isEmpty()) {
```
kaspersorensen: Instead of requiring column families to be called out in a way that's not part of MetaModel's public API, could we do something smart with e.g. column names that represent the column families? For example, we could use a dot-separated column name approach, where "family1.foo" and "family1.bar" represented two column in the same column family. I think we would need this in order to retain compatibility with the public API. I'm not a big fan of introducing HBase-specific methods in `UpdateCallback` since we've then kinda gone against the idea of using MetaModel in the first place (a uniform API for database access).

hbase/src/main/java/org/apache/metamodel/hbase/HBaseUpdateCallback.java
arjansh: This part has been refactored. You can now use the insertInto(Table) method as exposed by the api and add the columns and values to the returned RowInsertionBuilder.

core/src/main/java/org/apache/metamodel/insert/AbstractRowInsertionBuilder.java
arjansh: This constructor has been removed.

hbase/src/main/java/org/apache/metamodel/hbase/HBaseColumn.java
arjansh: I changed it to ColumnType.BINARY.

hbase/src/main/java/org/apache/metamodel/hbase/HBaseColumn.java
arjansh: I removed the unused constructors. For now I left the class public however, because it represents the manner in which an HBase column works. I think it is possible to use a Column, but then you need some logic to deduce the actual HBase column from the normal Column, using a colon to split the Column's name on and get an HBaseColumn based on that.

hbase/src/main/java/org/apache/metamodel/hbase/HBaseCreateTableBuilder.java
arjansh: When you create an HBase table, you specify the Column Families for the tables. After the table has been created the Column Families are fixed and can't be changed. Within each column family you have the freedom to write what you want, so qualifiers aren't part of the creation of the table, the are created on the fly when writing data to an HBase table. Therefore we only mention column families and not qualifiers or a combination of both when creating the Table.

arjansh: An option might be to drop the setColumnFamilies method and instead use the withColumn method to set the column families for this CreateTableBuilder, but then we have to make clear that the column added by the withColumn method represents a column family for the HBase perspective.

kaspersorensen: I see that we have `DEFAULT_COLUMN_TYPE_FOR_COLUMN_FAMILIES`. So maybe we could provide this way of creating column/column families:

```
Callback cb = ...
cb.createTable(schema, "My_table")
  .withColumn("family1").ofType(DEFAULT_COLUMN_TYPE_FOR_COLUMN_FAMILIES)
  .withColumn("family2.foo").ofType(STRING)
  .withColumn("family2.bar").ofType(INTEGER)
```

What that would amount to is two column families created: `family1` and `family2`. The schema for `family1` would not be defined, so it would just be represented as a single MM Column with type LIST, but the schema for `family2` will have been defined to contain `foo` and `bar`. Just like how we can "define" it via a `SimpleTableDef`.

hbase/src/main/java/org/apache/metamodel/hbase/HBaseClient.java
```diff
@@ -0,0 +1,197 @@
...
+                if (rowExists(table, rowKeyAsByteArray)) {
+                    table.delete(new Delete(rowKeyAsByteArray));
+                } else {
+                    logger.warn("Rowkey with value {} doesn't exist in the table", rowKey.toString());
+                }
+            } catch (IOException e) {
+                throw new MetaModelException(e);
+            }
+        } else {
+            logger.info("Have not deleted a row, which has an empty (\"\") rowKey.");
```
kaspersorensen: Should we maybe throw here instead? Or at the least log at error level. But I think throwing makes most sense.

hbase/src/main/java/org/apache/metamodel/hbase/HBaseDataContext.java
```diff
@@ -242,17 +240,18 @@ protected DataSet materializeMainSchemaTable(Table table, List<Column> columns,
     }
 
     private void setMaxRows(Scan scan, int maxRows) {
-        try {
-            // in old versions of the HBase API, the 'setMaxResultSize' method
-            // is not available
-            Method method = scan.getClass().getMethod("setMaxResultSize", long.class);
-            method.invoke(scan, (long) maxRows);
```
kaspersorensen: Is using `Scan.setMaxResultSize(...)` still not the recommended approach to setting max rows? If so, we should keep this approach. I think the PageFilter is considerably slower.

hbase/src/main/java/org/apache/metamodel/hbase/HBaseFamilyMap.java
```diff
@@ -106,9 +105,9 @@ public String toString() {
             if (sb.length() > 1) {
                 sb.append(',');
             }
-            sb.append(Arrays.toString(entry.getKey()));
+            sb.append(new String(entry.getKey()));
             sb.append('=');
-            sb.append(Arrays.toString(entry.getValue()));
+            sb.append(new String(entry.getValue()));
```
kaspersorensen: Isn't this problematic, since the byte-array may not be a string, and it may not even be decodeable as a string, so this could throw.

hbase/src/main/java/org/apache/metamodel/hbase/HBaseRowInsertionBuilder.java
```diff
@@ -0,0 +1,200 @@
...
+            }
+            if (columnNameParts.length == 2) {
+                return new HBaseColumn(columnNameParts[0], columnNameParts[1], getTable());
+            }
+            throw new MetaModelException("Can't determine column family for column \"" + columnName + "\".");
+        }
+    }
+
+    @Override
+    public RowInsertionBuilder value(final int columnIndex, final Object value) {
```
kaspersorensen: When you extend the abstract class that you use here, you shouldn't need to override these methods. Is there some special reason for doing that? I worry that any new method addition will be forgotten here and would rather that we just used the abstract class'es methods.

hbase/src/main/java/org/apache/metamodel/hbase/HBaseDataContext.java
arjansh: The max result size is not related to the number of returned rows. It sets the number of bytes which is returned by a single scan before it scans for more. So if you set it to 1, a scan will only return 1 byte and if a row contains of more then 1 byte it will just return the complete row and after that do a new scan for the next row.

hbase/src/main/java/org/apache/metamodel/hbase/HBaseRowInsertionBuilder.java
arjansh: The special reason is that we don't use the `_values` and `_columns` arrays used by the super classes, because when creating the insertion builder we can only pass the Table in the constructor and the available columns on an HBase table aren't fixed, because qualifiers are dynamic can be added on the fly, only the column families are fixed.

hbase/src/main/java/org/apache/metamodel/hbase/HBaseFamilyMap.java
arjansh: The problem however is that the `Arrays.toString` variant returns the key/value combination of `hey` and `yo` as `[104, 101, 121]` and `[121, 111]`, which you can hardly call a String representation. where the `new String` will typically return `hey` and `yo` in that case. In case some unexpected character encoding is used it may display some shady characters, but it should still work.

hbase/src/main/java/org/apache/metamodel/hbase/HBaseDataContext.java
kaspersorensen: Oh, good to know :-) Then the change certainly makes sense.

hbase/src/main/java/org/apache/metamodel/hbase/HBaseFamilyMap.java
kaspersorensen: I believe that there are byte combinations that are simply invalid. So I suspect it could throw exceptions in certain cases. I'm not an expert on this however. Anyone else who has certainty?

hbase/src/main/java/org/apache/metamodel/hbase/HBaseRowInsertionBuilder.java
kaspersorensen: In that case, shouldn't we then avoid the abstract class, and instead do a full implementation of the interface?

kaspersorensen: Wait ... So how would a user be able to pass in a column that's not already defined? I see that `HBaseTable` returns `HBaseColumn`s which represent that paths down into the column families as defined by the `SimpleTableDef`s. So it's kind of dynamic based on that `SimpleTableDef`. But is there some way that the columns will change after creating the simple table def? Or rather - in the middle of a update script? I don't see how that would be possible, but maybe it is, and then I think that deserves some clarification (since it's unusual for MetaModel).

arjansh: The columns are not defined, only the column families are, qualifiers are not. So you can insert values into columns which don't exist yet. Each row in an HBase table can have different columns from other rows in that table. Effectively that makes it harder to read from an HBase table in a manner that fits MetaModel then to write to an HBase table, because MetaModel typically writes the to same columns over and over again and that's no problem for HBase.

arjansh: I'm going to look a bit into the necessity of the using the abstract class, maybe a full implementation is indeed better in this case.

hbase/src/main/java/org/apache/metamodel/hbase/HBaseFamilyMap.java
arjansh: The behavior of this String constructor when the given bytes are not valid in the default charset is unspecified. There is another String constructor which also takes a Charset as an arguments, that method always replaces malformed-input and unmappable-character sequences with the charset's default replacement string. So maybe we should use that instead.

arjansh: The Apache Hadoop Hbase library has a utility method for this `org.apache.hadoop.hbase.util.Bytes#toString(byte[])`, so I changed it to use that instead.

hbase/src/main/java/org/apache/metamodel/hbase/HBaseRowInsertionBuilder.java
arjansh: I refactored the class so it no longer extend the `AbstractRowInsertionBuilder`, but implements its logic completely by itself.

kaspersorensen:
I like what this PR does, but it seems a bit one-off to me. It's nice to see us supporting on-the-fly column definitions during insert, but the same principle could also apply for other connectors we have, such as the ones for CouchDB, MongoDB, ElasticSearch and more. So since the approach is a bit bespoke for HBase, and we don't know if there will be some consolidation later if we end up implementing something a bit more generic for these types of datastores, I think for now we should add the `@InterfaceStability.Evolving` annotation to the HBaseDataContext. What do you think about that?

hbase/src/main/java/org/apache/metamodel/hbase/HBaseClient.java
```diff
@@ -0,0 +1,207 @@
...
+                    } else {
+                        qualifier = Bytes.toBytes(new String(""));
+                    }
+                    final byte[] value = getValueAsByteArray(values[i]);
+                    // A NULL value, doesn't get inserted in HBase
+                    // TODO: Do we delete the cell (and therefore the qualifier) if the table get's updated?
+                    if (value == null) {
+                        logger.warn("The value of column '{}:{}' is null. This insertion is skipped", columns[i]
+                                .getColumnFamily()
+                                .toString(), columns[i].getQualifier().toString());
```
arjansh: It is possible that an HBaseColumn doesn't have a qualifier, in which case this will throw a NullPointerException, so either do `Bytes.toString(qualifier)`, or an inline if to check if the qualifier is null and in that case return "" for it.

kaspersorensen:
However, the Travis build seems to have failed :-)

kaspersorensen:
Oh I think that's a glitch. I've seen that test fail on rare occasions before. @GerardDellemann you can trigger a new build by making an empty commit to your branch, like this:
```
```

GerardDellemann:
Done, build is working now

kaspersorensen:
Merged.

PR #183, opened 2018-08-13
add support for Oracle "TIMESTAMP WITH (LOCAL_)TIME_ZONE" column type.

LingHao:
Oracle has "TIMESTAMP WITH LOCAL_TIME_ZONE" and TIMESTAMP WITH TIME_ZONE" column type, which  MetaModel doesn't support now.


kaspersorensen:
Seems like a clean fix to me. Only remark is that you've used tabs instead of spaces in the Oracle class.

LingHao:
Sorry for tabs, I didn't notice the Coding guidelines. Do I need to fix them and PR again?

kaspersorensen:
You can just make the change in the same branch. The PR will then auto-update.

kaspersorensen:
Looks good - merging

PR #184, opened 2018-08-17
Update Hive version

GerardDellemann:
This PR updates Hive from version 1.2.1 to 2.3.2, which is the latest version.


kaspersorensen:
This looks fine, and I'll merge it, but just for the record - it's only a `test` scoped dependency so it doesn't really matter. The only thing it is used for is if you run the hive integration tests (which I'm assuming that you're doing then).

GerardDellemann:
Ok nice, I did run the tests with the Hive IT's.

PR #185, opened 2018-08-20
METAMODEL-1194 Don't support primary keys with Hive

GerardDellemann:
Hive doesn't support primary keys. This PR makes sure, that when a column is defined as a primary key, then  this is ignored in the create table statement.

### DIFF
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcCreateTableBuilder.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcCreateTableBuilder.java
@@ -115,21 +115,23 @@ final class JdbcCreateTableBuilder extends AbstractTableCreationBuilder<JdbcUpda
                 sb.append(" NOT NULL");
             }
         }
-        boolean primaryKeyExists = false;
-        for (int i = 0; i < columns.size(); i++) {
-            if (columns.get(i).isPrimaryKey()) {
-                if (!primaryKeyExists) {
-                    sb.append(", PRIMARY KEY(");
-                    sb.append(columns.get(i).getName());
-                    primaryKeyExists = true;
-                } else {
-                    sb.append(",");
-                    sb.append(columns.get(i).getName());
+        if (queryRewriter.isPrimaryKeySupported()) {
+            boolean primaryKeyExists = false;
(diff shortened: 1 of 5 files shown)


jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcCreateTableBuilder.java
```diff
@@ -115,21 +115,23 @@ private String createSqlStatement(Table table) {
...
-            if (columns.get(i).isPrimaryKey()) {
-                if (!primaryKeyExists) {
-                    sb.append(", PRIMARY KEY(");
-                    sb.append(columns.get(i).getName());
-                    primaryKeyExists = true;
-                } else {
-                    sb.append(",");
-                    sb.append(columns.get(i).getName());
+        if (queryRewriter.isPrimaryKeySupported()) {
+            boolean primaryKeyExists = false;
```
arjansh: Not completely related to this pull request, but something that we can address in this context anyway, because the added if statement here only adds to the cognitive complexity of this method. Lines 119 through 134 are a carbon copy of lines 130 through 145 from the super class `AbstractTableCreationBuilder`. So maybe it's an option to extract that code into a separate method on the `AbstractTableCreationBuilder`, which can be invoked here too, this both removes the copy/pasted code and reduces the cognitive complexity of this method.

kaspersorensen: I don't think it's that easy unfortunately. The abstract class is in the `core` module so it does not have access to `IQueryRewriter` which is a `jdbc` module specific interface. The purpose in the abstract class is really just to have a string (SQL) representation of the operation, where as here it is similar, but that string representation will actually be used to perform the operation with the JDBC driver.

kaspersorensen:
And finally a small request: Can you please create a JIRA issue for this improvement. That makes the life of the committer easier so that we (or you) can add a referenced entry into CHANGES.md when it is merged.

GerardDellemann:
Made the JIRA Issue: https://issues.apache.org/jira/browse/METAMODEL-1194

kaspersorensen:
LGTM. What about you @arjansh ?

arjansh:
@kaspersorensen LGTM too then.

PR #188, opened 2018-09-13
Set the MutableColumn.table value in Kafka columns (two-way reference).

davkrause:
Here is a Pull Request where I am fixing the two-way reference from MutableColumn back to Table for the new Apache Kafka MetaModel project. I am trying to connect Kafka to the project DataCleaner, which use MetaModel a lot. and have built some initial code, but it right now is not working because something in there tries to reference the table using Column.getTable() and so on. This Pull Request should fix that.


kaspersorensen:
LGTM! I like what you're doing to DataCleaner, too 😮 😃

kaspersorensen:
I've created https://issues.apache.org/jira/browse/METAMODEL-1197 for this patch and will be merging in the change in a moment.

PR #189, opened 2018-09-26
Fixed several javadoc issues

kaspersorensen:
I was actually trying to prepare a release, but found that these javadoc issues made my build fail on newer versions of the JDK.


kaspersorensen:
I'm merging based on lazy concensus.

PR #190, opened 2018-10-03
Added cache for prepared statements in JdbcUpdateCallback

jakubneubauer:
Closes METAMODEL-1199


jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
```diff
@@ -903,4 +905,17 @@ public String getDatabaseProductName() {
     public String getDatabaseVersion() {
         return _databaseVersion;
     }
+
+    public static int getSystemPropertyValue(String property, int defaultValue) {
```
kaspersorensen: We should probably make this protected scope. I don't think we should expose this method outside of the JDBC module.

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcUpdateCallback.java
```diff
@@ -42,18 +43,35 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import javafx.util.Pair;
```
kaspersorensen: We should not rely on javafx.

jdbc/src/test/java/org/apache/metamodel/jdbc/ConnectionWrapper.java
```diff
@@ -0,0 +1,286 @@
...
+import java.sql.SQLWarning;
+import java.sql.SQLXML;
+import java.sql.Savepoint;
+import java.sql.Statement;
+import java.sql.Struct;
+import java.util.Map;
+import java.util.Properties;
+import java.util.concurrent.Executor;
+
+/** Simple wrapper around Connection */
```
kaspersorensen: Javadoc convention says that there should be a header and footer line, so this should become
```
/**
 * Simple wrapper around Connection
 */
```

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcUpdateCallback.java
kaspersorensen: I'd create a private type for it I guess.

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcUpdateCallback.java
```diff
@@ -42,18 +43,35 @@
...
 
     public JdbcUpdateCallback(JdbcDataContext dataContext) {
         super(dataContext);
         _updateSummaryBuilder = new UpdateSummaryBuilder();
+        int maxCacheSize = JdbcDataContext.getSystemPropertyValue(JdbcDataContext.SYSTEM_PROPERTY_PREPARED_STATEMENT_CACHE_SIZE, 10);
+        _preparedStatementCache = CacheBuilder.newBuilder()
+                .maximumSize(maxCacheSize)
+                .initialCapacity(1)
+                .removalListener((removalNotification) -> {
+                    closePreparedStatement((PreparedStatement) removalNotification.getValue());
```
kaspersorensen: I'd implement the interface properly here instead of using a lambda. Then you don't have to do this cast which originates from the compiler not being able to infer the types I guess.

jdbc/src/test/java/org/apache/metamodel/jdbc/CloseableConnectionWrapper.java
```diff
@@ -44,13 +44,12 @@
  * closed or not, without actually closing it.
  */
 @Ignore
-public class CloseableConnectionWrapper implements Connection {
+public class CloseableConnectionWrapper extends ConnectionWrapper {
```
kaspersorensen: I don't understand why this change is/was needed. Can you explain?

jakubneubauer:
Today I discovered one issue with this MR, which means backward incompatibility. When using batching, it means JdbcBatchUpdateCallback, this bigger cache is not safe. Resp. the order of executed statements can be changed. Because the commands are buffered (via Statement.addBatch) and they are then flushed for each prepared statement together.
So for example, when user runs commands in this order: A, B, A, B, they are actually executed in order A, A, B, B. Moreover, it can be also B, B, A, A, because order of closing and flushing prepared statements from the cache is undefined.
So the bigger cache is safe to use only for non-batching JdbcSimpleUpdateCallback.

kaspersorensen:
So I suggest then only adding the cache to the non-batch update callback?

kaspersorensen:
Hi @jakubneubauer are you still working on this?

PR #191, opened 2018-11-08
Using Travis CI's neo4j 'service' instead of docker container

kaspersorensen:


kaspersorensen:
I'm self-merging this since no one seems to object :-)

PR #192, opened 2018-11-08
Added more flexible ways to create DataContextFactoryRegistry.

kaspersorensen:
In the Membrane sub-project it will become relevant (for METAMODEL-1159)
to be able to make copies of the "default" registry, but with additional
tenant-specific factories, too. So this change makes it easier for
Membrane to copy a DataContextFactoryRegistry.


kaspersorensen:
I'm self-merging this since no one seems to object :-)

PR #195, opened 2018-11-23
METAMODEL-1205: Made CassandraUnit based test skip on JDK 9+

kaspersorensen:
It seems that CassandraUnit simply does not work on JDK 9+, see related comments in the code. For now I've added a JUnit `Assume` call which ensures that the test code is skipped when CassandraUnit fails to come up. Eventually, when they fix CassandraUnit, we should be able to upgrade the CassandraUnit dependency and it would presumably start working again.

**Update**: This PR now also updates versions of: Guava, Hadoop, HBase, Jackson and SLF4j. I got into a situation with Cassandra Unit and Cassandra itself requiring a newer Guava, which then required an upgrade of Hadoop, which then required an update of HBase. Hurray for JAR hell. Luckily it seems to work just fine when everything is on latest and greatest versions. And as a developer I quite prefer that anyway. Should set us on a better path to support JDK 9, 10 and 11.

### DIFF
--- a/cassandra/pom.xml
+++ b/cassandra/pom.xml
@@ -20,8 +20,8 @@
 	<name>MetaModel module for Apache Cassandra database</name>
 
 	<properties>
-		<cassandra.driver.latest.version>3.0.2</cassandra.driver.latest.version>
-		<cassandraunit.latest.version>2.2.2.1</cassandraunit.latest.version>
+		<cassandra.driver.latest.version>3.6.0</cassandra.driver.latest.version>
+		<cassandraunit.latest.version>3.5.0.1</cassandraunit.latest.version>
 	</properties>
 
 	<dependencies>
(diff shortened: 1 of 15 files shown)


kaspersorensen:
D'oh! Seems cassandra unit now _requires_ the use of logback classic. That's kinda lame, but I'll switch it out for the test scope at least.

kaspersorensen:
Note the "Update" in my PR description above. A lot more dependencies have been updated in this MR now to resolve JAR hell conflicts..

LosD:
Code LGTM (I could comment on the files with only cosmetic changes, but I wont... -ish :)).

Not sure what I can test. I never used Cassandra, and I don't (currently) have Java 9+ or anything Hadoop. But I `Assume.assumeTrue` that the test skipping has already been tested. ;)

If anything specifically you want me to check (also if it means playing around with Hadoop, Cassandra or Java 9+), let me know.

kaspersorensen:
I don't think we need additional testing here - the CassandraUnit tests still run on Travis CI (just not in JDK9 builds)

PR #196, opened 2018-11-23
METAMODEL-1205: Updated SugarCRM JAX-WS/WSDL codegen to work in JDK9+

kaspersorensen:


kaspersorensen:
Damnit. Seems not to work in Travis CI (I think that's because it's running JDK 8)... We need a solution that works in BOTH version 8 and above.

Maybe something with different Maven profiles then?

LosD:
This is quite 'ewwww', but I guess there's no way around it, so LGTM.

PR #197, opened 2018-11-27
METAMODEL-1205: Fixed issue with overlapping woodstox JAR files

kaspersorensen:


kaspersorensen:
Self-merging this because I accidentally broke the `master` build before when merging those multiple METAMODEL-1205 PRs.

PR #198, opened 2018-11-28
Elasticsearch native build failure

arjansh:
Running unit test in `MetaModel-elasticsearch-native` module in reverse alphabetical order to make sure that the `ElasticSearchDataContextTest` unit test class is executed last, because it will fail the build if it is executed before other unit tests in the module. 

This is caused by a mismatch between the Apache maven surefire-plugin and the `org.elasticsearch.test.ESSingleNodeTestCase` class provided by the Elasticsearch test framework, which seems to set a custom SecurityManager on the JVM when it runs.


kaspersorensen:
Oh ... Yuck ... I'm OK with fixing it this way I guess, but isn't there another possible way, such as making the classes have a shared initializer or something?

kaspersorensen:
LGTM! Merging

PR #205, opened 2019-01-04
close ResultSet after get result

q977734161:
Call ResultSet.close() method when after getting result from ResultSet, otherwise it will cause memory overflow.(for example use impala
 jdbc)

https://issues.apache.org/jira/projects/METAMODEL/issues/METAMODEL-1208

### DIFF
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
@@ -289,7 +289,7 @@ public class JdbcDataContext extends AbstractDataContext implements UpdateableDa
         } catch (SQLException e) {
             throw JdbcUtils.wrapException(e, "retrieve schema and catalog metadata", JdbcActionType.METADATA);
         } finally {
-            close(null);
+            FileHelper.safeClose(rs);
         }
         return result;
     }
@@ -580,6 +580,7 @@ public class JdbcDataContext extends AbstractDataContext implements UpdateableDa
         } catch (SQLException e) {
             logger.error("Error retrieving catalog metadata", e);
         } finally {
+            FileHelper.safeClose(rs);
             close(connection);
             logger.debug("Retrieved {} catalogs", catalogs.size());
         }
@@ -752,9 +753,14 @@ public class JdbcDataContext extends AbstractDataContext implements UpdateableDa
         // Distinct schema names. metaData.getTables() is a denormalized
         // resultset
         Set<String> schemas = new HashSet<>();
-        ResultSet rs = metaData.getTables(_catalogName, null, null, JdbcUtils.getTableTypesAsStrings(_tableTypes));
-        while (rs.next()) {
-            schemas.add(rs.getString("TABLE_SCHEM"));
+        ResultSet rs = null;
+        try {
+            rs = metaData.getTables(_catalogName, null, null, JdbcUtils.getTableTypesAsStrings(_tableTypes));
+            while (rs.next()) {
+                schemas.add(rs.getString("TABLE_SCHEM"));
+            }
+        }finally {
+            FileHelper.safeClose(rs);
         }
         return schemas;
     }
@@ -791,13 +797,17 @@ public class JdbcDataContext extends AbstractDataContext implements UpdateableDa
                     result.add(name);
                 }
             } else {
-                ResultSet rs = metaData.getSchemas();
-                while (rs.next()) {
-                    String schemaName = rs.getString(1);
-                    logger.debug("Found schemaName: {}", schemaName);
-                    result.add(schemaName);
+                ResultSet rs = null;
+                try {
+                    rs = metaData.getSchemas();
+                    while (rs.next()) {
+                        String schemaName = rs.getString(1);
+                        logger.debug("Found schemaName: {}", schemaName);
+                        result.add(schemaName);
+                    }
+                }finally {
+                    FileHelper.safeClose(rs);
                 }
-                rs.close();
             }
 
             if (DATABASE_PRODUCT_MYSQL.equals(_databaseProductName)) {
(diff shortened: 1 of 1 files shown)


q977734161:
https://patch-diff.githubusercontent.com/raw/apache/metamodel/pull/205.patch

LosD:
Error might be unrelated. All runs are good, except on OracleJDK8, where an Excel test fails inside Apache POI.

It could easily be the Excel context that has issues. We recently changed it, and it has had some timing problems previously. Of course, the closing of the result set could change the timing slightly, exposing a problem.

kaspersorensen:
I'll have Travis rebuild it to see what happens ...

kaspersorensen:
Hmm. I overspoke - I cannot do that. @q977734161 you'll probably have to make an empty commit to trigger rebuild:

```
```

Related note though - it seems that even the master build failed on one of the 4 JVMs that it's building on. I think we have an unstable test that we should look at having fixed.

q977734161:
I try to make an empty commit to trigger rebuild:
`git commit --allow-empty -m "Empty commit to trigger rebuild"`
It seems build successfully.

LosD:
D'oh, I missed that it was in JDBC, then the Excel issue is unrelated for sure. :-)

In theory, we shouldn't need to close the resultset, as resultsets are supposed to close along with the statements, but drivers can probably easily break that convention, so it's a good thing to ensure the closure in our end.

q977734161:
I use mysql, It's normal,But use Impala jdbc,cause heap overheap.

When using ```jmap -histo pid``` command,there are many impala metaset classes,when add close method is normal.

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
```diff
@@ -289,6 +289,13 @@ private boolean usesCatalogsAsSchemas(DatabaseMetaData metaData) {
         } catch (SQLException e) {
             throw JdbcUtils.wrapException(e, "retrieve schema and catalog metadata", JdbcActionType.METADATA);
         } finally {
+            if(rs != null) {
```
kaspersorensen: Very minor, but our code style conventions are to have a space after `if` and before the `(`

kaspersorensen: Actually, even better would be for you to just use `FileHelper.safeClose(rs);` instead of this if + try-catch stuff.

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
```diff
@@ -580,6 +587,13 @@ public void close(Connection connection, ResultSet rs, Statement st) {
         } catch (SQLException e) {
             logger.error("Error retrieving catalog metadata", e);
         } finally {
+            if(rs != null){
+                try {
+                    rs.close();
```
kaspersorensen: Use `FileHelper.safeClose(rs);`

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
```diff
@@ -756,6 +770,9 @@ private String findDefaultSchema(final String defaultName, final List<String> sc
         while (rs.next()) {
             schemas.add(rs.getString("TABLE_SCHEM"));
         }
+        if(rs != null){
+            rs.close();
```
kaspersorensen: Use `FileHelper.safeClose(rs);`

q977734161:
@kaspersorensen Thanks for review,I resubmit the code https://github.com/apache/metamodel/pull/205/commits/a69f2d5a9521044f6067c5f571f33dcbc98bad70 ,but build fail.so I try to make an empty commit to trigger rebuild again:
It seems build successfully.

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
```diff
@@ -756,6 +758,7 @@ private String findDefaultSchema(final String defaultName, final List<String> sc
         while (rs.next()) {
             schemas.add(rs.getString("TABLE_SCHEM"));
         }
+        FileHelper.safeClose(rs);
```
kaspersorensen: This should be in a finally block that you create once you have the ResultSet:

```
ResultSet rs = ...
try {
  ...
} finally {
  FileHelper.safeClose(rs);
}
```

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
```diff
@@ -797,7 +800,7 @@ public String getIdentifierQuoteString() {
                     logger.debug("Found schemaName: {}", schemaName);
                     result.add(schemaName);
                 }
-                rs.close();
+                FileHelper.safeClose(rs);
```
kaspersorensen: This should also be in a finally block.

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
```diff
@@ -289,6 +289,7 @@ private boolean usesCatalogsAsSchemas(DatabaseMetaData metaData) {
         } catch (SQLException e) {
             throw JdbcUtils.wrapException(e, "retrieve schema and catalog metadata", JdbcActionType.METADATA);
         } finally {
+            FileHelper.safeClose(rs);
             close(null);
```
kaspersorensen: Hmm I think we should remove this line. Seems like it's not needed since `close(null)` will just do nothing. Maybe it's a left-over line from some previous version of the file where the result set got closed properly or something :-)

kaspersorensen:
Thanks a lot @q977734161 ... And sorry, I just discovered a few more changes that I'd like to request :-)

q977734161:
@kaspersorensen , I modify the code again, Thanks for review the code .

kaspersorensen:
LGTM!

q977734161:
Ok,Thanks @kaspersorensen for viewing the code.

PR #207, opened 2019-01-17
Issue: METAMODEL-156

RatishBansal:
JdbcDataContext.getDefaultSchemaName return SchemaName error

### DIFF
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
@@ -24,12 +24,7 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Set;
-import java.util.StringTokenizer;
+import java.util.*;
 
 import javax.sql.DataSource;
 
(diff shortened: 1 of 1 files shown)


kaspersorensen:
The code involved here is a bit complicated in the first place I guess, because it's located in the generic `JdbcDataContext` class which applies to all database types. So from a review angle I'm really trying to ensure we don't break something that was working before for some database other than yours. I came to think that maybe a safer way to ensure that is if you sort the schema names by length (longest schema name first). Then you can keep the old `indexOf` based comparison and avoid something subtly changing here. What do you think?

RatishBansal:
I think that again won't solve the problem.Suppose there are 3 schemas test1, test11 and test; and the schema mentioned in jdbc connection url is test.Our getDefaultSchema() method should return the `test` but if go by logic of sorting based on length test11 will be returned which is not the expected behavior.

kaspersorensen:
But in that case your sorted list would be:

* Test11
* Test1
* Test

And of these only `test` would be found inside the URL.

RatishBansal:
Yes our sorted list will 

- Test11
- Test1
- Test

and schema mentioned in the url will be `test'. But when we will loop through the sorted list we will first encounter `Test11` which will make condition `if (lastToken.indexOf(schemaName) != -1)` true and it will return the `Test11` as output. But we need `Test` as output.So I think we cannot use `indexOf`

kaspersorensen:
No because `lastToken` will be "test" in that example. And `"test".indexOf("test1") != -1` will return `false`.

RatishBansal:
Oh..My mistake , sorry for the confusion.Yes logic of sorting based on schema length looks good.I will make the change.Thanks

kaspersorensen:
👍

RatishBansal:
Not sure why `XmlSaxDataContextTest.testParentTagIndex` test is failing .It's passing in my local, also I don't think it has anything to do with `JdcDataContext` changes

kaspersorensen:
Agreed. Can you add an empty commit to trigger rebuild:

```
```

RatishBansal:
Build has passed successful. Looks like an previous was an intermittent test failure .

kaspersorensen:
Merged, thank you for the contribution!

PR #208, opened 2019-01-23
 the result of column.getNativeType() may be null and add pagination function to the Hive sql.

q977734161:
the result of column.getNativeType() may be null

### DIFF
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
@@ -289,7 +289,7 @@ public class JdbcDataContext extends AbstractDataContext implements UpdateableDa
         } catch (SQLException e) {
             throw JdbcUtils.wrapException(e, "retrieve schema and catalog metadata", JdbcActionType.METADATA);
         } finally {
-            close(null);
+            FileHelper.safeClose(rs);
         }
         return result;
     }
@@ -580,6 +580,7 @@ public class JdbcDataContext extends AbstractDataContext implements UpdateableDa
         } catch (SQLException e) {
             logger.error("Error retrieving catalog metadata", e);
         } finally {
+            FileHelper.safeClose(rs);
             close(connection);
             logger.debug("Retrieved {} catalogs", catalogs.size());
         }
@@ -752,9 +753,14 @@ public class JdbcDataContext extends AbstractDataContext implements UpdateableDa
         // Distinct schema names. metaData.getTables() is a denormalized
         // resultset
         Set<String> schemas = new HashSet<>();
-        ResultSet rs = metaData.getTables(_catalogName, null, null, JdbcUtils.getTableTypesAsStrings(_tableTypes));
-        while (rs.next()) {
-            schemas.add(rs.getString("TABLE_SCHEM"));
+        ResultSet rs = null;
+        try {
+            rs = metaData.getTables(_catalogName, null, null, JdbcUtils.getTableTypesAsStrings(_tableTypes));
+            while (rs.next()) {
+                schemas.add(rs.getString("TABLE_SCHEM"));
+            }
+        }finally {
+            FileHelper.safeClose(rs);
         }
         return schemas;
     }
@@ -791,13 +797,17 @@ public class JdbcDataContext extends AbstractDataContext implements UpdateableDa
                     result.add(name);
                 }
             } else {
-                ResultSet rs = metaData.getSchemas();
-                while (rs.next()) {
-                    String schemaName = rs.getString(1);
-                    logger.debug("Found schemaName: {}", schemaName);
-                    result.add(schemaName);
+                ResultSet rs = null;
+                try {
+                    rs = metaData.getSchemas();
+                    while (rs.next()) {
+                        String schemaName = rs.getString(1);
+                        logger.debug("Found schemaName: {}", schemaName);
+                        result.add(schemaName);
+                    }
+                }finally {
+                    FileHelper.safeClose(rs);
                 }
-                rs.close();
             }
 
             if (DATABASE_PRODUCT_MYSQL.equals(_databaseProductName)) {
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
@@ -289,6 +289,7 @@ public class JdbcDataContext extends AbstractDataContext implements UpdateableDa
         } catch (SQLException e) {
             throw JdbcUtils.wrapException(e, "retrieve schema and catalog metadata", JdbcActionType.METADATA);
         } finally {
+            FileHelper.safeClose(rs);
             close(null);
         }
         return result;
@@ -580,6 +581,7 @@ public class JdbcDataContext extends AbstractDataContext implements UpdateableDa
         } catch (SQLException e) {
             logger.error("Error retrieving catalog metadata", e);
         } finally {
+            FileHelper.safeClose(rs);
             close(connection);
             logger.debug("Retrieved {} catalogs", catalogs.size());
         }
@@ -756,6 +758,7 @@ public class JdbcDataContext extends AbstractDataContext implements UpdateableDa
         while (rs.next()) {
             schemas.add(rs.getString("TABLE_SCHEM"));
         }
+        FileHelper.safeClose(rs);
         return schemas;
     }
 
@@ -797,7 +800,7 @@ public class JdbcDataContext extends AbstractDataContext implements UpdateableDa
                     logger.debug("Found schemaName: {}", schemaName);
                     result.add(schemaName);
                 }
-                rs.close();
+                FileHelper.safeClose(rs);
             }
 
             if (DATABASE_PRODUCT_MYSQL.equals(_databaseProductName)) {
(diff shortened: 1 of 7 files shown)


q977734161:
I add pagination function to the Hive sql,could you help me review the code?

kaspersorensen:
I'd prefer the pagination stuff in a separate MR just so that we can accept/review it separately too. But for now this is OK. I'll merge your initial code change into master now.

kaspersorensen:
As for the Hive query rewriting logic, I suggest some fundamental edits:

 * I don't like the brittleness of the version checking code. This code will break once there is a new version 4 of Hive, or something like that. So instead of throwing exceptions I think it should use the logic that we consider safest.
 * I think it would be nicer to split out the logic into subclasses. We could maybe have a `Hive2QueryRewriter` and a `Hive3QueryRewriter`.
 * Most of the functionality of `rewriteQuery` is already existing in super classes. So hopefully we can built these two new subclasses in a way where much less is being overwritten. I believe that the complete OFFSET/LIMIT stuff is already covered in `LimitOffsetQueryRewriter`.

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
```diff
@@ -231,7 +232,27 @@ private JdbcDataContext(DataSource dataSource, Connection connection, TableType[
...
+            int majorVersion = 0;
+            String[] parts = databaseVersion.split("\\.");
+            if(parts.length < 2) {
+                logger.warn("Illegal Hive Version: " + databaseVersion + " (expected A.B.* format)");
+                parts[0] = "0";
+            } else {
+                majorVersion = Integer.valueOf(parts[0]);
+            }
+            switch (majorVersion){
+                case 1:
```
kaspersorensen: I suggest moving this down before the `default` case, then you don't need to repeat the case's code block

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
```diff
@@ -231,7 +232,27 @@ private JdbcDataContext(DataSource dataSource, Connection connection, TableType[
         } else if (DATABASE_PRODUCT_H2.equals(_databaseProductName)) {
             setQueryRewriter(new H2QueryRewriter(this));
         } else if (DATABASE_PRODUCT_HIVE.equals(_databaseProductName)) {
-            setQueryRewriter(new HiveQueryRewriter(this));
+            int majorVersion = 0;
+            String[] parts = databaseVersion.split("\\.");
+            if(parts.length < 2) {
+                logger.warn("Illegal Hive Version: " + databaseVersion + " (expected A.B.* format)");
+                parts[0] = "0";
```
kaspersorensen: This doesn't do anything since `parts` is unused from here on out

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
```diff
@@ -231,7 +232,27 @@ private JdbcDataContext(DataSource dataSource, Connection connection, TableType[
         } else if (DATABASE_PRODUCT_H2.equals(_databaseProductName)) {
             setQueryRewriter(new H2QueryRewriter(this));
         } else if (DATABASE_PRODUCT_HIVE.equals(_databaseProductName)) {
-            setQueryRewriter(new HiveQueryRewriter(this));
+            int majorVersion = 0;
```
kaspersorensen: I'd initialize this with `-1` to indicate exceptional value.

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/Hive2QueryRewriter.java
```diff
@@ -0,0 +1,67 @@
...
+        return super.rewriteColumnType(columnType, columnSize);
+    }
+    
+    @Override
+    public boolean isTransactional() {
+        return false;
+    }
+    
+	@Override
+	public boolean isPrimaryKeySupported() {
```
kaspersorensen: Indentation is wrong here.

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/Hive2QueryRewriter.java
```diff
@@ -0,0 +1,67 @@
...
+        super(dataContext);
+    }
+
+    @Override
+    public String rewriteColumnType(ColumnType columnType, Integer columnSize) {
+        if (columnType == ColumnType.INTEGER) {
+            return "INT";
+        }
+
+        if(columnType == ColumnType.STRING) {
```
kaspersorensen: Missing space after `if`

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/HiveQueryRewriter.java
```diff
@@ -30,6 +35,59 @@ public HiveQueryRewriter(JdbcDataContext dataContext) {
...
+        outerQuery.from(subQuerySelectItem);
+
+        final List<SelectItem> innerSelectItems = innerQuery.getSelectClause().getItems();
+        for (SelectItem selectItem : innerSelectItems) {
+            outerQuery.select(new SelectItem(selectItem, subQuerySelectItem));
+        }
+
+
+        final String rewrittenOrderByClause = rewriteOrderByClause(innerQuery, innerQuery.getOrderByClause());
+        final String rowOver = "ROW_NUMBER() " + "OVER(" + rewrittenOrderByClause + ")";
```
kaspersorensen: No need for the `" + "` part here.

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/HiveQueryRewriter.java
```diff
@@ -30,6 +35,59 @@ public HiveQueryRewriter(JdbcDataContext dataContext) {
...
+        innerQuery.select(new SelectItem(rowOver, "metamodel_row_number"));
+        innerQuery.getOrderByClause().removeItems();
+
+        final String baseQueryString = rewriteQuery(outerQuery);
+
+        if (maxRows == null) {
+            return baseQueryString + " WHERE metamodel_row_number > " + (firstRow - 1);
+        }
+
+        return baseQueryString + " WHERE metamodel_row_number BETWEEN " + firstRow + " AND "
```
kaspersorensen: I notice now how most of this code is duplicated with `DB2QueryRewriter`. Please make sure that the code is shared via a common super-class or a utility function or something like that.

jdbc/src/test/java/org/apache/metamodel/dialects/HiveQueryRewriterTest.java
```diff
@@ -0,0 +1,87 @@
...
+        Hive2QueryRewriter qr = new Hive2QueryRewriter(mockContext);
+
+        MutableColumn col1 = new MutableColumn("kkbh");
+        MutableColumn col2 = new MutableColumn("kkmc");
+        Query q = new Query().from(new MutableTable("5_t_kk_kkxx")).select(col1).select(col2)
+                .where(col1, OperatorType.EQUALS_TO, "5207281832").orderBy(col1).setFirstRow(5).setMaxRows(9);
+        String sql = qr.rewriteQuery(q);
+        assertEquals(sql,"SELECT kkbh, kkmc FROM 5_t_kk_kkxx WHERE kkbh = '5207281832' ORDER BY kkbh ASC LIMIT 9 OFFSET 4");
+    }
+}
```
kaspersorensen: Missing line end at the end of file

jdbc/src/test/java/org/apache/metamodel/dialects/HiveQueryRewriterTest.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.metamodel.schema.MutableColumn;
+import org.apache.metamodel.schema.MutableSchema;
+import org.apache.metamodel.schema.MutableTable;
+import org.apache.metamodel.util.TimeComparator;
+import org.easymock.EasyMock;
+import org.junit.Assert;
+
+public class HiveQueryRewriterTest extends TestCase {
+
+
```
kaspersorensen: No need for the double line breaks

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/RowNumberQueryRewriter.java
```diff
@@ -0,0 +1,68 @@
...
+
+        final Query outerQuery = new Query();
+        final FromItem subQuerySelectItem = new FromItem(innerQuery).setAlias("metamodel_subquery");
+        outerQuery.from(subQuerySelectItem);
+
+        final List<SelectItem> innerSelectItems = innerQuery.getSelectClause().getItems();
+        for (SelectItem selectItem : innerSelectItems) {
+            outerQuery.select(new SelectItem(selectItem, subQuerySelectItem));
+        }
+
```
kaspersorensen: No need for the double line breaks

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/HiveQueryRewriter.java
```diff
@@ -18,18 +18,46 @@
...
+            // We prefer to use the "LIMIT n" approach, if
+            // firstRow is not specified.
+            return super.rewriteQuery(query) + " LIMIT " + maxRows;
+        } else {
+            return getRowNumberSql(query, maxRows, firstRow);
+        }
+
+    }
+
+
```
kaspersorensen: No need for the 4 line breaks

PR #209, opened 2019-02-16
Added the DynamoDB module to the `full` module. And created DataContextFactory for it.

kaspersorensen:


dynamodb/src/main/resources/META-INF/services/org.apache.metamodel.factory.DataContextFactory
```diff
@@ -0,0 +1 @@
+org.apache.metamodel.dynamodb.DynamoDbDataContextFactory
```
LosD: Missing EOL at end of file (ignore if it's needed like this by the serviceloader, but I don't believe it is)

dynamodb/src/main/java/org/apache/metamodel/dynamodb/DynamoDbDataContextFactory.java
```diff
@@ -0,0 +1,64 @@
...
+    protected String getType() {
+        return "dynamodb";
+    }
+
+    @Override
+    public DataContext create(DataContextProperties properties, ResourceFactoryRegistry resourceFactoryRegistry) {
+        final AmazonDynamoDBClientBuilder clientBuilder =
+                AmazonDynamoDBClientBuilder.standard().withCredentials(getCredentials(properties));
+        final Object region = properties.toMap().get("region");
+        if (region != null && region instanceof String) {
```
LosD: Unneeded null check

dynamodb/src/main/java/org/apache/metamodel/dynamodb/DynamoDbDataContextFactory.java
```diff
@@ -0,0 +1,64 @@
...
+            clientBuilder.setRegion(region.toString());
```
LosD: Unneeded `toString()`

LosD:
Error seems unrelated (apparently more Excel instability)

dynamodb/src/main/java/org/apache/metamodel/dynamodb/DynamoDbDataContextFactory.java
kaspersorensen: Changed it to a cast then :-)

dynamodb/src/main/resources/META-INF/services/org.apache.metamodel.factory.DataContextFactory
kaspersorensen: Added EOL

dynamodb/src/main/java/org/apache/metamodel/dynamodb/DynamoDbDataContextFactory.java
LosD: The cast is unneeded as well, since you're doing the instanceof, the compiler already knows that its a string, so it is implicitly casted to a string.

LosD: Shit no, scratch that. I am dreaming of things working in Java as they are in TypeScript :D

PR #210, opened 2019-02-26
Apache Impala support

q977734161:
Apache Impala support.Because this is no public repo for Impala jdbc,so I do not add Impala dependency in pom.xml

### DIFF
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/ImpalaQueryRewriter.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/ImpalaQueryRewriter.java
@@ -22,6 +22,10 @@ import org.apache.metamodel.jdbc.JdbcDataContext;
 
 /**
  * Query rewriter for Apache Impala
+ *
+ * Because Impala uses the same metadata store as Hive to record information about table structure and properties，
+ * so Impala provides a high degree of compatibility with the Hive Query Language (HiveQL) ;
+ * @see <a href="http://impala.apache.org/docs/build/html/topics/impala_langref.html">impala.apache.org</a>
  **/
 public class ImpalaQueryRewriter extends Hive2QueryRewriter{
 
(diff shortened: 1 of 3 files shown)


jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/ImpalaQueryRewriter.java
```diff
@@ -0,0 +1,32 @@
...
+ * under the License.
+ */
+package org.apache.metamodel.jdbc.dialects;
+
+import org.apache.metamodel.jdbc.JdbcDataContext;
+
+/**
+ * Query rewriter for Apache Impala
+ **/
+public class ImpalaQueryRewriter extends Hive2QueryRewriter{
```
kaspersorensen: Is Impala's SQL dialect so closely related to Hive's SQL dialect that it should inherit it directly? Or is it that they should have a common super class? I think that when we choose inheritance it's always important to ask/verify that "X is a type of Y", not just that they share attributes or things. Like "A cat is an animal" because it _IS_ an animal, but say the class Dog defined "four legs", I would still advise against doing a "Cat extends Dog" relationship.

q977734161: Because Impala uses the same metadata store as Hive to record information about table structure and properties，so Impala provides a high degree of compatibility with the Hive Query Language (HiveQL) ;
[(http://impala.apache.org/docs/build/html/topics/impala_langref.html)]

kaspersorensen: OK that makes sense to me then. But can you then please add that to the class's javadoc comments so that this relationship is clear to anyone reading the code.

q977734161: Ok,I will do that.

kaspersorensen:
LGTM. Merging :-)

PR #211, opened 2019-03-06
METAMODEL-1207: fix regex pattern in order to fetch the correct version in all cases

gkatiforis:
Resolves METAMODEL-1207

The Regex pattern used to fetch the database major version is repleced with the correct one that covers all possible cases. Also the fix is confimed with unit test.

### DIFF
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/VersionParser.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/VersionParser.java
@@ -1,0 +1,61 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.metamodel.jdbc.dialects;
+
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Responsible to parse an input string and return a substring that
+ * represents the version based on {@link VERSION_REGEX}.
+ */
+class VersionParser {
+    private static final String VERSION_REGEX = "[0-9]+(\\.[0-9]+)+";
+
+
+    /**
+     * @param stringToParse the string that contains the version.
+     * @return If the version exists returns the full version else empty string.
+     */
+    public static String getVersion(String stringToParse){
+        if(stringToParse != null) {
+            Matcher matcher = Pattern.compile(VERSION_REGEX).matcher(stringToParse);
+            if (matcher.find()){
+                return matcher.group();
+            }
+        }
+        return "";
+    }
+
+    /**
+     * @param stringToParse the string that contains the version.
+     * @return If the version exists returns only the number(s) before the first dot else 0.
+     */
+    public static int getMajorVersion(String stringToParse){
+        String fullVersion = getVersion(stringToParse);
+        if(fullVersion != null) {
+            int firstDot = fullVersion.indexOf('.');
+            if(firstDot >= 0) {
+                return Integer.valueOf(fullVersion.substring(0, firstDot));
+            }
+        }
+        return 0;
+    }
+}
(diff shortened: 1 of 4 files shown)


jdbc/src/test/java/org/apache/metamodel/jdbc/dialects/OracleQueryRewriterTest.java
```diff
@@ -92,6 +92,14 @@ public void testOffsetFetchVersionCheck() throws SQLException {
         Assert.assertEquals("The query shouldn't be rewritten.", query.toSql(), qr.rewriteQuery(query));
     }
 
+    @Test
+    public void testOffsetFetchVersionSpecialCaseCheck() throws SQLException {
```
kaspersorensen: The test case here seems to rely on a number of other factors and as such I think the assertion is a bit of a by-product of the version parsing logic, not actually testing the parsing itself.

gkatiforis:
I could create a new VersionParser class and write the appropriate unit tests.  Which is the correct package to add this class?

kaspersorensen:
Probably just add it in the `org.apache.metamodel.jdbc` package and make the class non-public so that it's only visible from other classes in the same package.

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/VersionParser.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.regex.Pattern;
+
+/**
+ * Responsible to parse an input string and return a substring that
+ * represents the version based on {@link VERSION_REGEX}.
+ */
+class VersionParser {
+    private static final String VERSION_REGEX = "[0-9]+(\\.[0-9]+)+";
+
+
```
kaspersorensen: No need for the double newlines

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/VersionParser.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Responsible to parse an input string and return a substring that
+ * represents the version based on {@link VERSION_REGEX}.
+ */
+class VersionParser {
+    private static final String VERSION_REGEX = "[0-9]+(\\.[0-9]+)+";
```
kaspersorensen: I would suggest you compile this regex here and hold on to the `Pattern` instead of the `String`. That way you only compile it once.

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/VersionParser.java
```diff
@@ -0,0 +1,61 @@
...
+     */
+    public static int getMajorVersion(String stringToParse){
+        String fullVersion = getVersion(stringToParse);
+        if(fullVersion != null) {
+            int firstDot = fullVersion.indexOf('.');
+            if(firstDot >= 0) {
+                return Integer.valueOf(fullVersion.substring(0, firstDot));
+            }
+        }
+        return 0;
```
kaspersorensen: `0` is a bad exception value I think, because versions might easily be 0-based. Better to go with `-1`.

jdbc/src/test/java/org/apache/metamodel/jdbc/dialects/OracleQueryRewriterTest.java
kaspersorensen: I would delete this test case now.

jdbc/src/test/java/org/apache/metamodel/jdbc/dialects/VersionParserTest.java
```diff
@@ -0,0 +1,82 @@
...
+ * under the License.
+ */
+package org.apache.metamodel.jdbc.dialects;
+
+import junit.framework.TestCase;
+
+public class VersionParserTest extends TestCase {
+
+	public void testGetVersion() throws Exception {
+		String version = VersionParser.getVersion("10.11.0");
```
kaspersorensen: Tabs instead of spaces.

jdbc/src/test/java/org/apache/metamodel/jdbc/dialects/VersionParserTest.java
```diff
@@ -0,0 +1,82 @@
...
+	}
+	public void testGetMajorVersionWithLettersAndNumbers() throws Exception {
+		int version = VersionParser.getMajorVersion("10 55g test 10.11.0 test 100");
+		assertEquals(10, version);
+	}
+	public void testGetMajorVersionMultipleVersions() throws Exception {
+		int version = VersionParser.getMajorVersion("test 10.11.0V 11.11 test");
+		assertEquals(10, version);
+	}
+}
```
kaspersorensen: No newline at end of file.

jdbc/src/main/java/org/apache/metamodel/jdbc/dialects/VersionParser.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+package org.apache.metamodel.jdbc.dialects;
+
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Responsible to parse an input string and return a substring that
+ * represents the version based on {@link VERSION_REGEX}.
```
kaspersorensen: No point in referencing a private member (VERSION_REGEX) in the public documentation. I would just rephrase to something like.

`This class is responsible for parsing version numbers in software products. Version strings are expected to be numeric 
and dot-separated.`

kaspersorensen:
Looks good to me. One of the builds failed but it is unrelated (flaky Excel module test). I'll merge it in.

PR #212, opened 2019-03-13
METAMODEL-1210: Module to read .arff files

kaspersorensen:


kaspersorensen:
It's been a while and I could self-merge it based on "lazy concensus" but I'd love it if someone took a quick glance at this.

LosD:
A short scan showed nothing worrying. It's rather self-contained, so I guess we can make a quick merge.

Interesting file format, btw!

kaspersorensen:
Thanks for checking it @LosD. Merging! :-)

PR #213, opened 2019-03-19
 METAMODEL-1172: Improves searching based on an expression in MapValueFunction

gkatiforis:
Resolves METAMODEL-1172:

Adds support in cases that the outer most type is a list or array.
Adds support for searching in nested lists or multidimensional arrays.

Unit tests that covers all cases are included.


kaspersorensen:
The build failed on 1/4 of the JDKs that it's running on. Maybe just a flaky test somewhere. You can restart a build by adding an empty commit:
```
git commit -m "Empty commit to trigger rebuild" --allow-empty
```

kaspersorensen:
Looks good to me.

PR #214, opened 2019-03-22
Add a method of creating a table with splitKey

q977734161:
Usually hbase has a lot of data to write, in order to spread the write pressure, the pre-partition will be created in advance, so I add a method of creating a table with splitKey.

### DIFF
--- a/hbase/src/main/java/org/apache/metamodel/hbase/HBaseClient.java
+++ b/hbase/src/main/java/org/apache/metamodel/hbase/HBaseClient.java
@@ -151,7 +151,7 @@ final class HBaseClient {
      * @throws MetaModelException when a {@link IOException} is caught
      */
     public void createTable(final String tableName, final Set<String> columnFamilies) {
-        checkTableAndCf(tableName, columnFamilies);
+        checkTableAndColumnFamilies(tableName, columnFamilies);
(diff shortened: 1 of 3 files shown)


hbase/src/main/java/org/apache/metamodel/hbase/HBaseClient.java
```diff
@@ -151,26 +151,53 @@ private boolean rowExists(final Table table, final byte[] rowKey) throws IOExcep
...
+    /**
+     * Creates a HBase table based on a tableName and it's columnFamilies and splitKeys
+     * @param tableName
+     * @param columnFamilies
+     * @param splitKeys
+     * @throws IllegalArgumentException when any parameter is null
+     * @throws MetaModelException when a {@link IOException} is caught
+     */
+    public void createTable(final String tableName, final Set<String> columnFamilies, byte[][] splitKeys) {
+        checkTableAndCf(tableName, columnFamilies);
```
kaspersorensen: I think you should spell out "column families" instead of the "cf" abbreviation

q977734161: ok,I will spell out "column families" instead of the "cf" abbreviation

kaspersorensen:
LGTM!

PR #218, opened 2019-04-08
Use Scan.addColumn(...) instead  of Scan.addFamily(...) when has qualifier.

q977734161:
when columns contain family and column name,using Scan.addColumn(...) method to reduce network pressure and decrease unnecessary columns.

### DIFF
--- a/hbase/src/main/java/org/apache/metamodel/hbase/HBaseDataContext.java
+++ b/hbase/src/main/java/org/apache/metamodel/hbase/HBaseDataContext.java
@@ -222,9 +222,11 @@ public class HBaseDataContext extends QueryPostprocessDataContext implements Upd
         for (Column column : columns) {
             if (!column.isPrimaryKey()) {
                 final int colonIndex = column.getName().indexOf(':');
+                final int len = column.getName().length();
                 if (colonIndex != -1) {
                     String family = column.getName().substring(0, colonIndex);
-                    scan.addFamily(family.getBytes());
+                    String colName = column.getName().substring(colonIndex + 1, len);
+                    scan.addColumn(family.getBytes(),colName.getBytes());
                 } else {
                     scan.addFamily(column.getName().getBytes());
                 }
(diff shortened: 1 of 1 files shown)


hbase/src/main/java/org/apache/metamodel/hbase/HBaseDataContext.java
```diff
@@ -222,9 +222,11 @@ protected DataSet materializeMainSchemaTable(Table table, List<Column> columns,
         for (Column column : columns) {
             if (!column.isPrimaryKey()) {
                 final int colonIndex = column.getName().indexOf(':');
+                final int len = column.getName().length();
                 if (colonIndex != -1) {
                     String family = column.getName().substring(0, colonIndex);
-                    scan.addFamily(family.getBytes());
+                    String colName = column.getName().substring(colonIndex + 1, len);
```
kaspersorensen: Remember to add `final` :-)

hbase/src/main/java/org/apache/metamodel/hbase/HBaseDataContext.java
```diff
@@ -222,9 +222,11 @@ protected DataSet materializeMainSchemaTable(Table table, List<Column> columns,
...
+                    scan.addColumn(family.getBytes(),colName.getBytes());
```
kaspersorensen: ... and a space after the comma :-)

q977734161: Ok,I will modify the code at once and thanks Peter kaspersorensen for reviewing.

kaspersorensen:
Sorry, seems the test flakyness is not completely over. Can you commit another empty commit to trigger rebuild?

q977734161:
ok,I will commit another empty commit to trigger rebuild.

kaspersorensen:
Merging this and #219

PR #219, opened 2019-04-15
Improved stability of Salesforce test and exception handling.

kaspersorensen:
This should avoid failures such as the one that broke the assertion in
Travis CI recently; https://api.travis-ci.org/v3/job/517589252/log.txt


kaspersorensen:
Merging this and #218

PR #220, opened 2019-04-24
Recursively get SelectItems from FilterItems when using nested loop join

arjansh:
Fixes MM-1211.

When determining the Carthesian product for a query over multiple tables make sure that when determining which FilterItems are applicable for the query the SelectItems for compound FilterItems are also taken into consideration by recursively inspecting the children of the compound FilterItems.


kaspersorensen:
Sorry for not noticing this PR for so long. Looks good to me!

PR #221, opened 2019-06-12
Access CSVParser in a synchronized manner

arjansh:
Fixes MM-1213.

The _SingleLineCsvRow_ class uses _com.opencsv.CsvParser_ to read lines from a CSV file. If this is done in a parallel manner the results can be corrupted because the _CsvParser_ itself doesn't provide thread safety. Therefore we have to access it in a synchronized manner.


kaspersorensen:
That's a great find and we need to fix it and certainly keep the test case. But I wonder if the solution is right. We're going to be adding a ton of locking and unlocking to our parallel reader which might very well just make it slower than the non-parallel one. I think a "shared nothing" approach to the problem might be better. Could we either instantiate the CsvParser every time, or is there a static way of parsing a line which does not interfere with the CsvParser's state?

arjansh:
I agree that adding the synchronization may not be the best solution performance wise. I don't know what the overhead for instantiating the CsvParser every time is, but I can look into that. I will also check if there is an option to parse the line in a more static manner.

arjansh:
I removed the synchronization in favor of instantiating the CSVParser each time. Opencsv doesn't provide an option to parse a line in a static way.

To achieve this I deprecated the current constructor for SingleLineCsvDataSet and introduced a new one which takes has a _CsvConfiguration_ object as a parameter and I moved the _createParser_  method which creates the CSVParser to the _CsvConfiguration_  class.

I'm not sure if this is the best approach, so please provide some feedback.

csv/src/main/java/org/apache/metamodel/csv/CsvConfiguration.java
```diff
@@ -195,4 +199,17 @@ public String toString() {
                 + ", separatorChar=" + separatorChar + ", quoteChar=" + quoteChar + ", escapeChar=" + escapeChar
                 + ", failOnInconsistentRowLength=" + failOnInconsistentRowLength + "]";
     }
+    
+    public ICSVParser createParser() {
```
kaspersorensen: Hmm I'd ideally like this somewhere else than in CsvConfiguration. I consider CsvConfiguration part of MetaModel's public API and so it would be bad to also expose OpenCSV classes. Just in case we ever need to switch to another CSV parser library.

csv/src/main/java/org/apache/metamodel/csv/SingleLineCsvDataSet.java
```diff
@@ -44,20 +44,56 @@
...
+     * @param maxRows
+     * @param columnsInTable
+     * @param failOnInconsistentRowLength
+     * 
+     * @deprecated When instantiating a {@link SingleLineCsvDataSet} using this constructor the {@link ICSVParser} can
+     *             be reused in different threads in a parallel manner which is not promised to work by the
+     *             {@link ICSVParser}. Use
+     *             {@link #SingleLineCsvDataSet(BufferedReader, CsvConfiguration, List, Integer, int, boolean)} instead.
+     */
+    @Deprecated
```
kaspersorensen: This class is not public however, so we don't even need to worry about deprecation etc. Just whack the old constructor.

csv/src/main/java/org/apache/metamodel/csv/CsvConfiguration.java
```diff
@@ -195,4 +199,17 @@ public String toString() {
...
+        if (getEscapeChar() == getQuoteChar()) {
+            return new RFC4180ParserBuilder().withSeparator(getSeparatorChar()).withQuoteChar(getQuoteChar()).build();
+        } else {
+            return new CSVParserBuilder()
+                    .withSeparator(getSeparatorChar())
+                    .withQuoteChar(getQuoteChar())
+                    .withEscapeChar(getEscapeChar())
+                    .build();
```
LosD: Does the builder also have thread safety issues? Otherwise it might be a good idea to save instantiations by keeping the builder instead of using a new one every time.

arjansh: As far as I know the builder doesn't have those issues, but I refrained from that up until now because the _RFC4180ParserBuilder_ and the _CSVParserBuilder_ don't share one super class or interface.

LosD: Ah, yes, that definitely makes it harder to do in a sane way (and is a bit silly, given that they at least share the `build()` method).

csv/src/main/java/org/apache/metamodel/csv/CsvParserBuilder.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+package org.apache.metamodel.csv;
+
+import com.opencsv.CSVParserBuilder;
+import com.opencsv.ICSVParser;
+import com.opencsv.RFC4180ParserBuilder;
+
+class CsvParserBuilder {
+    final CSVParserBuilder _csvParserBuilder;
+    final RFC4180ParserBuilder _rfc4180ParserBuilder;
```
kaspersorensen: Should probably be private.

arjansh: I now made them private.

kaspersorensen:
Travis CI errors are unrelated (old annoying flaky tests).

PR #222, opened 2019-07-17
METAMODEL-1151: Fixed by adding DataContextFactory for SugarCRM.

kaspersorensen:
As per https://issues.apache.org/jira/browse/METAMODEL-1151 only SugarCRM was missing a DataContextFactory, so I added it :-)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 ### Apache MetaModel _WIP_
 
+ * [METAMODEL-1151] - Created DataContextFactory implementations for earch remaining module.
  * [METAMODEL-1213] - Fixed concurrency bug in single-line CSV parallel reader.
  * [METAMODEL-1211] - Fixed bug of getting all SelectItems when performing nested loop joins.
  * Improved HBase query building by scanning columns instead of column families.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 ### Apache MetaModel _WIP_
 
- * [METAMODEL-1151] - Created DataContextFactory implementations for earch remaining module.
+ * [METAMODEL-1151] - Created DataContextFactory implementations for each remaining module.
  * [METAMODEL-1213] - Fixed concurrency bug in single-line CSV parallel reader.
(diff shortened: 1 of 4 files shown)


sugarcrm/src/main/java/org/apache/metamodel/sugarcrm/SugarCrmDataContextFactory.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public class SugarCrmDataContextFactory extends AbstractDataContextFactory {
+
+    @Override
+    protected String getType() {
+        return "sugarcrm";
+    }
+
+    @Override
+    public DataContext create(DataContextProperties properties, ResourceFactoryRegistry resourceFactoryRegistry)
```
LosD: Can't remember if we have "rules" here, but why not `final`? Arguments are just as susceptible to being manipulated as variables, and can be just as surprising (if not more).

(of course, in as simple a function as this, it doesn't really matter for any of them)

CHANGES.md
```diff
@@ -1,5 +1,6 @@
 ### Apache MetaModel _WIP_
 
+ * [METAMODEL-1151] - Created DataContextFactory implementations for earch remaining module.
```
LosD: Typo, earch => each

dynamodb/src/main/java/org/apache/metamodel/dynamodb/DynamoDbDataContextFactory.java
```diff
@@ -31,6 +32,9 @@
 import com.amazonaws.services.dynamodbv2.AmazonDynamoDB;
 import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClientBuilder;
 
+/**
+ * {@link DataContextFactory} for DynamoDB
```
LosD: A liiiiittle bit useless JavaDoc comment. :)

kaspersorensen:
Thanks for the review @LosD, I love that you're obnoxiously pedantic and detail oriented ;-) I corrected the PR as per your feedback items.

LosD:
It's what I do :smile:

LGTM! Will merge when I'm back at my computer (you can of course also just do it now if you prefer)

PR #223, opened 2019-07-22
MM-1214: DefaultTableAliasedSchema.equals for schema comparison.

jhorcicka:
Fixes MM-1214. 

DefaultTableAliasedSchema.equals(...) method was added to use "wrappedSchema" field for comparison.


kaspersorensen:
I wonder if this is sensible. Would you really say that the "alias" schema is equal to the schema that it's aliasing? I think that may create other problems down the line.

If it turns out that I'm wrong then I have another remark, namely that the `hashCode()` method should be implemented according to the same logic then.

jhorcicka:
Good point. 

I tried another approach instead. In `CompositeDataContext.getDataContext` I can check the actual type of `dcSchema` and cast it. I believe it's less intrusive now. What is your opinion about that, @kaspersorensen?

core/src/main/java/org/apache/metamodel/CompositeDataContext.java
```diff
@@ -122,7 +123,8 @@ private DataContext getDataContext(Table table) {
                         Schema dcSchema = dc.getSchemaByName(schema.getName());
                         if (dcSchema != null) {
                             // second round = try with schema equals method
-                            if (dcSchema.equals(schema)) {
+                            if (dcSchema.equals(schema) || (dcSchema instanceof DefaultTableAliasedSchema
```
kaspersorensen: I would use the interface `WrappingSchema` here.

jhorcicka: Updated. Thanks for the tip.

kaspersorensen:
And now you unfortunately fell victim to one of our sporadically failing tests. But LGTM now.

kaspersorensen:
Please also update the CHANGES.md file. Then I will merge.

jhorcicka:
Updated.

CHANGES.md
```diff
@@ -1,9 +1,9 @@
 ### Apache MetaModel _WIP_
 
- * [METAMODEL-1151] - Created DataContextFactory implementations for each remaining module.
```
kaspersorensen: We shouldn't remove this item

jhorcicka: Oh, sure. That was my merging error. Updated.

PR #225, opened 2019-08-08
METAMODEL-1217 - Fixes dropping JDBC Tables with view as TableType

GerardDellemann:
Fixes: [METAMODEL-1217 - Drop table doesn't work for views with JDBC Tables](https://issues.apache.org/jira/browse/METAMODEL-1217)

### DIFF
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDropTableBuilder.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDropTableBuilder.java
@@ -66,12 +66,12 @@ final class JdbcDropTableBuilder extends AbstractTableDropBuilder implements Tab
     protected String createSqlStatement() {
         final Table table = getTable();
         final FromItem fromItem = new FromItem(table);
-        final String tableLabel = _queryRewriter.rewriteFromItem(fromItem);
+        final String qualifiedTableName = _queryRewriter.rewriteFromItem(fromItem);
 
         if (table.getType() != null && table.getType() == TableType.VIEW) {
-            return "DROP VIEW " + tableLabel;
+            return "DROP VIEW " + qualifiedTableName;
         } else {
-            return "DROP TABLE " + tableLabel;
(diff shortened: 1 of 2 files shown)


jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDropTableBuilder.java
```diff
@@ -63,10 +64,15 @@ public void execute() {
...
     protected String createSqlStatement() {
-        FromItem fromItem = new FromItem(getTable());
-        String tableLabel = _queryRewriter.rewriteFromItem(fromItem);
+        final Table table = getTable();
+        final FromItem fromItem = new FromItem(table);
+        final String tableLabel = _queryRewriter.rewriteFromItem(fromItem);
 
-        return "DROP TABLE " + tableLabel;
+        if (table.getType() != null && table.getType() == TableType.VIEW) {
+            return "DROP VIEW " + tableLabel;
```
kaspersorensen: I wonder if we could simply go like this:

```
TableType type = table.getType() == null ? TableType.TABLE : table.getType();
return "DROP " + type + " " + tableLabel;
```

Oh, and I would maybe rename `tableLabel` to `qualifiedTableName` or something.

GerardDellemann: TableType has the following values: TABLE, VIEW, SYSTEM_TABLE, GLOBAL_TEMPORARY, LOCAL_TEMPORARY, ALIAS, SYNONYM, OTHER.
In SQL you either drop a table or a view (you don't drop a system_table, that's a table). I would say that the VIEW value is only relevant for droping views. All the other values beloing to droping tables.
TableType has no toString implementation. You could implement a custom toString just for this JdbcDropTableBuilder#createSqlStatement() method, but I don't think that would be a very representable implementation of that method.

GerardDellemann: I'll change the variable name.

PR #226, opened 2019-08-28
MM-1218 Compare schemas with name null

GerardDellemann:
Closes: [METAMODEL-1218 - Can't compare schema's without a name](https://issues.apache.org/jira/browse/METAMODEL-1218)


kaspersorensen:
Seems reasonable to me. I wonder if we may want to look at other attributes of the schema such as the tables or whatever. Problem is to avoid it becoming cyclical :-)

GerardDellemann:
I think that goes a little to far. If the need for this arises, then we could do that.

PR #227, opened 2019-09-02
MM-1219 Returning schemas SQL Server

GerardDellemann:
Closes: [MM-1219 - Returning schemas for SQL Server](https://issues.apache.org/jira/browse/METAMODEL-1219).

This is one possible solution to the problem I've reported in JIRA.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 ### Apache MetaModel _WIP_
 
+ * [METAMODEL-1219] - Improved returning schemas for SQL Server
  * [METAMODEL-1217] - Fixes dropping JDBC Tables with view as TableType
  * [METAMODEL-1215] - Removed deprecated code from pre-Java 8 times.
  * [METAMODEL-1214] - CompositeDataContext uses wrappedSchema for comparison when searching for it in the cache.
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
@@ -802,7 +802,7 @@ public class JdbcDataContext extends AbstractDataContext implements UpdateableDa
         return schemas;
     }
 
-    private boolean hasDefaultSQLServerSchema(final DatabaseMetaData metaData) throws SQLException {
+    private static boolean hasDefaultSQLServerSchema(final DatabaseMetaData metaData) throws SQLException {
         ResultSet schemas = null;
         try {
             schemas = metaData.getSchemas();
--- a/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
+++ b/jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
@@ -101,7 +101,8 @@ public class JdbcDataContext extends AbstractDataContext implements UpdateableDa
     public static final String DATABASE_PRODUCT_IMPALA = "Impala";
 		
     private static final String DEFAULT_SCHEMA_NAME_SQLSERVER = "dbo";
-    private static final String RESULT_SET_COLUMN_NAME_SCHEMA_SQLSERVER = "TABLE_SCHEM";
+
+    private static final String SCHEMA_NAME_IDENTIFIER = "TABLE_SCHEM";
 
     public static final ColumnType COLUMN_TYPE_CLOB_AS_STRING =
             new ColumnTypeImpl("CLOB", SuperColumnType.LITERAL_TYPE, String.class, true);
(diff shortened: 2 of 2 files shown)


jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
```diff
@@ -778,18 +780,33 @@ private String findDefaultSchema(final String defaultName, final List<String> sc
...
+        // Find DBO schema if present
         try {
-            rs = metaData.getTables(_catalogName, null, null, JdbcUtils.getTableTypesAsStrings(_tableTypes));
-            while (rs.next()) {
-                schemas.add(rs.getString("TABLE_SCHEM"));
+            resultSet = metaData.getSchemas();
+            while (resultSet.next()) {
+                if (resultSet.getString("TABLE_SCHEM").equals(DEFAULT_SCHEMA_NAME_SQLSERVER)) {
+                    schemas.add(DEFAULT_SCHEMA_NAME_SQLSERVER);
+					break;
```
kaspersorensen: Weird indentation here

GerardDellemann: ah made that editing later with Notepad++ :)

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
```diff
@@ -778,18 +780,33 @@ private String findDefaultSchema(final String defaultName, final List<String> sc
...
-        // Distinct schema names. metaData.getTables() is a denormalized
-        // resultset
-        Set<String> schemas = new HashSet<>();
-        ResultSet rs = null;
+        // Distinct schema names. metaData.getTables() is a denormalized resultset
+        final Set<String> schemas = new HashSet<>();
+        ResultSet resultSet = null;
+
+        // Find DBO schema if present
         try {
```
arjansh: Since `ResultSet` is `AutoCloseable`, isn't it better to change this to try-with-resources?

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
```diff
@@ -778,18 +780,33 @@ private String findDefaultSchema(final String defaultName, final List<String> sc
...
+                    break;
+                }
             }
         } finally {
-            FileHelper.safeClose(rs);
+            FileHelper.safeClose(resultSet);
         }
+
+        // Find other schemas
+        try {
```
arjansh: Same here as on line 788, since `ResultSet` is `AutoCloseable`, isn't it better to change this to try-with-resources?

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
```diff
@@ -778,18 +780,33 @@ private String findDefaultSchema(final String defaultName, final List<String> sc
...
+        ResultSet resultSet = null;
+
+        // Find DBO schema if present
         try {
-            rs = metaData.getTables(_catalogName, null, null, JdbcUtils.getTableTypesAsStrings(_tableTypes));
-            while (rs.next()) {
-                schemas.add(rs.getString("TABLE_SCHEM"));
+            resultSet = metaData.getSchemas();
+            while (resultSet.next()) {
+                if (resultSet.getString("TABLE_SCHEM").equals(DEFAULT_SCHEMA_NAME_SQLSERVER)) {
```
arjansh: The String `"TABLE_SCHEM"` is used twice now in the changed code, can you extract it into a constant?

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
arjansh: The added piece of code essentially checks if the default SQL server schema is available, and if so adds it to the schema, then it executes the already existing logic. For understand-ability of the code it might be better to extract the logic which checks if the default SQL server schema is available into a separate method, e.g. 

```
    if (hasDefaultSQLServerSchema(metaData) {
        schemas.add(DEFAULT_SCHEMA_NAME_SQLSERVER);
    }
```

and

```
    private boolean hasDefaultSQLServerSchema(final DatabaseMetaData metaData) throws SQLException {
        try (ResultSet schemas = metaData.getSchemas()) {
            while (schemas.next()) {
                if (schemas.getString("TABLE_SCHEM").equals(DEFAULT_SCHEMA_NAME_SQLSERVER)) {
                    return true;
                }
            }
        }
        return false;
    }
```

GerardDellemann: I wanted to implement that, but then the FileHelper#safeClose(Object... objects) won't be called anymore. I'm not sure if this is a better method then what's implemented by the AutoCloseable#close() implementation?

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
GerardDellemann: See previous answer.

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
GerardDellemann: sure

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
GerardDellemann: sure

arjansh: I personally think we should use the try-with-resources. The `FileHelper#safeClose(Object)` is most probably used right now because this functionality was originally implemented in a version of java which didn't support the try-with-resources construct yet. @kaspersorensen Am I correct regarding this?

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
```diff
@@ -99,6 +99,9 @@
     public static final String DATABASE_PRODUCT_HIVE = "Apache Hive";
     public static final String DATABASE_PRODUCT_SQLITE = "SQLite";
     public static final String DATABASE_PRODUCT_IMPALA = "Impala";
+		
+    private static final String DEFAULT_SCHEMA_NAME_SQLSERVER = "dbo";
+    private static final String RESULT_SET_COLUMN_NAME_SCHEMA_SQLSERVER = "TABLE_SCHEM";
```
arjansh: I would propose a different name for this constant, because `'TABLE_SCHEM"` isn't SQL Server specific. I would instead call it TABLE_SCHEMA_NAME_IDENTIFIER or SCHEMA_NAME_IDENTIFIER or something similar.

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
kaspersorensen: Well yes and no. There is a difference... the `safeClose` method will log but ignore any exceptions. It was built from a mentality of "close this resource, but if that fails I don't know that we can do anything about it".

I'm OK with changing it though.

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
GerardDellemann: Implemented try-with-resources

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
GerardDellemann: Sure

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcDataContext.java
GerardDellemann: Implemented try-with-resources

CHANGES.md
```diff
@@ -1,5 +1,6 @@
 ### Apache MetaModel _WIP_
 
+ * [METAMODEL-1219] - Improved returning schemas for SQL Server
```
arjansh: The pull request looks good to me, it can't be merged however, because this change conflicts with another update to the CHANGES.md file on apache:master. Can you merge apache:master into your branch?

GerardDellemann: Resolved the conflict

arjansh:
Looks good to me.

PR #229, opened 2019-09-18
Upgrade to Elasticsearch 7.3.1

arjansh:
Fixes METAMODEL-1221.

Biggest differences between 7.x and 5.x:
- Elasticsearch 7.x no longer supports multiple types in indexes, so we can no longer create multiple tables in a schema.
- The TransportClient in Elasticsearch has been deprecated.

Regarding my proposed changes:
- I chose to deprecate the classes in the elasticsearch-native module, but we may also choose to remove that module.
- As for the elasticsearch-rest module:
-- Because Elasticsearch 7 no longer supports types, I also don't use any of the (now deprecated) apis which can be used to specify a type when using the rest client. 
-- When trying to create a new table or inserting into a new table when there already is a default type available in the index a MetaModelException will be thrown.
-- Elasticsearch 7 has a default type which isn't explicitly used when accessing it, but it is there and it is called "_doc". I added this as a constant, so we can use that as table name when querying Elasticsearch.
-- The `ElasticSeachRestUpdateCallback#execute(ActionRequest)` method is public, but I would like to make it package private, because it contains an Elasticsearch internal class in its method signature and is only meant to be used internally by other classes in the same package. Is this something I can just change or do I have to do some evil trickery?


kaspersorensen:
I haven't worked with ES 7 myself so my review is going to be superficial. To answer this question:
```
The ElasticSeachRestUpdateCallback#execute(ActionRequest) method is public, but I would like to make it package private, because it contains an Elasticsearch internal class in its method signature and is only meant to be used internally by other classes in the same package. Is this something I can just change or do I have to do some evil trickery?
```
I would say that you can just make it package private. I don't think we should expose this sort of stuff as an API.

elasticsearch/native/src/main/java/org/apache/metamodel/elasticsearch/nativeclient/ElasticSearchDataContextFactory.java
```diff
@@ -131,10 +135,16 @@ public DataContext create(DataContextProperties properties, ResourceFactoryRegis
     }
 
     private Client createTransportClient(DataContextProperties properties) {
-        final Settings settings = Settings.builder().put().put("name", "MetaModel").put("cluster.name", getCluster(properties)).build();
+        final Settings settings = Settings
+                .builder()
+                .put("name", "MetaModel")
```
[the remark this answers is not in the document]
arjansh: I had to remove one "put()" here to get it to work, otherwise I wouldn't have touched this piece of code at all. Since the nativeclient will be deprected as part of this PR, I'd rather change as little of the internal implementation as possible.

kaspersorensen:
Now the build failed :-/

kaspersorensen:
LGTM. Again, not an ES 7.x expert, but I guess we're at lazy concensus time now. Do you want it merged @arjansh ?

arjansh:
Hi @kaspersorensen, I asked @jhorcicka if he can also look at this PR, even though he's not a committer on MetaModel, but he's been implementing some code last weeks which uses Elasticsearch 7.3.1, so he may have some useful feedback. I'd like to wait for that first.

jhorcicka:
While using this branch, I get the following NPE: 

```
ERROR 16:30:26 ElasticSearchRestDataContext - Unexpected error during detectTable for document type '_doc'
java.lang.NullPointerException
	at org.apache.metamodel.elasticsearch.common.ElasticSearchMetaDataParser.parse(ElasticSearchMetaDataParser.java:41)
	at org.apache.metamodel.elasticsearch.rest.ElasticSearchRestDataContext.detectTable(ElasticSearchRestDataContext.java:187)
	at org.apache.metamodel.elasticsearch.rest.ElasticSearchRestDataContext.detectSchema(ElasticSearchRestDataContext.java:155)
	at org.apache.metamodel.elasticsearch.rest.ElasticSearchRestDataContext.<init>(ElasticSearchRestDataContext.java:111)
	at org.apache.metamodel.elasticsearch.rest.ElasticSearchRestDataContext.<init>(ElasticSearchRestDataContext.java:124)
	at org.datacleaner.extension.elasticsearch.MockElasticSearchDatastore.createDatastoreConnection(MockElasticSearchDatastore.java:28)
	at org.datacleaner.connection.UsageAwareDatastore.getDatastoreConnection(UsageAwareDatastore.java:97)
	at org.datacleaner.connection.UsageAwareDatastore.openConnection(UsageAwareDatastore.java:108)
	at org.datacleaner.connection.ElasticSearchDatastore.openConnection(ElasticSearchDatastore.java:183)
	at org.datacleaner.extension.elasticsearch.ElasticSearchIndexAnalyzer.init(ElasticSearchIndexAnalyzer.java:100)
```

It would be nice to fix it here. Add a null check for "metaDataInfo".

elasticsearch/rest/src/test/java/org/apache/metamodel/elasticsearch/rest/ElasticSearchRestDataContexFactoryIT.java
```diff
@@ -55,22 +63,24 @@
     public void setUp() throws Exception {
         dockerHostAddress = ElasticSearchRestDataContextIT.determineHostName();
 
-        externalClient = new ElasticSearchRestClient(RestClient.builder(new HttpHost(dockerHostAddress, 9200)).build());
+        externalClient = new RestHighLevelClient(RestClient.builder(new HttpHost(dockerHostAddress, 9200)));
```
jhorcicka: 9200 is actually used on multiple locations in this module. A constant would be nice.

elasticsearch/rest/src/test/java/org/apache/metamodel/elasticsearch/rest/ElasticSearchRestDataContextIT.java
```diff
@@ -119,22 +117,30 @@ private static void insertPeopleDocuments() throws IOException {
...
         assertThat(table.getColumnNames(), containsInAnyOrder("_id", "message", "postDate", "user"));
 
         assertEquals(ColumnType.STRING, table.getColumnByName("user").getType());
         assertEquals(ColumnType.DATE, table.getColumnByName("postDate").getType());
         assertEquals(ColumnType.BIGINT, table.getColumnByName("message").getType());
 
-        try (DataSet ds = dataContext.query().from(indexType1).select("user").and("message").execute()) {
+        dataContext.refreshSchemas();
+
+        try (DataSet ds = dataContext.query().from(DEFAULT_TABLE_NAME).select("user").and("message").execute()) {
```
jhorcicka: final?

elasticsearch/rest/src/test/java/org/apache/metamodel/elasticsearch/rest/ElasticSearchRestDataContextIT.java
```diff
@@ -144,8 +150,10 @@ public void testSimpleQuery() throws Exception {
 
     @Test
     public void testDocumentIdAsPrimaryKey() throws Exception {
-        Table table = dataContext.getDefaultSchema().getTableByName("tweet2");
-        Column[] pks = table.getPrimaryKeys().toArray(new Column[0]);
+        indexType2TweeterDocuments();
+
+        final Table table = dataContext.getDefaultSchema().getTableByName(DEFAULT_TABLE_NAME);
+        final Column[] pks = table.getPrimaryKeys().toArray(new Column[0]);
```
jhorcicka: Please use a better name (primaryKeys, primaryKeyColumns).

elasticsearch/rest/src/test/java/org/apache/metamodel/elasticsearch/rest/ElasticSearchRestDataContextIT.java
```diff
@@ -155,14 +163,24 @@ public void testDocumentIdAsPrimaryKey() throws Exception {
...
+    }
+
     @Test
     public void testExecutePrimaryKeyLookupQuery() throws Exception {
-        Table table = dataContext.getDefaultSchema().getTableByName("tweet2");
-        Column[] pks = table.getPrimaryKeys().toArray(new Column[0]);
+        indexType2TweeterDocuments();
+
+        final Table table = dataContext.getDefaultSchema().getTableByName(DEFAULT_TABLE_NAME);
+        final Column[] pks = table.getPrimaryKeys().toArray(new Column[0]);
```
jhorcicka: Please use a better name (primaryKeys, primaryKeyColumns).

elasticsearch/rest/src/test/java/org/apache/metamodel/elasticsearch/rest/ElasticSearchRestDataContextIT.java
```diff
@@ -171,45 +189,56 @@ public void testExecutePrimaryKeyLookupQuery() throws Exception {
         }
     }
 
+    @Test
+    public void testMissingPrimaryKeyLookupQuery() throws Exception {
+        indexType2TweeterDocuments();
+
+        final Table table = dataContext.getDefaultSchema().getTableByName(DEFAULT_TABLE_NAME);
+        final Column[] pks = table.getPrimaryKeys().toArray(new Column[0]);
```
jhorcicka: Please use a better name (primaryKeys, primaryKeyColumns).

elasticsearch/rest/src/test/java/org/apache/metamodel/elasticsearch/rest/ElasticSearchRestDataContextIT.java
```diff
@@ -347,8 +361,16 @@ public void run(UpdateCallback callback) {
 
     @Test
     public void testWhereColumnEqualsValues() throws Exception {
-        try (DataSet ds = dataContext.query().from(bulkIndexType).select("user").and("message").where("user")
-                .isEquals("user4").execute()) {
+        indexBulkDocuments(INDEX_NAME, 10);
+
+        try (DataSet ds = dataContext
```
jhorcicka: I'd rather use a better name for "ds", everywhere (dataSet).

elasticsearch/rest/src/test/java/org/apache/metamodel/elasticsearch/rest/ElasticSearchRestDataContextIT.java
```diff
@@ -412,19 +465,22 @@ public void testWhereColumnInValues() throws Exception {
 
     @Test
     public void testGroupByQuery() throws Exception {
-        Table table = dataContext.getDefaultSchema().getTableByName(peopleIndexType);
+        insertPeopleDocuments();
+
+        final Table table = dataContext.getDefaultSchema().getTableByName(DEFAULT_TABLE_NAME);
 
-        Query q = new Query();
+        final Query q = new Query();
```
jhorcicka: "q" => "query"

elasticsearch/common/src/main/java/org/apache/metamodel/elasticsearch/common/ElasticSearchUtils.java
```diff
@@ -267,6 +264,10 @@ public static ColumnType getColumnTypeFromElasticSearchType(final String metaDat
     }
 
     public static Row createRow(final Map<String, Object> sourceMap, final String documentId, final DataSetHeader header) {
+        if (sourceMap == null) {
+            return null;
```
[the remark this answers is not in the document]
arjansh: This is a static helper method which creates a row for a source map and as such if no source map is passed as an argument it shouldn't return a row. I'll add a javadoc comment which explains a bit what the method does to make this clear.

elasticsearch/rest/src/main/java/org/apache/metamodel/elasticsearch/rest/ElasticSearchRestClient.java
```diff
@@ -18,117 +18,80 @@
...
+            }
+
+            @Override
+            public Object getValue() {
+                return entry.getValue().getSourceAsMap();
+            }
+
+            @Override
+            public Object setValue(final Object value) {
+                return null;
```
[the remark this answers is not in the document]
arjansh: This is a private method used only in the method right above, where we only read from the Entry, so there is no need to implement this method.

elasticsearch/common/src/main/java/org/apache/metamodel/elasticsearch/common/ElasticSearchUtils.java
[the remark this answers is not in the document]
jhorcicka: OK

elasticsearch/rest/src/main/java/org/apache/metamodel/elasticsearch/rest/ElasticSearchRestClient.java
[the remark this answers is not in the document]
jhorcicka: OK

jhorcicka:
All my points were addressed.

elasticsearch/common/src/main/java/org/apache/metamodel/elasticsearch/common/ElasticSearchUtils.java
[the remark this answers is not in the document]
LosD: While the reasoning is good, it doesn't really change the worry: This breaks the API, so at the very least it should be considered if it requires a major release.

arjansh:
When doing lots of updates (more than 500) in a short period of time (less than 60 seconds), Elasticsearch will fail with this message: "Trying to create too many scroll contexts. Must be less than or equal to: [500]". This is caused by the fact that in the DeleteAndInsertBuilder#getRowsToUpdate() method a DataSet is created, which is converted to a List of rows, which is returned by the method, but the DataSet isn't closed, as a result of which, in case of Elasticsearch, the scroll request for the DataSet isn't cleared and will stay alive for 60 seconds.

I want to address this problem within this pull request.

LosD:
If its a separate problem from the upgrade, and this has already been reviewed, wouldn't a separate issue and PR be better?

arjansh:
That's a thought that I was having too, but I also have to fix a little bug in the AbstractElasticSearchDataSet#close() method, which is also involved, because the logic which detects if it should actually close the DataSet isn't implemented in the right manner (so right now it doesn't actually close it when the close method is invoked. And I want to add a few tests to demonstrate the issue.

But you maybe right and it maybe better to do that in a separate PR, but then this one preferably should be merged first.

arjansh:
Hmm, I don't even think this one actually has to be merged before. I'm gonna try it on master branch with Elasticsearch 5.6.3 too.

arjansh:
Hmm, if I would fix it based on master I would probably afterwards have to refactor the unit tests I create for it to work with Elasticsearch 7.3.1, so preferably I would implement it based on this branch.

LosD:
It's not a huge thing in any case, so if it's much easier to just close the scroll context here, it makes sense.

arjansh:
Okay, I'll stick to my original plan then and fix it within the context if this PR.

elasticsearch/rest/src/main/java/org/apache/metamodel/elasticsearch/rest/ElasticSearchRestDataSet.java
```diff
@@ -50,12 +50,15 @@ public ElasticSearchRestDataSet(final RestHighLevelClient client, final SearchRe
...
-            logger.warn("Could not clear scroll.", e);
+        
+        if (scrollId != null) {
+            final ClearScrollRequest clearScrollRequest = new ClearScrollRequest();
+            clearScrollRequest.addScrollId(scrollId);
+            try {
+                client.clearScroll((ClearScrollRequest) clearScrollRequest, RequestOptions.DEFAULT);
+            } catch (final IOException e) {
+                logger.warn("Could not clear scroll.", e);
+            }
```
LosD: Is this a local operation, or does the scroll live on the server? If the latter (as the "...Request" name seems to imply), isn't this a very expensive operation, when it's not done in batches?

(sorry, it's been waaaaay too long ago that I touched ES)

arjansh: It lives on the server and yes, it is a relatively expensive operation. Typically the scroll id is used when iterating over larger sets of data, so you do a query and get a thousand results, out of which only 400 are returned by the first HTTP request, then you use the scroll id to get the second 400, and you clear the scroll id, once you're done with your `DataSet`. This is a typical use case for which you use the scrolling mechanism and you should not have to clear it often.

What we do in MetaModel however is that when updating data using the `DeleteAndInsertBuilder` it first gets all rows to update (resulting in a scroll id, which is also retained on the Elasticsearch cluster), then it deletes all these rows, then it updates them locally (that is the Row objects) and then it inserts them into the actual table.

If you update data by quickly in succession calling MetaModel's update mechanism using the `DeleteAndInsertBuilder` it will quickly create a lot of "scrolls" of which MetaModel only allows 500. Within the `DeleteAndInsertBuilder` we however don't know on beforehand if the update query which is fired will only update one (or just a few) records (in case we could opt not to use the scrolling mechanism at all, because of the overhead it introduces) or if it will update thousands of rows.

So if we want to optimize that, then we should probably come up with something which I think is out of scope for this PR.

LosD: Ouch. Rock, meet hard place. I agree, working correctly trumps performance, and it sounds like a big change.

arjansh: A bit out of scope, but looking at the `DeleteAndInsertBuilder`, if a run-time exception occurs somewhere in the logic invoked on line 61, we're essentially losing data, because the data which has been deleted won't be reinserted. Am I correct here?

LosD: Yeah, I think so, but that's more or less life without transactions. It can't really be done the other way (inserts before deletes) AFAIK, because indexes of the unique type would get angry.

arjansh:
@kaspersorensen , @LosD Is it possible that one of you takes another look at this PR and merges it if it is passes review?

LosD:
@arjansh I can take a look tonight, but... What's going on with the builds? All builds are dead, but none of them seem related to this. 2 are Neo4J, one is missing dependencies and one has a broken Java installer. :open_mouth:

arjansh:
@losd See my other PR (https://github.com/apache/metamodel/pull/231) for the Travis build problem.

LosD:
@arjan Ah, I'm blind. I'll merge that one first, then rerun this one :)

arjansh:
@losd If you merge the other one, then I can merge "master" into this PR's branch, which should trigger a new build.

LosD:
@arjan Sure, I just can't do it until I'm home tonight.

arjansh:
@LosD No need to hurry.

arjan:
:wave: I'm another @arjan ;-)

LosD:
I'm merging the other issue momentarily, but please fix these two issues as you're merging master:
- Missing a corresponding issue tracker issue
- There's no CHANGES.md update

arjansh:
@losd It's a bit late for me now, I'll do that tomorrow morning.

LosD:
Sure :)

elasticsearch/native/src/main/java/org/apache/metamodel/elasticsearch/nativeclient/ElasticSearchDataContextFactory.java
LosD: Is the name anything but a label? If it is, it probably needs to be configurable. Maybe it should be anyway, since MetaModel is just a library for the actual client. If anyone is interested in a label for their client (e.g. for logging purposes, connection lists, branding, whatever), I guess we should let them.

LosD: (OTOH, it's not something new, so it's probably out of the scope of this. Nevermind!)

elasticsearch/rest/src/main/java/org/apache/metamodel/elasticsearch/rest/ElasticSearchRestDataContext.java
```diff
@@ -133,21 +142,24 @@ public ElasticSearchRestDataContext(final ElasticSearchRestClient client, String
...
-                    logger.error("Unexpected error during detectTable for document type '{}'", documentType, e);
+                final Map<String, Object> properties = (Map<String, Object>) mappingConfiguration.get(ElasticSearchMetaData.PROPERTIES_KEY);
+
+                if (properties != null) {
+                    try {
+                        final SimpleTableDef table = detectTable(properties, tableName);
+                        result.add(table);
+                    } catch (Exception e) {
+                        logger
+                                .error("Unexpected error during detectTable for document mapping type '{}'", tableName,
```
LosD: Just cosmetic, but this looks rather odd. Is this really how it's supposed to be formatted?

elasticsearch/rest/src/main/java/org/apache/metamodel/elasticsearch/rest/ElasticSearchRestUpdateCallback.java
```diff
@@ -135,20 +140,32 @@ public void execute(final ActionRequest action) {
...
+            if (action instanceof BulkRequest) {
+                result = client.bulk((BulkRequest) action, RequestOptions.DEFAULT);
+            } else if (action instanceof IndexRequest) {
+                result = client.index((IndexRequest) action, RequestOptions.DEFAULT);
+            } else if (action instanceof DeleteRequest) {
+                result = client.delete((DeleteRequest) action, RequestOptions.DEFAULT);
+            } else if (action instanceof ClearScrollRequest) {
+                result = client.clearScroll((ClearScrollRequest) action, RequestOptions.DEFAULT);
+            } else if (action instanceof SearchScrollRequest) {
+                result = client.scroll((SearchScrollRequest) action, RequestOptions.DEFAULT);
```
LosD: That's... Rather nasty. It seems to do nothing, why is this needed?

LosD: Ah, I'm blind, it's of course not the same method called. Still looks terrible, though. :)

elasticsearch/rest/src/test/java/org/apache/metamodel/elasticsearch/rest/ElasticSearchRestDataContexFactoryIT.java
```diff
@@ -18,114 +18,122 @@
...
+        externalClient.indices().create(new CreateIndexRequest(INDEX_NAME), RequestOptions.DEFAULT);
 
-        final Map<String, Object> source = new LinkedHashMap<>();
-        source.put("mytext", "dummy");
+        final PutMappingRequest putMappingRequest = new PutMappingRequest(INDEX_NAME);
+        putMappingRequest
+                .source(ElasticSearchUtils
+                        .getMappingSource(new MutableTable(DEFAULT_TABLE_NAME, TableType.TABLE, new MutableSchema(
+                                INDEX_NAME), new MutableColumn("name", ColumnType.STRING), new MutableColumn("age",
+                                        ColumnType.INTEGER))));
```
LosD: More of these odd formattings. It kinda looks like it's broken up in the wrong order:
- "Oh, there's no space for .source, BREAK IT UP."
- "Oh, there's no space for .getMappingSource, BREAK IT UP."
- "Oh look, now there's space for .source again. Nah, I'll leave it there."

Oh well, maybe it's how it's supposed to be, but it sure looks weird.

LosD:
A few comments, but nothing serious. Of course, as I wrote, it's been a long time since I used Java, and it's been an even longer time since I played with ES, so I can't promise that I didn't miss anything.

elasticsearch/rest/src/main/java/org/apache/metamodel/elasticsearch/rest/ElasticSearchRestUpdateCallback.java
LosD: (I guess an adapter with overloaded methods would be more readable, but it would also be more code)

elasticsearch/native/src/main/java/org/apache/metamodel/elasticsearch/nativeclient/ElasticSearchDataContextFactory.java
arjansh: As you noticed, it was already there and it's part of the "native client", which is going to be deprecated as part of this Pull Request, so I'd rather not change it.

elasticsearch/rest/src/main/java/org/apache/metamodel/elasticsearch/rest/ElasticSearchRestDataContext.java
arjansh: On one line it's just over 120 characters, so my code formatter breaks it up into this, but it does look better on one line (and it's still under 130 characters on one line, so that's acceptable, so I'll manually put it on one line and make sure I don't format it).

LosD:
LGTM, merging! :)

PR #231, opened 2019-10-11
Travis CI build fails because it now uses xenial linux distribution

arjansh:
The last successful MetaModel build on Travis CI is job #1054 (https://travis-ci.org/apache/metamodel/jobs/594569177) , which still uses trusty Linux distribution. After that Travis CI apparently changed the default distribution to use to xenial and since then there are no more successful builds, e.g. see #1056 (https://travis-ci.org/apache/metamodel/jobs/595029156).

The openjdk8 and oraclejdk11 builds fail on neo4j module and the oraclejdk9 and oraclejdk10 builds fail because some of the used dependencies can't be resolved.

Others are encountering problems with neo4j now too, e.g. see https://travis-ci.community/t/neo4j-service-start-is-probably-broken-by-change-to-xenial-as-the-default-build-environment/5381.

I suggest to explicitly set the Linux distribution to use to trusty for now and investigate how we can solve these xenial distribution related problems on a separate branch and move to using that once we get it to work.

My first commit only adds a newline in the readme file, to illustrate the problem. I will add another commit to have Travis CI use the trusty distribution afterwards.


LosD:
I'll fix part of it, and skip for the other, but two issues with this PR: Missing a corresponding issue tracker issue, and there's no CHANGES.md update.

LosD:
Merged

PR #234, opened 2019-12-04
MM-82 Detect Column Types

GerardDellemann:
Solves the column type detection for [METAMODEL-82](https://issues.apache.org/jira/browse/METAMODEL-82). You can now configure that you want to detect the Column Types and set how many rows you want to scan.

I've tried to make everything backwards compatable. I've added 2 new test files. This resulted in also finding another minor bug in the current code: the number to string formatting was different between xls and xlsx when working with a dot to seperate thousands. This is fixed with the new method ExcelUtils#getNumericCellValueAsStringUsingBuildinFormat(...).


excel/src/main/java/org/apache/metamodel/excel/ExcelConfiguration.java
```diff
@@ -102,17 +115,39 @@ public boolean isSkipEmptyColumns() {
 		return skipEmptyColumns;
 	}
 
+    /**
+     * Defines if columns in the excel spreadsheet should be validated on datatypes while
```
kaspersorensen: Instead of "validated" we should probably write "detected". And we should also say what happens when this is `false` - then all columns will be of STRING type.

GerardDellemann:
> 
> 
> Really appreciate the backwards-compatible approach used here, and the fact that all existing tests are unaltered. My only nitpick is that there are some indentation differences in some of the files. For example `ExcelConfiguration`. But it does also look like the source files have inconsistent indentation so I can live with it, just a note for us that it looks like we need to reformat some files.

Yeah, I didn't apply any formatting on purpose, because then I usually have to undo a lot of changes it makes on code that I did not tough. I can try to apply that now.
I also think that this project's code in general really needs refactoring, because the current is hard to understand and messy right now. That should probably be done in another PR.

excel/src/main/java/org/apache/metamodel/excel/ExcelConfiguration.java
GerardDellemann: I've changed this JavaDoc. Do note that in order to stay backwards compatible it's either String or VarChar depending on the Column Name Line configuration. I try to explain the differences in the JavaDoc.
I think we should only return String soon. The current situation is hard to understand and very confusing.

excel/src/main/java/org/apache/metamodel/excel/DefaultSpreadsheetReaderDelegate.java
```diff
@@ -146,34 +157,196 @@ private MutableTable createTable(final Workbook wb, final Sheet sheet) {
...
+                return cell.getNumericCellValue() % 1 == 0 ? ColumnType.INTEGER : ColumnType.DOUBLE;
+            }
+        case BOOLEAN:
+            return ColumnType.BOOLEAN;
+        case FORMULA:
+            final FormulaEvaluator evaluator = cell
+            .getSheet()
+            .getWorkbook()
+            .getCreationHelper()
+            .createFormulaEvaluator();
```
arjansh: These lines are formatted in a really weird manner. Next to that, I think that for performance reasons, it's better to create the `FormulaEvaluator` only once and then reuse it, since it's tied to the `Workbook` and not `Cell` specific.

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -443,4 +578,14 @@ public static DataSet getDataSet(Workbook workbook, Sheet sheet, Table table, Ex
         final DataSet dataSet = new XlsDataSet(selectItems, workbook, rowIterator);
         return dataSet;
     }
+
+    private static String getNumericCellValueAsStringUsingBuildinFormat(final CellStyle cellStyle, final double cellValue) {
```
arjansh: Can you maybe rename this method to `getNumericCellValueAsString`? Or maybe just `convertToString`?

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -222,7 +222,62 @@ public static String getCellValue(Workbook wb, Cell cell) {
             } else {
                 // TODO: Consider not formatting it, but simple using
```
arjansh: Now with the new methodology implemented for converting the cell to String, I think you should also remove this "TODO".

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -222,7 +222,62 @@ public static String getCellValue(Workbook wb, Cell cell) {
...
+        default:
+            throw new IllegalStateException("Unknown cell type: " + cell.getCellType());
+        }
+
+        logger.debug("cell {} resolved to value: {}", cellCoordinate, result);
+
+        return result;
+    }
+
+    public static Object getCellValueAsObject(final Workbook wb, final Cell cell) {
```
arjansh: Can you make this method private and rename the `wb` parameter to `workbook`?

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -222,7 +222,62 @@ public static String getCellValue(Workbook wb, Cell cell) {
...
+        if (cell == null) {
+            return null;
+        }
+
+        final String cellCoordinate = "(" + cell.getRowIndex() + "," + cell.getColumnIndex() + ")";
```
arjansh: This `String` is only used inside logging statements, for performance reasons please use format objects when logging and add the `cell.getRowIndex()` and `cell.getColumnIndex()` as arguments to the logging method.

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -237,13 +292,34 @@ public static String getCellValue(Workbook wb, Cell cell) {
         return result;
     }
 
+    public static Object getCellValueChecked(final Workbook wb, final Cell cell, final ColumnType expectedColumnType) {
```
arjansh: Can you make this method private and rename the `wb` parameter to `workbook`?

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -237,13 +292,34 @@ public static String getCellValue(Workbook wb, Cell cell) {
...
+        final Object value = getCellValueAsObject(wb, cell);
+        if (value == null || value.getClass().equals(expectedColumnType.getJavaEquivalentClass())) {
+            return value;
+        }
+
+        if (!(value.getClass().equals(Integer.class) && expectedColumnType
+                .getJavaEquivalentClass()
+                .equals(Double.class)) && logger.isWarnEnabled()) {
+            // Don't log when a Integer value is in a Double column type
```
arjansh: Can you move this outside the if, it applies to the if, and in this manner, it's inside the scope of the resolved if statement. Also can you change "a Integer" to "an Integer"?

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -237,13 +292,34 @@ public static String getCellValue(Workbook wb, Cell cell) {
...
+            final String cellCoordinate = "(" + cell.getRowIndex() + "," + cell.getColumnIndex() + ")";
```
arjansh: You only use this `String` when logging a message, already using a format, please remove this String and add two arguments to the log method instead.

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -237,13 +292,34 @@ public static String getCellValue(Workbook wb, Cell cell) {
...
+            logger.warn(warning);
+        }
+        return null;
+    }
+
     private static String getFormulaCellValue(Workbook wb, Cell cell) {
         // first try with a cached/precalculated value
         try {
             double numericCellValue = cell.getNumericCellValue();
             // TODO: Consider not formatting it, but simple using
```
arjansh: Can you also remove this "TODO"?

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -281,6 +357,57 @@ private static String getFormulaCellValue(Workbook wb, Cell cell) {
...
+
+            // calculates the formula and puts it's value back into the cell
+            final Cell evaluatedCell = evaluator.evaluateInCell(cell);
+
+            return getCellValueAsObject(wb, evaluatedCell);
+        } catch (final RuntimeException e) {
+            logger
+            .warn("Exception occurred while evaluating formula at position ({},{}): {}", cell.getRowIndex(),
+                    cell.getColumnIndex(), e.getMessage());
+            // Some exceptions we simply log - result will be then be the actual formula
```
arjansh: I think it's a bit of overkill to log an extra error next to the warning which already contains the "message" of the thrown exception.

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -281,6 +357,57 @@ private static String getFormulaCellValue(Workbook wb, Cell cell) {
...
+            } else {
+                logger.error("Unexpected exception occurred while evaluating cell formula: " + cell, e);
+            }
+        }
+
+        // last resort: return the string formula
+        return cell.getCellFormula();
+    }
+
+    private static Number getIntegerOrDoubleValueFromDouble(final double value) {
```
arjansh: I really don't like the name of this issue, maybe name it `convertToNumber` or `getDoubleAsNumber` instead.

excel/pom.xml
```diff
@@ -40,7 +40,7 @@ under the License.
 		<dependency>
 			<groupId>org.apache.poi</groupId>
 			<artifactId>poi-ooxml</artifactId>
-			<version>4.0.1</version>
+			<version>4.1.0</version>
```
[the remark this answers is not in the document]
GerardDellemann: done

excel/src/main/java/org/apache/metamodel/excel/DefaultSpreadsheetReaderDelegate.java
GerardDellemann: done, turned it into a class member, not sure if that goes to far?

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
GerardDellemann: Done

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
GerardDellemann: done

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
GerardDellemann: done

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
GerardDellemann: done

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
GerardDellemann: done

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
GerardDellemann: done

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
GerardDellemann: ok, removed the extra warning and refactored the catching

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
GerardDellemann: done

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
GerardDellemann: done

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
GerardDellemann: Actually, it's being used in so many places. I made a static method for this.

excel/src/main/java/org/apache/metamodel/excel/DefaultSpreadsheetReaderDelegate.java
arjansh: I think that goes to far, because the FormulaEvaluator is tied to the Workbook, which itself is not stored as a member class.

GerardDellemann: ok, I get the workbook from the cell now

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -232,18 +279,36 @@ public static String getCellValue(Workbook wb, Cell cell) {
             throw new IllegalStateException("Unknown cell type: " + cell.getCellType());
         }
 
-        logger.debug("cell {} resolved to value: {}", cellCoordinate, result);
+        logger.debug("cell {} resolved to value: {}", getCellCoordinates(cell), result);
 
         return result;
     }
 
+    public static Object getCellValueChecked(final Workbook workbook, final Cell cell,
```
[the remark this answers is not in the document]
GerardDellemann: done

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -253,8 +318,9 @@ private static String getFormulaCellValue(Workbook wb, Cell cell) {
         // evaluate cell first, if possible
         try {
             if (logger.isInfoEnabled()) {
-                logger.info("cell({},{}) is a formula. Attempting to evaluate: {}",
-                        new Object[] { cell.getRowIndex(), cell.getColumnIndex(), cell.getCellFormula() });
+                logger
```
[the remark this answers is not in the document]
GerardDellemann: ok

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -281,6 +348,51 @@ private static String getFormulaCellValue(Workbook wb, Cell cell) {
...
+            }
+
+            final FormulaEvaluator evaluator = workbook.getCreationHelper().createFormulaEvaluator();
+
+            // calculates the formula and puts it's value back into the cell
+            final Cell evaluatedCell = evaluator.evaluateInCell(cell);
+
+            return getCellValueAsObject(workbook, evaluatedCell);
+        } catch (final FormulaParseException e) {
+            logger.error("Parse exception occurred while evaluating cell formula: " + cell, e);
```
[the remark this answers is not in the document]
GerardDellemann: ok, I've also changed it in the getFormulaCellValue(...) method then.

excel/src/main/java/org/apache/metamodel/excel/DefaultSpreadsheetReaderDelegate.java
arjansh: Now we're back to the situation where I first commented on. I think there are three options to solve this:

1. Introduce a `LazyRef<FormulaEvaluator>` variable which you declare, for example, in `detectColumnTypes(Row, Iterator<Row>, ColumnType[] columnTypes`, and then pass on to the methods invoked by that method, which gets the FormulaEvaluator only the first time it's needed and reuses that for the rest of the process.
2. Introduce an inner class called `ColumnTypeScanner` (or something similar), which you instantiate on line 137 (inside the `createTable(Workbook, Sheet)` method and where you pass the Sheet into the constructor. This `ColumnTypeScanner` then contains all column type scanner related methods and also has a `FormulaEvaluator` field, which is set once the `FormulaEvaluator` is used for the first time, and after that it can be reused.
3. Same as option 2, but instead of using an inner class, create a normal class. This makes reusing some of the `DefaultSpreadsheetReaderDelegate` class's methods a bit harder.

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -232,18 +230,93 @@ public static String getCellValue(Workbook wb, Cell cell) {
...
+        final Object value = getCellValueAsObject(workbook, cell);
+        if (value == null || value.getClass().equals(expectedColumnType.getJavaEquivalentClass())) {
+            return value;
+        }
+
+        // Don't log when an Integer value is in a Double column type
+        if (!(value.getClass().equals(Integer.class) && expectedColumnType
+                .getJavaEquivalentClass()
+                .equals(Double.class)) && logger.isWarnEnabled()) {
+            final String warning = String
```
arjansh: Why assign this to a String outside the `logger.warn` invocation, which has built in logic for doing stuff like this?

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -264,12 +338,13 @@ private static String getFormulaCellValue(Workbook wb, Cell cell) {
...
-            logger.warn("Exception occurred while evaluating formula at position ({},{}): {}",
-                    new Object[] { cell.getRowIndex(), cell.getColumnIndex(), e.getMessage() });
+            logger
+                    .warn("Exception occurred while evaluating formula at position ({},{}): {}", cell.getRowIndex(),
+                            cell.getColumnIndex(), e.getMessage());
             // Some exceptions we simply log - result will be then be the
             // actual formula
             if (e instanceof FormulaParseException) {
-                logger.error("Parse exception occurred while evaluating cell formula: " + cell, e);
+                logger.warn("Parse exception occurred while evaluating cell formula: " + cell, e);
```
arjansh: I would not touch this code, unless you really improve it, which would be to get rid of the if (e instanceof ...Exception)  construction and just use normal catches instead, a bit like this:
```
        } catch (FormulaParseException e) {
            logger.warn("Parse exception occurred while evaluating cell formula: " + cell, e);
        } catch (IllegalArgumentException e) {
            logger.warn("Illegal formula argument occurred while evaluating cell formula: " + cell, e);
        } catch (RuntimeException e) {
            logger
                    .warn("Exception occurred while evaluating formula at position ({},{}): {}", cell.getRowIndex(),
                            cell.getColumnIndex(), e.getMessage());
        }
```

But in general, given the content of the different logged messages, I would just catch the runtime exception, and get rid of the complete if then else construction with all the different log message, which, to me, don't add any real value as they won't make it any more clear where what went wrong in what manner.

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -281,6 +356,51 @@ private static String getFormulaCellValue(Workbook wb, Cell cell) {
...
+        try {
+            return getDoubleAsNumber(cell.getNumericCellValue());
+        } catch (final Exception e) {
+            if (logger.isInfoEnabled()) {
+                logger.info("Failed to fetch cached/precalculated formula value of cell: " + cell, e);
+            }
+        }
+
+        // evaluate cell first, if possible
+        try {
```
arjansh: This piece of code looks like a carbon copy of line 326-348, with the only difference that in those lines, on line 339 you invoke `getCellValue`, where here on line 377 you invoke `getCallValueAsObject`,  I would propose you refactor this to get rid of the duplicated boiler plate code around these invocations (so it's no longer duplicated, just available once).

excel/src/main/java/org/apache/metamodel/excel/DefaultSpreadsheetReaderDelegate.java
GerardDellemann: I like the ColumnTypeScanner as Inner Class. I've positioned the instantion in a different place, but I think you'll probably like this better too :).

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -195,34 +192,35 @@ public static String getCellValue(Workbook wb, Cell cell) {
...
+                        .debug("Getting error code for ({},{}) failed!: {}", cell.getRowIndex(), cell.getColumnIndex(),
+                                e.getMessage());
                 if (cell instanceof XSSFCell) {
                     // hack to get error string, which is available
                     String value = ((XSSFCell) cell).getErrorCellString();
                     errorResult = value;
                 } else {
-                    logger.error("Couldn't handle unexpected error scenario in cell: " + cellCoordinate, e);
+                    logger
+                            .error("Couldn't handle unexpected error scenario in cell: (" + cell.getRowIndex() + ","
```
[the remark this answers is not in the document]
GerardDellemann: ok

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -232,18 +230,93 @@ public static String getCellValue(Workbook wb, Cell cell) {
...
+        switch (cell.getCellType()) {
+        case BLANK:
+        case _NONE:
+            result = null;
+            break;
+        case BOOLEAN:
+            result = Boolean.valueOf(cell.getBooleanCellValue());
+            break;
+        case ERROR:
+            String errorResult;
```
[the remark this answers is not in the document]
GerardDellemann: Good one 👍

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
GerardDellemann: yeah true, changed

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
GerardDellemann: ok, then I also apply that change to getFormulaCellValueAsObject(...)

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
GerardDellemann: ok, I've put it into a method

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -232,53 +249,119 @@ public static String getCellValue(Workbook wb, Cell cell) {
...
         }
 
-        logger.debug("cell {} resolved to value: {}", cellCoordinate, result);
+        logger.debug("cell ({},{}) resolved to value: {}", cell.getRowIndex(), cell.getColumnIndex(), result);
 
         return result;
     }
 
-    private static String getFormulaCellValue(Workbook wb, Cell cell) {
+    private static Object getErrorResult(final Cell cell) {
```
arjansh: This method always returns a `String` (unless it throws an Exception of course), so why not change the return signature to `String`?

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -232,53 +249,119 @@ public static String getCellValue(Workbook wb, Cell cell) {
...
+        String errorResult;
```
arjansh: You don't need a variable to put the "errorResult" in, you can instead just return it when you get it.

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -232,53 +249,119 @@ public static String getCellValue(Workbook wb, Cell cell) {
...
+        final Cell evaluatedCell = getEvaluatedCell(workbook, cell);
+        if (evaluatedCell != null) {
+            return getCellValueAsObject(workbook, evaluatedCell);
+        } else {
+            // last resort: return the string formula
+            return cell.getCellFormula();
+        }
+    }
+
+    private static Cell getEvaluatedCell(final Workbook workbook, final Cell cell) {
```
arjansh: Consider renaming this to `evaluateCell`.

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
GerardDellemann: true, I'll change

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
GerardDellemann: sure

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
GerardDellemann: sure

excel/src/main/java/org/apache/metamodel/excel/ExcelUtils.java
```diff
@@ -189,40 +185,61 @@ public static String getCellValue(Workbook wb, Cell cell) {
...
-                    // hack to get error string, which is available
-                    String value = ((XSSFCell) cell).getErrorCellString();
-                    errorResult = value;
-                } else {
-                    logger.error("Couldn't handle unexpected error scenario in cell: " + cellCoordinate, e);
-                    throw e;
-                }
-            }
-            result = errorResult;
+            result = (String) getErrorResult(cell);
```
arjansh: You now no longer need to cast the return type of `getErrorResult(Cell)` to `String`.

GerardDellemann: correct

PR #236, opened 2019-12-11
METAMODEL-82 - Fix integers in double columns

GerardDellemann:
Closes: [METAMODEL-82](https://issues.apache.org/jira/browse/METAMODEL-82)

I found a bug in my implementation [MM-PR234](https://github.com/apache/metamodel/pull/234): When Integer values were in a Double column, then these values were set to NULL when querying the result.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 ### Apache MetaModel _WIP_
+ * [METAMODEL-82] - Detect Column Types with Excel datastores
  * [METAMODEL-1221] - Upgrade to Elasticsearch 7.3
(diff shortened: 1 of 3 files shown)


arjansh:
@GerardDellemann Can you also update the CHANGES.md file (in the root folder of the MetaModel project), by adding METAMODEL-82?

GerardDellemann:
> 
> 
> @GerardDellemann Can you also update the CHANGES.md file (in the root folder of the MetaModel project), by adding METAMODEL-82?

yeah forgot that, it's done

PR #237, opened 2019-12-15
METAMODEL-1201 Remove guava dependency in package jdbc

RatishBansal:
The actual code using guava class CharMatcher was removed in PR https://github.com/apache/metamodel/pull/195 (METAMODEL-1205)
But the guava dependency is still part of POM.
Since the dependency is no more required removing the same


arjansh:
Looks good to me!

PR #239, opened 2020-03-21
METAMODEL-1224: Fixed PostgreSQL compatibility (upgrading driver+tests)

kaspersorensen:
Reproduced the issue in METAMODEL-1224 and tried upgrading the driver. This fixed the issue it seems, so here's an updated test since a few error messages have been updated from the driver's side as well.

Also, with the old driver the underlying issue seems to be that the schema name comes out as blank. I'm not sure why, but probably just a driver-and-database version conflict. I updated the code to not generate queries with blank schema names, but in those cases follow the routes of not prefixing tables with schema names.

### DIFF
--- a/core/src/main/java/org/apache/metamodel/query/FromItem.java
+++ b/core/src/main/java/org/apache/metamodel/query/FromItem.java
@@ -222,7 +222,8 @@ public class FromItem extends BaseObject implements QueryItem, Cloneable {
         }
         StringBuilder sb = new StringBuilder();
         if (_table != null) {
-            if (_table.getSchema() != null && _table.getSchema().getName() != null && !_table.getSchema().getName().isEmpty()) {
+            if (_table.getSchema() != null && _table.getSchema().getName() != null
+                    && !_table.getSchema().getName().isEmpty()) {
                 sb.append(_table.getSchema().getName());
                 sb.append('.');
             }
--- a/core/src/main/java/org/apache/metamodel/query/FromItem.java
+++ b/core/src/main/java/org/apache/metamodel/query/FromItem.java
@@ -222,7 +222,7 @@ public class FromItem extends BaseObject implements QueryItem, Cloneable {
         }
         StringBuilder sb = new StringBuilder();
         if (_table != null) {
-            if (_table.getSchema() != null && _table.getSchema().getName() != null) {
+            if (_table.getSchema() != null && _table.getSchema().getName() != null && !_table.getSchema().getName().isEmpty()) {
                 sb.append(_table.getSchema().getName());
                 sb.append('.');
             }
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/metamodel/query/FromItem.java
```diff
@@ -222,7 +222,7 @@ public String toStringNoAlias(boolean includeSchemaInColumnPaths) {
         }
         StringBuilder sb = new StringBuilder();
         if (_table != null) {
-            if (_table.getSchema() != null && _table.getSchema().getName() != null) {
+            if (_table.getSchema() != null && _table.getSchema().getName() != null && !_table.getSchema().getName().isEmpty()) {
```
arjansh: I'm not sure what the code style conventions are for MetaModel, but this line is really long. Maybe change it to:

```
            if (_table.getSchema() != null && _table.getSchema().getName() != null 
                    && !_table.getSchema().getName().isEmpty()) {
```

jdbc/src/test/java/org/apache/metamodel/jdbc/dialects/PostgresqlQueryRewriterTest.java
```diff
@@ -46,4 +52,18 @@ public void testInsertNullMap() throws SQLException {
...
+    
+    @Test
+    public void testApproximateCountQueryAndBlankSchemaName() {
+        final SelectItem selectItem = SelectItem.getCountAllItem();
+        selectItem.setFunctionApproximationAllowed(true);
+        final Query query = new Query();
+        query.select(selectItem);
+        query.from(new MutableTable("tbl").setSchema(new MutableSchema("")));
+        assertEquals("SELECT APPROXIMATE COUNT(*) FROM tbl", query.toSql());
+
```
arjansh: Again, not sure about the code style conventions, but lines 55 and 64 contain trailing spaces (which for some reason aren't visible in the browser view), can you remove these?

jdbc/src/test/java/org/apache/metamodel/jdbc/integrationtests/PostgresqlTest.java
```diff
@@ -502,15 +502,22 @@ public void run(UpdateCallback cb) {
...
+        try {
+            
+            final Optional<Integer> insertedRows = updateSummary.getInsertedRows();
+            assertTrue(insertedRows.isPresent());
+            assertEquals(2, insertedRows.get().intValue());
+            
+            final Optional<Iterable<Object>> generatedKeys = updateSummary.getGeneratedKeys();
+            assertTrue(generatedKeys.isPresent());
+            assertEquals("[1, 2]", generatedKeys.get().toString());
+
```
arjansh: Lines 507, 511 and 515 contain trailing spaces, can you remove these?

kaspersorensen:
Updated as per PR review remarks

PR #240, opened 2020-04-02
METAMODEL-1225 Add ColumnTypingStrategies and use them in CSV module.

LiveSatomi:
There are two strategies for column typing, all text and manual.
The CSV module can use manual typing to filter based on numerical
greater than and less than operations.

### DIFF
--- a/core/src/main/java/org/apache/metamodel/schema/typing/ColumnTypingContext.java
+++ b/core/src/main/java/org/apache/metamodel/schema/typing/ColumnTypingContext.java
@@ -1,0 +1,52 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.metamodel.schema.typing;
+
+import org.apache.metamodel.schema.ColumnType;
+import org.apache.metamodel.schema.Table;
+
+/**
+ * Defines the context for configuring the type for a single column in a
+ * {@link ColumnTypingStrategy} session.
+ */
+public interface ColumnTypingContext {
+
+    /**
+     * Gets the index of the column being configured.
+     *
+     * @return the column index
+     */
+    int getColumnIndex();
+
+    /**
+     * Gets the {@link Table} that the column is to pertain to. If the table is
+     * not yet available then this may return null.
+     *
+     * @return the associated table
+     */
+    Table getTable();
+
+    /**
+     * Gets the intrinsic column type, if this is defined in the datastore
+     * itself. This may be in the form of a header or such.
+     *
+     * @return the column type representing the datastore's column type
+     */
+    ColumnType getIntrinsicColumnType();
+}
--- a/core/src/main/java/org/apache/metamodel/schema/typing/ColumnTypingContextImpl.java
+++ b/core/src/main/java/org/apache/metamodel/schema/typing/ColumnTypingContextImpl.java
@@ -1,0 +1,76 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.metamodel.schema.typing;
+
+import org.apache.metamodel.schema.ColumnType;
+import org.apache.metamodel.schema.Table;
+
+
+/**
+ * An implementation of {@link ColumnTypingContext} that holds necessary context about the column being configured.
+ */
+public class ColumnTypingContextImpl implements ColumnTypingContext {
+
+    private final int columnIndex;
+
+    private final Table table;
+
+    private final ColumnType intrinsicColumnType;
+
+
+    /**
+     * Creates a context to conifgure a column for a specific table.
+     * @param table               The table that contains the column
+     * @param intrinsicColumnType The column type that represents the type configured in the datastore
+     * @param columnIndex         the index in the table of the column being configured.
+     */
+    public ColumnTypingContextImpl( Table table, ColumnType intrinsicColumnType, int columnIndex ) {
+        this.table = table;
+        this.intrinsicColumnType = intrinsicColumnType;
+        this.columnIndex = columnIndex;
+    }
+
+
+    /**
+     * Creates a context a column to be configured.
+     * @param columnIndex the index in the table of the column being configured.
+     */
+    public ColumnTypingContextImpl( int columnIndex ) {
+        this( null, null, columnIndex );
+    }
+
+
+    @Override
+    public int getColumnIndex() {
+        return columnIndex;
+    }
+
+
+    @Override
+    public Table getTable() {
+        return table;
+    }
+
+
+    @Override
+    public ColumnType getIntrinsicColumnType() {
+        return intrinsicColumnType;
+    }
+
+}
--- a/core/src/main/java/org/apache/metamodel/schema/typing/ColumnTypingSession.java
+++ b/core/src/main/java/org/apache/metamodel/schema/typing/ColumnTypingSession.java
@@ -1,0 +1,48 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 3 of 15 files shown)


LiveSatomi:
As mentioned in the JIRA comment, let me know if there is too much duplication between the typing package and naming package.
Also, do we need to handle trying to update the datastore with a data type that is different than the one you configured? I haven't tested but it seems to me you can now make a write to csv data stores that you can't read back (unless you reconfigured the datacontext to use string types which will always work)

csv/src/test/java/org/apache/metamodel/csv/CsvDataContextTest.java
```diff
@@ -492,7 +490,7 @@ public void testMaterializeTable() throws Exception {
     public void testAlternativeDelimitors() throws Exception {
         File file = new File("src/test/resources/csv_semicolon_singlequote.csv");
         CsvDataContext dc = new CsvDataContext(file, semicolonConfiguration);
-        Table table = dc.getSchemas().get(0).getTables().get(0);
```
LiveSatomi, line 495: This tripped me up for a while, it turns out _getSchemas().get(0)_ returns the schema for information_schema while _getDefaultSchema()_ returns the schema for specified resource. Now that the typing on the schema matters for reads dataSet.next() includes a cast which was failing.

csv/src/main/java/org/apache/metamodel/csv/CsvDataContextFactory.java
```diff
@@ -59,7 +62,16 @@ public DataContext create(DataContextProperties properties, ResourceFactoryRegis
             columnNamingStrategy = new CustomColumnNamingStrategy(columnNames);
         }
 
-        final CsvConfiguration configuration = new CsvConfiguration(columnNameLineNumber, columnNamingStrategy,
+        final ColumnTypingStrategy columnTypingStrategy;
+        if (properties.getTableDefs() == null) {
+            columnTypingStrategy = null;
+        } else {
+            final SimpleTableDef firstTable = properties.getTableDefs()[0];
```
kaspersorensen, line 69: Nice touch to use table defs in the factory.

core/src/main/java/org/apache/metamodel/schema/typing/ColumnTypingContext.java
```diff
@@ -0,0 +1,52 @@
...
+     */
+    Table getTable();
+
+    /**
+     * Gets the intrinsic column type, if this is defined in the datastore
+     * itself. This may be in the form of a header or such.
+     *
+     * @return the column type representing the datastore's column type
+     */
+    ColumnType getIntrinsicColumnType();
```
kaspersorensen, line 51: I don't much see the need for this, since we always have "String" as the intrinsic column type for CSVs. I'm guessing the method is here to make an equivalent to the ColumnNamingStrategy class, but in that case we might have a column name from the file itself.

core/src/main/java/org/apache/metamodel/schema/typing/CustomColumnTypingStrategy.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private final List<ColumnType> columnTypes;
+
+
+    /**
+     * Creates the strategy based on the provided types.
+     *
+     * @param columnTypes a list of column types to be applied to a table.
+     */
+    public CustomColumnTypingStrategy( List<ColumnType> columnTypes ) {
```
kaspersorensen, line 42: I suggest also having a constructor that takes a SimpleTableDef here (since that representation of core table metadata is widely used), and then use that in the CsvDataContextFactory.

core/src/main/java/org/apache/metamodel/schema/typing/ColumnTypingContext.java
LiveSatomi: Maybe I don't understand what is meant by intrinsic, but I thought this method was to try to get the data type in an automated/best-guess way such as using information_schema for RDBMSs or checking the schema of a JSON file. For CSV yes I think it will always be STRING. Maybe there is already some other way to deal with the other data sources.

core/src/main/java/org/apache/metamodel/schema/typing/ColumnTypingSession.java
```diff
@@ -0,0 +1,48 @@
...
+     *            column index, intrinsic type etc. if available.
+     * @return the type to provide to the column.
+     */
+    public ColumnType getNextColumnType( ColumnTypingContext ctx );
+
+    /**
+     * Ends the column typing session.
+     */
+    @Override
+    void close();
```
arjansh, line 47: Since the `close` method is already defined on the `Closeable` interface, it's not necessary to define it here again. You could change this to a default implementation of the `close` method, i.e.:
```
    @Override
    default void close() {
    }
```

In this manner you wouldn't have to implement it on the anonymous inner classes returned by the `startColumnTypingSession` on the `CustomColumnTypingStrategy` and `DefaultColumnTypingStrategy` classes.

core/src/main/java/org/apache/metamodel/schema/typing/ColumnTypingStrategies.java
```diff
@@ -0,0 +1,67 @@
...
+    private ColumnTypingStrategies() {
+    }
+
+
+    /**
+     * The default strategy assumes all column types are text.
+     *
+     * @return a strategy that defaults to text
+     */
+    public static ColumnTypingStrategy defaultStrategy() {
```
arjansh, line 42: Java method names are by convention preferably verbs, so then this should be `getDefaultStrategy` and the ones below should be `getCustomStrategy`.

core/src/test/java/org/apache/metamodel/schema/typing/CustomColumnTypingStrategyTest.java
```diff
@@ -0,0 +1,36 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class CustomColumnTypingStrategyTest {
+
+    @Test
+    public void testTypeConfiguration() throws Exception {
+        ColumnTypingStrategy strategy = ColumnTypingStrategies.customTypes(ColumnType.STRING, ColumnType.NUMBER);
+        try ( final ColumnTypingSession session = strategy.startColumnTypingSession() ) {
```
arjansh, line 31: In a try with resources statement the resource is implicitly final, so here you can remove the `final` modifier.

csv/src/main/java/org/apache/metamodel/csv/CsvDataContextFactory.java
```diff
@@ -22,8 +22,11 @@
 import org.apache.metamodel.factory.AbstractDataContextFactory;
 import org.apache.metamodel.factory.DataContextProperties;
 import org.apache.metamodel.factory.ResourceFactoryRegistry;
+import org.apache.metamodel.schema.ColumnType;
```
arjansh, line 25: The import org.apache.metamodel.schema.ColumnType is never used.

csv/src/test/java/org/apache/metamodel/csv/CsvDataContextTest.java
```diff
@@ -833,8 +831,14 @@ public void testCustomColumnNames() throws Exception {
         final String thirdColumnName = "third";
         final String fourthColumnName = "fourth";
 
+        final ColumnType firstColumnType = ColumnType.STRING;
```
arjansh, line 834: I don't think it adds real value to put these constants into the xxxColumnType variables, especially since they're only used once. It doesn't make the code more understandable or more elegant.

csv/src/test/java/org/apache/metamodel/csv/SingleLineCsvDataSetTest.java
```diff
@@ -64,4 +70,84 @@ public void testMalformedLineParsing() throws Exception {
         assertFalse(ds.next());
         ds.close();
     }
-}
\ No newline at end of file
+
+    public void testCustomTyping() throws Exception {
+        ColumnType columnType1 = ColumnType.DATE;
```
arjansh, line 75: Same here as above. In my opinion just using the constant is more clear than adding it to a variable which doesn't add a meaning to it.

core/src/main/java/org/apache/metamodel/schema/typing/ColumnTypingContextImpl.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private final Table table;
+
+
+    /**
+     * Creates a context to conifgure a column for a specific table.
+     * @param table               The table that contains the column
+     * @param columnIndex         the index in the table of the column being configured.
+     */
+    public ColumnTypingContextImpl( Table table, int columnIndex ) {
```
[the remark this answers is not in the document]
LiveSatomi: I will go back and double check the space padding in the rest of the files. My work uses different formatting rules and I'm not used to handling multiple rulesets in my IDE.

I agree with making parameters final when possible. Does this also apply to parameters of lambda functions? for example an instance of the ColumnTypingSession after adding the default implementation to close:
```
        return (final ColumnTypingContext ctx) -> {
            if ( iterator.hasNext() ) {
                return iterator.next();
            }
            return null;
        };
```
or 
```
        return ctx -> {
            if ( iterator.hasNext() ) {
                return iterator.next();
            }
            return null;
        };
```

Personally I treat parameters as final whether they are marked as such or not, and I don't see any final lambda parameters in the project so far.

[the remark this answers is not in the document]
arjansh: You make a valid point, because with the second example you're able to assign a new Object (if you were able to instantiate `ColumnTypingContext`) to the `ctx` parameter. Even though I wouldn't consider doing that.

It's just that in practice I always see the second example (the one you implemented), so for lambdas let's not start adding finals (and types) to the formal parameters now.

arjansh:
@kaspersorensen You had some comments on this pull requests, which mark it with "changes requested". Have these been addressed? If so, can we merge the pull request?

kaspersorensen:
Sorry guys @LiveSatomi  @arjansh , had been buried in work for a bit. Gave this my approval now :-)

csv/src/main/java/org/apache/metamodel/csv/CsvConfiguration.java
```diff
@@ -80,13 +83,15 @@ public CsvConfiguration(int columnNameLineNumber, String encoding, char separato
...
-                multilineValues);
+        this(columnNameLineNumber, null, null, encoding, separatorChar, quoteChar, escapeChar,
+                failOnInconsistentRowLength, multilineValues);
     }
 
-    public CsvConfiguration(int columnNameLineNumber, ColumnNamingStrategy columnNamingStrategy, String encoding,
-            char separatorChar, char quoteChar, char escapeChar, boolean failOnInconsistentRowLength,
-            boolean multilineValues) {
+
+    public CsvConfiguration(int columnNameLineNumber, ColumnNamingStrategy columnNamingStrategy,
```
arjansh, line 91: By replacing the old public constructor with this constructor, compile problems can occur in other projects using this library. Can you please retain the old constructor and have it invoke this constructor with a `null` value for the `ColumnTypeStrategy` parameter?

PR #241, opened 2020-06-30
(METAMODEL-1226) Metadata for numeric data types does not include num…

tkingsbury42:
https://issues.apache.org/jira/browse/METAMODEL-1226

### DIFF
--- a/core/src/main/java/org/apache/metamodel/schema/Column.java
+++ b/core/src/main/java/org/apache/metamodel/schema/Column.java
@@ -86,6 +86,14 @@ public interface Column extends Comparable<Column>, Serializable, NamedStructure
     public Integer getColumnSize();
 
     /**
+     * Gets the number of decimal digits this column.
+     *
+     * @return the number of decimal digits for this column or null if the size is not
+     *         determined or known.
+     */
+    public Integer getDecimalDigits();
+
+    /**
      * Gets the native type of this column. A native type is the name of the
      * data type as defined in the datastore.
      * 
--- a/core/src/main/java/org/apache/metamodel/schema/Column.java
+++ b/core/src/main/java/org/apache/metamodel/schema/Column.java
@@ -86,7 +86,7 @@ public interface Column extends Comparable<Column>, Serializable, NamedStructure
     public Integer getColumnSize();
 
     /**
-     * Gets the number of decimal digits this column.
+     * Gets the number of decimal digits for this column.
      *
      * @return the number of decimal digits for this column or null if the size is not
      *         determined or known.
--- a/core/src/main/java/org/apache/metamodel/schema/ImmutableColumn.java
+++ b/core/src/main/java/org/apache/metamodel/schema/ImmutableColumn.java
@@ -73,18 +73,7 @@ public final class ImmutableColumn extends AbstractColumn implements Serializabl
      */
     public ImmutableColumn(String name, ColumnType type, Table table, int columnNumber, Integer columnSize,
                            String nativeType, Boolean nullable, String remarks, boolean indexed, String quote, boolean primaryKey) {
-        this.name = name;
-        this.type = type;
-        this.table = table;
-        this.columnNumber = columnNumber;
-        this.columnSize = columnSize;
-        this.decimalDigits = null;
-        this.nativeType = nativeType;
-        this.nullable = nullable;
-        this.remarks = remarks;
-        this.indexed = indexed;
-        this.quote = quote;
-        this.primaryKey = primaryKey;
+        this(name, type, table, columnNumber, columnSize, null, nativeType, nullable, remarks, indexed, quote, primaryKey);
     }
 
     /**
--- a/core/src/main/java/org/apache/metamodel/schema/ImmutableColumn.java
+++ b/core/src/main/java/org/apache/metamodel/schema/ImmutableColumn.java
@@ -38,6 +38,7 @@ public final class ImmutableColumn extends AbstractColumn implements Serializabl
     private final Boolean nullable;
     private final String remarks;
     private final Integer columnSize;
+    private final Integer decimalDigits;
     private final String nativeType;
     private final boolean indexed;
     private final boolean primaryKey;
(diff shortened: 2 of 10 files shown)


core/src/main/java/org/apache/metamodel/schema/Column.java
```diff
@@ -85,6 +85,14 @@
      */
     public Integer getColumnSize();
 
+    /**
+     * Gets the number of decimal digits this column.
```
kaspersorensen, line 89: Nit-pick: Missing a "for" (this column) in this sentence.

core/src/main/java/org/apache/metamodel/schema/ImmutableColumn.java
```diff
@@ -144,6 +147,9 @@ public Integer getColumnSize() {
         return columnSize;
     }
 
+    @Override
+    public Integer getDecimalDigits() { return decimalDigits; }
```
kaspersorensen, line 151: Styling-wise we prefer a separate line for the body of the method.

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcMetadataLoader.java
```diff
@@ -383,6 +343,52 @@ private void loadColumns(JdbcTable table, DatabaseMetaData metaData) {
...
+    private void processColumn(JdbcTable table, boolean convertLobs, ResultSet rs, int columnNumber) throws SQLException {
+        final String columnName = rs.getString(4);
+        if (_identifierQuoteString == null && new StringTokenizer(columnName).countTokens() > 1) {
+            logger.warn("column name contains whitespace: \"" + columnName + "\".");
+        }
+
+        final int jdbcType = rs.getInt(5);
+        final String nativeType = rs.getString(6);
+        final Integer columnSize = rs.getInt(7);
+        Integer decimalDigits = rs.getInt(9);
```
kaspersorensen, line 355: We should add `final` to this one.

core/src/main/java/org/apache/metamodel/schema/ImmutableColumn.java
```diff
@@ -71,12 +72,13 @@
      *            whether the column is a primary key or not
      */
     public ImmutableColumn(String name, ColumnType type, Table table, int columnNumber, Integer columnSize,
-            String nativeType, Boolean nullable, String remarks, boolean indexed, String quote, boolean primaryKey) {
+                           Integer decimalDigits, String nativeType, Boolean nullable, String remarks, boolean indexed, String quote, boolean primaryKey) {
```
arjansh, line 75: By adding a parameter to this constructor, the code will cause compilation errors in other project using this constructor. Can you please add an extra constructor with the signature of the original version of this constructor, which invokes this constructor with a default value for `decimalDigits`? In that manner you also don't have to change all the unit test code.

core/src/main/java/org/apache/metamodel/schema/MutableColumn.java
```diff
@@ -70,9 +71,10 @@ public MutableColumn(String name, ColumnType type, Table table, int columnNumber
     }
 
     public MutableColumn(String name, ColumnType type, Table table, int columnNumber, Integer columnSize,
-            String nativeType, Boolean nullable, String remarks, boolean indexed, String quote) {
+                         Integer decimalDigits, String nativeType, Boolean nullable, String remarks, boolean indexed, String quote) {
```
arjansh, line 74: Same comment as for the `ImmutableColumn` constructor change, by adding a parameter to this constructor, the code will cause compilation errors in other project using this constructor. Can you please add an extra constructor with the signature of the original version of this constructor, which invokes this constructor with a default value for `decimalDigits`?

tkingsbury42: Fixed!

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcMetadataLoader.java
tkingsbury42: Fixed!

core/src/main/java/org/apache/metamodel/schema/ImmutableColumn.java
```diff
@@ -38,12 +38,55 @@
...
+     * @param indexed
+     *            whether the column is indexed or not
+     * @param quote
+     *            the quote character(s) of the column
+     * @param primaryKey
+     *            whether the column is a primary key or not
+     */
+    public ImmutableColumn(String name, ColumnType type, Table table, int columnNumber, Integer columnSize,
+                           String nativeType, Boolean nullable, String remarks, boolean indexed, String quote, boolean primaryKey) {
+        this.name = name;
```
arjansh, line 76: It is common practice to use constructor chaining when using multiple constructors for a java class, so then you could change this to:

```
        this(name, type, table, columnNumber, columnSize, null, nativeType, nullable, remarks, indexed, quote,
                primaryKey);
```

tkingsbury42: OK, this should be ready for another look.

core/src/main/java/org/apache/metamodel/schema/ImmutableColumn.java
arjansh: This looks good now.

core/src/main/java/org/apache/metamodel/schema/MutableColumn.java
arjansh: This looks good now.

core/src/main/java/org/apache/metamodel/schema/ImmutableColumn.java
arjansh: This looks good now.

arjansh:
All my comments have been addressed. @kaspersorensen Can you give it a second look, so that if you approve too, we can merge it?

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcMetadataLoader.java
kaspersorensen: Doesn't look like it's fixed. Did you push the change?

tkingsbury42:
That's a strange build failure. Is there a way to retrigger it and try again?

arjansh:
@tkingsbury42 The easiest way to trigger the build is by pushing a new change. E.g. add one whitespace in a comment and then push the change.

jdbc/src/main/java/org/apache/metamodel/jdbc/JdbcMetadataLoader.java
tkingsbury42: Doh. Pilot error. Look again!

PR #242, opened 2020-07-03
Better handling of nested objects in Elasticsearch data

arjansh:
Fixes METAMODEL-1227.

In case an Elasticsearch document contains nested objects, these should be designated the `ColumnType.MAP`,  because they're resolved to `Map` objects, not to `String` objects.


jhorcicka:
Reviewed, LGTM.

PR #243, opened 2020-07-06
Workaround for fieldnames with dots in elasticsearch

arjansh:
MetaModel is unable to work with a document indexed by Elasticsearch which contains dots in its field names. Note that this is actually caused by Elasticsearch, because the mapping returned for such a document by Elasticsearch doesn't match the source returned by Elasticsearch when getting the source of a search hit. (see https://github.com/elastic/elasticsearch-hadoop/issues/853 and related issues for more info on this).

Note that the branch for this PR is branch off of the branch for https://github.com/apache/metamodel/pull/242, because it extends the Integration Test which is introduced in that PR.

### DIFF
--- a/elasticsearch/common/src/main/java/org/apache/metamodel/elasticsearch/common/ElasticSearchUtils.java
+++ b/elasticsearch/common/src/main/java/org/apache/metamodel/elasticsearch/common/ElasticSearchUtils.java
@@ -293,7 +293,7 @@ public class ElasticSearchUtils {
                     } else if (column.getType() == ColumnType.MAP && value == null) {
                         // Because of a bug in Elasticsearch, when field names contain dots, it's possible that the
                         // mapping of the index described a column to be of the type "MAP", while it's based on a number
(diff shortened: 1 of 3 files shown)


elasticsearch/common/src/main/java/org/apache/metamodel/elasticsearch/common/ElasticSearchUtils.java
```diff
@@ -290,6 +290,23 @@ public static Row createRow(final Map<String, Object> sourceMap, final String do
                         } else {
                             values[i] = valueToDate;
                         }
+                    } else if (column.getType() == ColumnType.MAP && value == null) {
+                        // Because of a bug in Elasticsearch, when field names contain dots, it's possible that the
+                        // mapping of the index described a column to be of the type "MAP", while it's based on a number
+                        // of fields contains dots in their name. In this case we may have to work around that
```
jhorcicka, line 296: "contains dots..." => "containing dots..."

PR #244, opened 2020-12-17
MM-1229: FixedWidth works with UrlResource/HttpInputStream.

jhorcicka:
Fixes https://issues.apache.org/jira/browse/METAMODEL-1229. 

EbcdicReader in FixedWidthDataContext required BufferedInputStream, now other input streams are also supported (new BufferedInputStream(originalInputStream) is created). 
This enables using a URL (UrlResource) for a path in FixedWidth datastore.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 ### Apache MetaModel [WIP]
 
+ * [METAMODEL-1229] - FixedWidth works with UrlResource/HttpInputStream.
  * [METAMODEL-1228] - Better handling of fieldnames with dots in Elasticsearch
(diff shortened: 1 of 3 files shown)


arjansh (changes requested):
The change looks good to me, but can you add a unit test where you construct a `FixedWidthDataContext` with a `UrlResource` which demonstrates you couldn't use it before this change and you can use it now?

arjansh:
The Travis failures seem unrelated. LGTM!

arjansh (changes requested):
LGTM! One small more thing however; can you add the fixed issue to the `CHANGES.md` file in the root folder of the project?

PR #247, opened 2021-01-11
Build on Travis fails most of the time

arjansh:
Fixes https://issues.apache.org/jira/browse/METAMODEL-1232.

- I refactored the `ExcelDataContextTest` from using JUnit 3 to JUnit 4, so it can use the ` TemporaryFolder` rule for copying files when using them it tests. It seems this works better than when copying them to the target folder.
- I refactored the logic used by `ElasticSearchUtils#createRow(...)`  for creating a row based on a MAP column type. It uses `computeIfAbsent` to put a `HashMap`  in a `Map`, but then within the method that created that `HashMap`, it would already add the created `HashMap` to the `Map`, which is something that should be handled by the `computeIfAbsent` itself, which apparently can lead to `ConcurrentModificationExpection`s being thrown.

### DIFF
--- a/excel/src/test/java/org/apache/metamodel/excel/ExcelDataContextTest.java
+++ b/excel/src/test/java/org/apache/metamodel/excel/ExcelDataContextTest.java
@@ -21,7 +21,6 @@ package org.apache.metamodel.excel;
 import static org.junit.Assert.*;
 
 import java.io.File;
-import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.stream.IntStream;
@@ -64,7 +63,6 @@ public class ExcelDataContextTest {
      * 
      * @param path
      * @return
-     * @throws IOException 
      */
     private File copyOf(String path) {
         final File srcFile = new File(path);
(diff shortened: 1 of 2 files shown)


excel/src/test/java/org/apache/metamodel/excel/ExcelDataContextTest.java
```diff
@@ -928,14 +970,15 @@ public void testInsertingValueOfValidColumnType() {
         assertTrue(dataSet.next());
     }
 
-    @Test(expected = MetaModelException.class)
+    @Test
```
kaspersorensen, line 973: Why does this no longer throw the exception? Do we want it not to?

arjansh: It never threw an exception. Because we were using JUnit3 the expected exception part of the annotation was ignored.

arjansh:
@kaspersorensen Can you maybe look at my reply to your comment? And if it satisfies you maybe merge this Pull Request?

arjansh:
@kaspersorensen Can you maybe give this pull request another look. It should make the build more stable so we don't get failing unit tests in all pull requests which are unrelated to these pull requests.

kaspersorensen:
Thanks for the nudge Arjan! Yes this looks good to me then!

PR #248, opened 2021-02-12
METAMODEL-1233: Elasticsearch uses 'match' query for text based columns conditions in WHERE clause

jhorcicka:
Fixes https://issues.apache.org/jira/browse/METAMODEL-1233 

Elasticsearch used a 'term' query for `WHERE text_based_column = 'text-value'` situations. This PR introduces a new condition covering this scenario and creating a 'match' query instead.

### DIFF
--- a/elasticsearch/common/src/test/java/org/apache/metamodel/elasticsearch/common/ElasticSearchUtilsTest.java
+++ b/elasticsearch/common/src/test/java/org/apache/metamodel/elasticsearch/common/ElasticSearchUtilsTest.java
@@ -18,15 +18,24 @@
  */
 package org.apache.metamodel.elasticsearch.common;
 
-import junit.framework.TestCase;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
 import org.apache.metamodel.data.DataSetHeader;
 import org.apache.metamodel.data.Row;
 import org.apache.metamodel.data.SimpleDataSetHeader;
+import org.apache.metamodel.query.FilterItem;
+import org.apache.metamodel.query.OperatorType;
 import org.apache.metamodel.query.SelectItem;
 import org.apache.metamodel.schema.ColumnType;
 import org.apache.metamodel.schema.MutableColumn;
+import org.elasticsearch.index.query.QueryBuilder;
 
-import java.util.*;
+import junit.framework.TestCase;
 
 public class ElasticSearchUtilsTest extends TestCase {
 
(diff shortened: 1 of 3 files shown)


kaspersorensen:
We should probably fix the `IN` operator also then. It's using a termsQuery, too.

jhorcicka:
> We should probably fix the `IN` operator also then. It's using a termsQuery, too.

Actually, DIFFERENT_FROM also does (did) not work. I changed that too.

elasticsearch/common/src/test/java/org/apache/metamodel/elasticsearch/common/ElasticSearchUtilsTest.java
```diff
@@ -60,4 +69,40 @@ public void testCreateRowWithParsableDates() throws Exception {
...
+    /**
+     * For text-based conditions a 'match' query is recommended (instead of 'term' query).
+     * In case of 'DIFFERENT_FROM', we need a 'bool' query with 'must not' and 'match' query.
+     */
+    public void testBoolQueryIsCreatedForTextDifferentFrom() {
+        final SelectItem selectItem = new SelectItem(new MutableColumn("column_name", ColumnType.STRING));
+        final FilterItem filterItem = new FilterItem(selectItem, OperatorType.DIFFERENT_FROM, "text-value");
+        final QueryBuilder queryBuilder =
+                ElasticSearchUtils.createQueryBuilderForSimpleWhere(Collections.singletonList(filterItem), null);
+        assertEquals("bool", queryBuilder.getName());
```
arjansh, line 93: Can you also add assertions to validate that the queryBuilder contains the "must not" and "match" parts in the query?

Technically, you can add something like this:

```
        final Map<String, Object> queryMap = XContentHelper
                .convertToMap(XContentType.JSON.xContent(), queryBuilder.toString(), false);
```

That converts the queryBuilder object to a Map containing the values you're interested in, so then you can add some assertions to check if it contains what you're expecting. The same applies to the other test case (below this one)

arjansh (changes requested):

elasticsearch/common/src/test/java/org/apache/metamodel/elasticsearch/common/ElasticSearchUtilsTest.java
jhorcicka: Thanks for the tip! Updated.

arjansh (changes requested):
The code looks good now. One more thing, can you add the issue that's fixed by this Pull Request to the CHANGES.md in the root folder?

jhorcicka:
> The code looks good now. One more thing, can you add the issue that's fixed by this Pull Request to the CHANGES.md in the root folder?

Yes, sure, I keep forgetting to do that.

PR #250, opened 2021-03-16
MM-1235: AbstractDataContext.searchColumn was fixed for situation when schema and column are equal strings

jhorcicka:
Fixes https://issues.apache.org/jira/browse/METAMODEL-1235. 

Problem in AbstractDataContext.searchColumn when schema name and column name are equal was fixed.

### DIFF
--- a/core/src/main/java/org/apache/metamodel/AbstractDataContext.java
+++ b/core/src/main/java/org/apache/metamodel/AbstractDataContext.java
@@ -309,22 +309,21 @@ public abstract class AbstractDataContext implements DataContext {
      */
     private Column searchColumn(String schemaNameSearch, String columnNameOriginal, String columnNameSearch) {
         final Schema schema = getSchemaByName(schemaNameSearch);
-        if (schema == null) {
-            return null;
-        }
-        if (columnNameSearch.equals(schemaNameSearch)) {
-            return getColumn(schema, columnNameSearch);
-        } else if (columnNameSearch.startsWith(schemaNameSearch)) {
-            String tableAndColumnPath = columnNameOriginal.substring(schemaNameSearch.length());
-            if (tableAndColumnPath.charAt(0) == '.') {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/metamodel/AbstractDataContext.java
```diff
@@ -308,19 +308,17 @@ public final Column getColumnByQualifiedLabel(final String columnName) {
...
-                if (tableAndColumnPath.charAt(0) == '.') {
-                    tableAndColumnPath = tableAndColumnPath.substring(1);
-
-                    Column column = getColumn(schema, tableAndColumnPath);
-                    if (column != null) {
-                        return column;
-                    }
-                }
+        final Schema schema = getSchemaByName(schemaNameSearch);
+        if (schema == null) {
```
arjansh, line 312: This method by default returns `null` if it can't find anything. So maybe change this to `if (schema != null) {` here.

arjansh (commented):
Can you merge master into this branch and see if Travis then builds successfully?

core/src/main/java/org/apache/metamodel/AbstractDataContext.java
jhorcicka: OK

jhorcicka:
> Can you merge master into this branch and see if Travis then builds successfully?

I did, now it fails with oracle jdk8 only, in Excel module. This seems unrelated.

arjansh:
@jhorcicka We forgot to add the last merged PRs related issues to the CHANGES.md, so maybe in the context of this PR, you can also add this to the CHANGES.md:

```
### Apache MetaModel [WIP]

 * [METAMODEL-1236] - Elasticsearch: not/empty and is/not null do not work for text based columns
 * [METAMODEL-1235] - Elasticsearch: "String index out of range" when schema name and column name are equal
 * [METAMODEL-1232] - Improve stability of Travis CI builds

```

PR #252, opened 2021-05-20
Updated getSchemas to directly call internal method

sanveer-osahan:
Currently the time complexity for getSchemas method is O(n^2) considering n number of schemas.
The getSchemaByName call in it can be replaced with getSchemaByNameInternal to reduce the complexity to O(n)


kaspersorensen:
Nice little improvement. Looks good to me.

